PR #4, opened 2022-07-22
Couple of improvements

oerdnj:
This PR contains:

* Add SPDX headers, so tools like `reuse lint` can be used
* Use GCC Visibility attribute instead of providing the list of symbols to the linker (this is recommended usage)
* Generate .tar.xz by default (it's up to your decision to not generate .tar.gz all)
* Modernize the autotools usage a little bit
* Added a GitHub Action to run `reuse lint` automatically


oerdnj:
I've added one extra commit that fixes compilation on macOS (and BSDs) - the endian support is horrible. And fixes couple logic errors.

oerdnj:
> I've added one extra commit that fixes compilation on macOS (and BSDs) - the endian support is horrible. And fixes couple logic errors.

If needed, I can chop up the last commit into as series of smaller ones, but I felt it's not that important given the project is still very fresh.

simo5:
Hi @oerdnj thanks for the PR, however this looks like it should split in several independent ones.
As it is there are parts I like, parts I do not and parts where I need to think about it.

simo5:
> This PR contains:
> 
>     * Add SPDX headers, so tools like `reuse lint` can be used

I do not like much SPDX headers, what is `reuse lint` and why should I care ?
Where does the CC0 license come from ?
Everything in this repository is definitely ASL-2.0 unless explicitly marked otherwise.
Please retain the COPYING file.
This stuff should probably go in its own PR.

>     * Use GCC Visibility attribute instead of providing the list of symbols to the linker (this is recommended usage)

This is recommended when there are many symbols to handle, but in this case there is a single symbols to expose and nothing else. In this case the export file is safer, please drop this patch.


>     * Generate .tar.xz by default (it's up to your decision to not generate .tar.gz all)

This one is fine, I have not yet really worked on the installation and release process part, but I am not against using tar.xz

>     * Modernize the autotools usage a little bit

I am not really a fan of the m4 directory but not against it, looks harmless enough.

>     * Added a GitHub Action to run `reuse lint` automatically

See above, why should I care?



As for the endian patch, I do not care about other platforms at this stage so please drop this path from this PR in any case.

Also please do no mix fixes about logic errors (the <= 2048) with other cosmetic changes.

simo5:
Since you proposed all these changes as separate PRs (Thanks!!) I am going to close this one.

PR #5, opened 2022-07-22
Instead of listing the exported symbols, use GCC Visibility

oerdnj:
Set the default symbol visibility to hidden and add PUBLIC macro that
expands to __attribute__((__visibility__("default"))) and use the
macro to export the OSSL_provider_init symbol.


simo5:
As I mentioned in #4 I do not think this is warranted, I prefer the current method

oerdnj:
Fine. I prefer to use the visibility, but both approaches have merit.

PR #7, opened 2022-07-22
Add SPDX license headers

oerdnj:
Instead of using a custom copyright header, use the well-defined
SPDX format, so the licensing and copyright information can be
checked with 'reuse lint'.


oerdnj:
> I do not like much SPDX headers, what is `reuse lint` and why should I care ?

You do not have to care, but the downstream does - both end consumers and distributions. It makes life much easier.

f.e. Linux kernel uses that: https://www.kernel.org/doc/html/v4.18/process/license-rules.html and I've added this to BIND 9 which made us to worry about one less thing because it's just automated and works better than any custom scripts. Sure, this is small project, but right because this is small project, it's easy to get this right from the begining.

The `reuse lint` comes from https://reuse.software. It's very helpful tool that just automates the license and copyright checks.

> Where does the CC0 license come from ?

This is basically for stuff that is unlicenceable (because of simplicity like Makefile.am with two lines or .gitignore), so it's basically saying "public domain". It's ok to have everything under Apache-2.0

> Everything in this repository is definitely ASL-2.0 unless explicitly marked otherwise.

Sure thing, as said above this is mostly for .gitignore and other simple stuff.

> Please retain the COPYING file.

Ok, re-added it as symlink. Is that ok?

simo5:
> > I do not like much SPDX headers, what is `reuse lint` and why should I care ?
> 
> You do not have to care, but the downstream does - both end consumers and distributions. It makes life much easier.
> 
> f.e. Linux kernel uses that: https://www.kernel.org/doc/html/v4.18/process/license-rules.html and I've added this to BIND 9 which made us to worry about one less thing because it's just automated and works better than any custom scripts. Sure, this is small project, but right because this is small project, it's easy to get this right from the begining.

Meh, I do not care as you may guess, but I do care to avoid extremes, see below.


> The `reuse lint` comes from https://reuse.software. It's very helpful tool that just automates the license and copyright checks.

Is this what required you to put a SPDX line even in the README.md file?
if so please drop the patch that adds the GitHub Action, I am definitely not going to add a SPDX line or a copyright notice for that matter on a 3 line README.

> 
> > Where does the CC0 license come from ?
> 
> This is basically for stuff that is unlicenceable (because of simplicity like Makefile.am with two lines or .gitignore), so it's basically saying "public domain". It's ok to have everything under Apache-2.0

Please drop CC0, also please *do not* add the SPDX ASL header to Makefiles, README, .gitignore and other similar stuff, it is just silly.
File that are "unlicenseable" do not need a license in the first pleace. And they fall under the general project license otherwise.

Also please do not change the files under the /oasis/ directory they are unmodified copies.
I may add a README that states as much if needed. For those, if you want SPDX headers please ask Oasis to add one to their files and then we'll import new ones from them once they have done that.

> > Everything in this repository is definitely ASL-2.0 unless explicitly marked otherwise.
> 
> Sure thing, as said above this is mostly for .gitignore and other simple stuff.

See above

> > Please retain the COPYING file.
> 
> Ok, re-added it as symlink. Is that ok?

No, please retain as is.

Also please undo the reformatting you caused to the Apache-2.0.txt file, git should just report an unmodified rename. 

Also for C file where I am ok adding the SPDX header please do it this way:
/* Copyright (C) 2022 Simo Sorce <simo@redhat.com>
   SPDX-License-Identifier: Apache-2.0 */

No more than two lines per file please

simo5:
Ok now that I have seen the REUSE action I am 100% against it, please drop it.

oerdnj:
> Ok now that I have seen the results of the REUSE action I am 100% against it, please drop it.

I think it's useful to make things that can be automated to be automated rather than manually maintained, but this is not the hill I want to die on. Closed the PR.

simo5:
Note that I was not against adding SPDX lines where it matters,
but adding them on exports files or READMEs or .gitignore is just insane.
Once people working on REUSE make it sane and check just the code files we can add it.

oerdnj:
> Note that I was not against adding SPDX lines where it matters,
> 
> but adding them on exports files or READMEs or .gitignore is just insane.
> 
> Once people working on REUSE make it sane and check just the code files we can add it.

It's possible to handle these files separately in .reuse/ directory like this: https://gitlab.isc.org/isc-projects/bind9/-/blob/main/.reuse/dep5 and keep the non-.c non-.h and external files untouched.

PR #8, opened 2022-07-22
Add compatibility shim for endian related functions

oerdnj:
The <endian.h> isn't POSIX, add a compatibility shim (taken and
relicensed from BIND 9) to support various BSDs and macOS systems.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -15,6 +15,7 @@ pkcs11_provider_la_SOURCES = \
 	oasis/pkcs11f.h \
 	oasis/pkcs11t.h \
 	pkcs11.h \
+	platform/endian.h \
 	provider.h \
 	provider.c \
 	signature.c \
--- a/src/store.c
+++ b/src/store.c
@@ -2,7 +2,7 @@
 
 #include "provider.h"
 #include <openssl/store.h>
-#include <endian.h>
+#include "platform/endian.h"
 #include <string.h>
 
 struct p11prov_uri {
--- a/.gitignore
+++ b/.gitignore
@@ -34,5 +34,6 @@ src/*.o
 *~
 src/*~
 test-driver
+tests/openssl.cnf
 tests/tokens
 tests/tmp
--- a/src/platform/endian.h
+++ b/src/platform/endian.h
@@ -1,0 +1,158 @@
+/* Copyright (C) Internet Systems Consortium, Inc. ("ISC") - see COPYING */
+
+#ifndef _P11PROV_ENDIAN_H
+#define _P11PROV_ENDIAN_H
+
+#if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__NetBSD__) || \
+	defined(__OpenBSD__) || defined(__bsdi__)
+
+#include <sys/endian.h>
+
+/*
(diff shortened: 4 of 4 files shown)


src/Makefile.am
```diff
@@ -5,8 +5,9 @@ lib_LTLIBRARIES = pkcs11_provider.la
 SHARED_EXT=@SHARED_EXT@
 
 pkcs11_provider_la_SOURCES = \
-	asymmetric_cipher.c \
+asymmetric_cipher.c \
```
simo5, line 8: why this change?

src/endian.h
```diff
@@ -0,0 +1,155 @@
+/* Copyright (C) Internet Systems Consortium, Inc. ("ISC") - see COPYING */
+
+#pragma once
```
simo5, line 3: I would rather use the usual ifdef based guards here

simo5 (changes requested):
Boy is it ugly to support multiple platforms ... I wonder if this is the end of it, do we have specific uses in mind though?

That said, please move the file under a directory named "isc" like with the "oasis" headers.

src/Makefile.am
oerdnj: > Boy is it ugly to support multiple platforms ... I wonder if this is the end of it, do we have specific uses in mind though?

It compiles fine on macOS now.

> That said, please move the file under a directory named "isc" like with the "oasis" headers.

Why? There's nothing ISC specific, and I (personally) wrote the header, so it's ok to have it as public domain or whatever. There's no magic in there.

oerdnj: > why this change?

This was unintentional. I do have some white space formatting macros in my emacs, so perhaps it reformatted the whitespace. I'll revert this.

src/endian.h
oerdnj: `#pragma once` is well supported even in ancient compilers, but ok

oerdnj: Also with `#pragma once` you don't have to come up with fancy names like `_P11PROV_ENDIAN_H`.

src/store.c
```diff
@@ -2,7 +2,7 @@
 
 #include "provider.h"
 #include <openssl/store.h>
-#include <endian.h>
+#include "endian.h"
```
simo5, line 5: Move endian.h -> isc/endian.h please

simo5 (changes requested):

simo5:
I assume this si a file generally provided by ISC ?
This is why I asked to move it under an isc/ directory.
If it is not something ISC regularly distributes (like the Oasis files) and I misunderstood, then I think I would still like to add it under a directory like "compat" or "platform" to indicate this is not a file to use to put in random new stuff but only compatibility layers for other platforms.

oerdnj:
It's not something that we distribute. It's just I had to solve this for BIND 9, so platform/ works for me...

src/Makefile.am
```diff
@@ -7,6 +7,7 @@ SHARED_EXT=@SHARED_EXT@
 pkcs11_provider_la_SOURCES = \
 	asymmetric_cipher.c \
 	debug.c \
+	endian.h \
```
simo5, line 10: think this need to be changed accordingly

oerdnj: Yep, good catch, fixed.

simo5 (approved):

PR #9, opened 2022-07-22
Fix operator precedence errors

oerdnj:
As reported by clang, the & has lower precedence than == operator.
Use parentheses as needed when checking for a set bit.


simo5:
Sounds like a clang specific thing, works fine with gcc, but I see no harm in tightening this up

PR #11, opened 2022-07-22
Make RTLD_DEEPBIND optional

oerdnj:
The RTLD_DEEPBIND flag is not present on all platforms.  Use it only if
it's present on the platform.

### DIFF
--- a/src/provider.c
+++ b/src/provider.c
@@ -526,6 +526,10 @@ done:
     return ret;
 }
 
+#if !defined(RTLD_DEEPBIND)
+#define RTLD_DEEPBIND 0
+#endif
+
 static int p11prov_module_init(P11PROV_CTX *ctx)
 {
     CK_RV (*c_get_function_list)(CK_FUNCTION_LIST_PTR_PTR);


src/provider.c
```diff
@@ -544,7 +544,11 @@ static int p11prov_module_init(P11PROV_CTX *ctx)
 
     dlerror();
     ctx->dlhandle = dlopen(ctx->module,
-                           RTLD_NOW | RTLD_LOCAL | RTLD_DEEPBIND);
+                           RTLD_NOW | RTLD_LOCAL
+#if defined(RTLD_DEEPBIND)
+			   | RTLD_DEEPBIND
+#endif /* defined(RTLD_DEEPBIND) */
+	    );
```
simo5, line 551: Uuuh this is too ugly,
what about adding at the top of the file:

#if !defined(RTLD_DEEPBIND)
#define RTLD_DEEPBIND 0
#endif

this will deuglify usage later.

simo5 (changes requested):

src/provider.c
oerdnj: Sure, works for me.

src/provider.c
```diff
@@ -526,6 +526,10 @@ static int refresh_slots(P11PROV_CTX *ctx)
     return ret;
 }
 
+#if defined(RTLD_DEEPBIND)
```
simo5, line 529: This should be !defined

oerdnj: I swear I typed the `!` and it disappeared on its own :)

simo5 (approved):

PR #12, opened 2022-07-22
Add headers to the Makefile.am

oerdnj:
The automake needs to know what files are part of the distribution
when making distribution tarball and to calculate the dependencies.
Add all used headers to the pkcs11_provider_la_SOURCES.


simo5:
Heh as I mentioned I haven't looked into distribution yet, never even once ran make dist and make distcheck :-D

Thanks for catching this stuff, please just add your sign-off and I will merge.

PR #13, opened 2022-07-22
Fix mismatch between CK_UTF8CHAR_PTR and const char *

oerdnj:
Some compilers (clang) complain when CK_UTF8CHAR_PTR is passed to
const char * argument because char might or might not be signed
depending on the platform.  Properly cast the values to silence
the warnings.

### DIFF
--- a/src/store.c
+++ b/src/store.c
@@ -451,27 +451,30 @@ static int p11prov_store_load(void *ctx,
 
         /* skip slots that do not match */
         if (obj->parsed_uri->model &&
-            strncmp(obj->parsed_uri->model, token.model, 16) != 0)
+            strncmp(obj->parsed_uri->model,
+                    (const char *)token.model, 16) != 0)
             continue;
         if (obj->parsed_uri->manufacturer &&
             strncmp(obj->parsed_uri->manufacturer,
-                    token.manufacturerID, 32) != 0)
+                    (const char *)token.manufacturerID, 32) != 0)
             continue;
         if (obj->parsed_uri->token &&
-            strncmp(obj->parsed_uri->token, token.label, 32) != 0)
+            strncmp(obj->parsed_uri->token,
+                    (const char *)token.label, 32) != 0)
             continue;
         if (obj->parsed_uri->serial &&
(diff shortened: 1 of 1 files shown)


src/store.c
```diff
@@ -451,27 +451,27 @@ static int p11prov_store_load(void *ctx,
...
-                    token.manufacturerID, 32) != 0)
+                    (const char *)token.manufacturerID, 32) != 0)
             continue;
         if (obj->parsed_uri->token &&
-            strncmp(obj->parsed_uri->token, token.label, 32) != 0)
+            strncmp(obj->parsed_uri->token, (const char *)token.label, 32) != 0)
             continue;
         if (obj->parsed_uri->serial &&
-            strncmp(obj->parsed_uri->serial, token.serialNumber, 16) != 0)
+            strncmp(obj->parsed_uri->serial, (const char *)token.serialNumber, 16) != 0)
```
simo5, line 464: please fold lines like this that go over 80 chars

simo5 (changes requested):
Makes the code a bit ugly, but it is what it is.

src/store.c
```diff
@@ -451,27 +451,28 @@ static int p11prov_store_load(void *ctx,
 
         /* skip slots that do not match */
         if (obj->parsed_uri->model &&
-            strncmp(obj->parsed_uri->model, token.model, 16) != 0)
+            strncmp(obj->parsed_uri->model, (const char *)token.model, 16) != 0)
```
simo5, line 454: pleas fold

src/store.c
```diff
@@ -451,27 +451,28 @@ static int p11prov_store_load(void *ctx,
...
             continue;
         if (obj->parsed_uri->manufacturer &&
             strncmp(obj->parsed_uri->manufacturer,
-                    token.manufacturerID, 32) != 0)
+                    (const char *)token.manufacturerID, 32) != 0)
             continue;
         if (obj->parsed_uri->token &&
-            strncmp(obj->parsed_uri->token, token.label, 32) != 0)
+            strncmp(obj->parsed_uri->token, (const char *)token.label, 32) != 0)
```
simo5, line 461: please fold

simo5 (approved):

PR #15, opened 2022-07-22
Add initial .clang-format style and reformat the sources using it

oerdnj:
This is an initial example of how .clang-format can be configured. It doesn't precisely match the current style, but no tool can do that.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,72 @@
+BasedOnStyle: LLVM
+ColumnLimit: 80
+IndentWidth: 4
+UseTab: Never
+BraceWrapping:
+  AfterCaseLabel:  false
+  AfterClass:      true
+  AfterControlStatement: Never
+  AfterEnum:       false
+  AfterFunction:   true
+  AfterNamespace:  true
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  AfterExternBlock: false
+  BeforeCatch:     false
+  BeforeElse:      false
+  BeforeLambdaBody: false
+  BeforeWhile:     false
+  IndentBraces:    false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+BreakBeforeBinaryOperators: NonAssignment
+BreakBeforeConceptDeclarations: true
+BreakBeforeBraces: Linux
+BreakBeforeInheritanceComma: false
+BreakInheritanceList: BeforeColon
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+AlignAfterOpenBracket: Align
+AlignEscapedNewlines: DontAlign
+AllowAllArgumentsOnNextLine: true
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortIfStatementsOnASingleLine: true
+AllowAllParametersOfDeclarationOnNextLine: false
+BinPackArguments: true
+BinPackParameters: true
+#InsertTrailingCommas: Wrapped
+ReflowComments: false
+Cpp11BracedListStyle: false
+AlignArrayOfStructures: None
+AlignTrailingComments: false
+AllowShortFunctionsOnASingleLine: Empty
+AlwaysBreakBeforeMultilineStrings: true
+IndentGotoLabels: false
+SortIncludes: Never
+SpaceBeforeCpp11BracedList: true
+SpaceBeforeCtorInitializerColon: false
+SpaceBeforeInheritanceColon: false
+SpaceBeforeParensOptions:
+  AfterControlStatements: true
+  AfterForeachMacros: false
+  AfterFunctionDefinitionName: false
+  AfterFunctionDeclarationName: false
+  AfterIfMacros: false
+  AfterOverloadedOperator: false
+  BeforeNonEmptyParentheses: false
+SpaceBeforeRangeBasedForLoopColon: false
+SpaceInEmptyBlock: false
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 2
+SpacesInAngles:  Never
+SpacesInConditionalStatement: false
+SpacesInContainerLiterals: false
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+SpaceBeforeSquareBrackets: false
--- a/src/asymmetric_cipher.c
+++ b/src/asymmetric_cipher.c
@@ -311,7 +311,7 @@ static struct {
     { CKM_RSA_PKCS, RSA_PKCS1_PADDING, OSSL_PKEY_RSA_PAD_MODE_PKCSV15 },
     { CKM_RSA_PKCS_OAEP, RSA_PKCS1_OAEP_PADDING, OSSL_PKEY_RSA_PAD_MODE_OAEP },
     { CKM_RSA_X9_31, RSA_X931_PADDING, OSSL_PKEY_RSA_PAD_MODE_X931 },
-    { CK_UNAVAILABLE_INFORMATION, 0, NULL }
+    { CK_UNAVAILABLE_INFORMATION, 0, NULL },
 };
 
 /* only the ones we can support */
@@ -329,7 +329,7 @@ static struct {
     { "SHA256", CKM_SHA256, CKG_MGF1_SHA256 },
     { "SHA224", CKM_SHA224, CKG_MGF1_SHA224 },
     { "SHA1", CKM_SHA_1, CKG_MGF1_SHA1 },
-    { NULL, 0, 0 }
+    { NULL, 0, 0 },
 };
 
 static const char *p11prov_rsaenc_digest_name(CK_MECHANISM_TYPE digest)
@@ -531,11 +531,11 @@ static const OSSL_PARAM *p11prov_rsaenc_gettable_ctx_params(void *ctx,
         OSSL_PARAM_utf8_string(OSSL_ASYM_CIPHER_PARAM_OAEP_DIGEST, NULL, 0),
         OSSL_PARAM_utf8_string(OSSL_ASYM_CIPHER_PARAM_MGF1_DIGEST, NULL, 0),
         OSSL_PARAM_octet_string(OSSL_ASYM_CIPHER_PARAM_OAEP_LABEL, NULL, 0),
-/*
+        /*
         OSSL_PARAM_uint(OSSL_ASYM_CIPHER_PARAM_TLS_CLIENT_VERSION, NULL),
         OSSL_PARAM_uint(OSSL_ASYM_CIPHER_PARAM_TLS_NEGOTIATED_VERSION, NULL),
-*/
-        OSSL_PARAM_END
+        */
+        OSSL_PARAM_END,
     };
     return params;
 }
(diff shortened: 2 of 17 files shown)


oerdnj:
If there's something that does specifically bothers you, I can take a look if that's configurable: https://releases.llvm.org/14.0.0/tools/clang/docs/ClangFormatStyleOptions.html

But I prefer the style to stay consistent across all committers over matching every bit of my preferred style.

There's one nit though - I think that even the single line statements should be surrounded by braces - as a defensive programming, I find these:

```
        if (obj->parsed_uri->model &&
            strncmp(obj->parsed_uri->model, (const char *)token.model, 16) != 0)
            continue;
        if (obj->parsed_uri->manufacturer &&
            strncmp(obj->parsed_uri->manufacturer,
                    (const char *)token.manufacturerID,
                    32) != 0)
            continue;
        if (obj->parsed_uri->token &&
            strncmp(obj->parsed_uri->token, (const char *)token.label, 32) != 0)
            continue;
        if (obj->parsed_uri->serial && strncmp(obj->parsed_uri->serial,
                                               (const char *)token.serialNumber,
                                               16) != 0)
            continue;
```

extremely hard to parse.

src/asymmetric_cipher.c
```diff
@@ -69,8 +70,8 @@ static void *p11prov_rsaenc_dupctx(void *ctx)
     if (encctx->oaep_params.pSourceData) {
         CK_RSA_PKCS_OAEP_PARAMS_PTR src = &encctx->oaep_params;
         CK_RSA_PKCS_OAEP_PARAMS_PTR dst = &newctx->oaep_params;
-        dst->pSourceData = OPENSSL_memdup(src->pSourceData,
-                                          src->ulSourceDataLen);
+        dst->pSourceData =
+            OPENSSL_memdup(src->pSourceData, src->ulSourceDataLen);
```
simo5, line 74: I am ok with this

src/asymmetric_cipher.c
```diff
@@ -106,7 +107,9 @@ static int p11prov_rsaenc_encrypt_init(void *ctx, void *provkey,
     P11PROV_OBJECT *obj = (P11PROV_OBJECT *)provkey;
 
     p11prov_debug("encrypt init (ctx=%p, key=%p, params=%p)\n",
-                  ctx, provkey, params);
+                  ctx,
+                  provkey,
+                  params);
```
simo5, line 113: But this one is a bit gross, is there a way to make indentation greedy and stay on the line until 80 chars are hit and only then start a new line?

In rare cases having one argument per line is better, but it is generally wasteful.

src/asymmetric_cipher.c
```diff
@@ -117,10 +120,9 @@ static int p11prov_rsaenc_encrypt_init(void *ctx, void *provkey,
     return p11prov_rsaenc_set_ctx_params(ctx, params);
 }
 
-static int p11prov_rsaenc_encrypt(void *ctx,
-                                  unsigned char *out, size_t *outlen,
-                                  size_t outsize,
-                                  const unsigned char *in, size_t inlen)
+static int p11prov_rsaenc_encrypt(void *ctx, unsigned char *out, size_t *outlen,
+                                  size_t outsize, const unsigned char *in,
+                                  size_t inlen)
```
simo5, line 125: so this one is doing what I'd like to happen for argument passed to a function as well, see above.

src/asymmetric_cipher.c
```diff
@@ -293,43 +300,41 @@ static int p11prov_rsaenc_decrypt(void *ctx,
...
-    { CKM_RSA_PKCS, RSA_PKCS1_PADDING, OSSL_PKEY_RSA_PAD_MODE_PKCSV15 },
-    { CKM_RSA_PKCS_OAEP, RSA_PKCS1_OAEP_PADDING, OSSL_PKEY_RSA_PAD_MODE_OAEP },
-    { CKM_RSA_X9_31, RSA_X931_PADDING, OSSL_PKEY_RSA_PAD_MODE_X931 },
-    { CK_UNAVAILABLE_INFORMATION, 0, NULL }
-};
+    {CKM_RSA_X_509, RSA_NO_PADDING, OSSL_PKEY_RSA_PAD_MODE_NONE},
+    {CKM_RSA_PKCS, RSA_PKCS1_PADDING, OSSL_PKEY_RSA_PAD_MODE_PKCSV15},
+    {CKM_RSA_PKCS_OAEP, RSA_PKCS1_OAEP_PADDING, OSSL_PKEY_RSA_PAD_MODE_OAEP},
+    {CKM_RSA_X9_31, RSA_X931_PADDING, OSSL_PKEY_RSA_PAD_MODE_X931},
+    {CK_UNAVAILABLE_INFORMATION, 0, NULL}};
```
simo5, line 322: I think leaving spaces within braces make it more legible, but it is not a total deal breaker, is this configurable?

src/asymmetric_cipher.c
```diff
@@ -293,43 +300,41 @@ static int p11prov_rsaenc_decrypt(void *ctx,
...
-    { "SHA3-256", CKM_SHA3_256, CKG_MGF1_SHA3_256 },
-    { "SHA3-224", CKM_SHA3_224, CKG_MGF1_SHA3_224 },
-    { "SHA512", CKM_SHA512, CKG_MGF1_SHA512 },
-    { "SHA384", CKM_SHA384, CKG_MGF1_SHA384 },
-    { "SHA256", CKM_SHA256, CKG_MGF1_SHA256 },
-    { "SHA224", CKM_SHA224, CKG_MGF1_SHA224 },
-    { "SHA1", CKM_SHA_1, CKG_MGF1_SHA1 },
-    { NULL, 0, 0 }
-};
+} digest_map[] = {{"SHA3-512", CKM_SHA3_512, CKG_MGF1_SHA3_512},
```
simo5, line 328: here going to a new line would be better, is it configurable?
Or does it do it only if the inititalization argument is too long like with the padding_map case?

src/asymmetric_cipher.c
```diff
@@ -530,12 +535,13 @@ static const OSSL_PARAM *p11prov_rsaenc_gettable_ctx_params(void *ctx,
...
-*/
-        OSSL_PARAM_END
-    };
+        /*
+                OSSL_PARAM_uint(OSSL_ASYM_CIPHER_PARAM_TLS_CLIENT_VERSION,
+           NULL),
+                OSSL_PARAM_uint(OSSL_ASYM_CIPHER_PARAM_TLS_NEGOTIATED_VERSION,
+           NULL),
+        */
+        OSSL_PARAM_END};
```
simo5, line 544: This one is really ugly, can indentation be turned off for comments?

src/store.c
```diff
@@ -451,20 +448,19 @@ static int p11prov_store_load(void *ctx,
...
-                    (const char *)token.label, 32) != 0)
+            strncmp(obj->parsed_uri->token, (const char *)token.label, 32) != 0)
             continue;
-        if (obj->parsed_uri->serial &&
-            strncmp(obj->parsed_uri->serial,
-                    (const char *)token.serialNumber, 16) != 0)
+        if (obj->parsed_uri->serial && strncmp(obj->parsed_uri->serial,
+                                               (const char *)token.serialNumber,
+                                               16) != 0)
             continue;
```
simo5, line 464: This one is pretty bad

simo5:
> If there's something that does specifically bothers you, I can take a look if that's configurable: https://releases.llvm.org/14.0.0/tools/clang/docs/ClangFormatStyleOptions.html
> 
> But I prefer the style to stay consistent across all committers over matching every bit of my preferred style.

Overall I do not mind formatters, but sometimes they make so me silly choices.
Is it possible to make the formatter indent only the changes made to a file but not the entire file in order to preserve exceptions (like so specifically formatted comments) as needed ?

> There's one nit though - I think that even the single line statements should be surrounded by braces - as a defensive programming, I find these:
> 
> ```
>         if (obj->parsed_uri->model &&
>             strncmp(obj->parsed_uri->model, (const char *)token.model, 16) != 0)
>             continue;
>         if (obj->parsed_uri->manufacturer &&
>             strncmp(obj->parsed_uri->manufacturer,
>                     (const char *)token.manufacturerID,
>                     32) != 0)
>             continue;
>         if (obj->parsed_uri->token &&
>             strncmp(obj->parsed_uri->token, (const char *)token.label, 32) != 0)
>             continue;
>         if (obj->parsed_uri->serial && strncmp(obj->parsed_uri->serial,
>                                                (const char *)token.serialNumber,
>                                                16) != 0)
>             continue;
> ```
> 
> extremely hard to parse.

I think adding braces here is just fine, I do not mind. In fact I prefer braces even for sinle line if constructs, I just got a bit lazy in this case.

simo5 (commented):
It is not too bad, but let's see if it can be improved (I know subjective, but style often is).

.clang-format
```diff
@@ -0,0 +1,10 @@
+BasedOnStyle: LLVM
+IndentWidth: 4
+UseTab: Never
+BreakBeforeBraces: Linux
+AlignAfterOpenBracket: Align
+AllowShortIfStatementsOnASingleLine: true
+AllowAllArgumentsOnNextLine: false
```
simo5, line 7: I guess this should be turned to true for one of my requests?

.clang-format
```diff
@@ -0,0 +1,10 @@
...
+BinPackArguments: false
+BinPackArguments: false
```
simo5, line 9: Unintentionally doubled?

.clang-format
```diff
@@ -0,0 +1,10 @@
+BasedOnStyle: LLVM
+IndentWidth: 4
+UseTab: Never
+BreakBeforeBraces: Linux
+AlignAfterOpenBracket: Align
```
simo5, line 5: BAS_AlwaysBreak ?

simo5: Also add:
AlignEscapedNewlines DontAlign

simo5: Also add:
AllowAllArgumentsOnNextLine: false

simo5:
do you mind if I experiment and rebase your branch ?
I think we can get much closer to some of the nits I don't like.

.clang-format
oerdnj: It's the BinPack arguments bellow I think. I'll play with it a bit more, I don't know all the options by heart. It's more like tune it once and then check if newer clang-format fixes something or adds new option that was missing.

.clang-format
oerdnj: Ack, the other one should have been the other BinPack<something>...

oerdnj:
I'll also just one shot the single line braces.  I thought that clang-format does have support for it already but seems it does not, but I can do that with clang-tidy, but it's not that easy to wire it in, so it's more one-time only.

simo5:
I'll probably continue review after the weekend, I think we can get close enough.

oerdnj:
> do you mind if I experiment and rebase your branch ?
> 
> I think we can get much closer to some of the nits I don't like.

I absolutely don't mind, go ahead.  I absolutely also mind a little about the exact style, although going full K&R would be a little hard ;).

Once we nail this, there's also a GitHub Action for clang-format or we can write one.

src/exchange.c
```diff
@@ -47,7 +47,12 @@ static struct {
     DM_ELEM_SHA(384),
     DM_ELEM_SHA(224),
     { "SHA1", CKM_SHA_1, CKD_SHA1_KDF, 20 },
-    { NULL, 0, 0, 0, }
+    {
+        NULL,
+        0,
+        0,
+        0,
+    },
```
simo5, line 55: This is still off, clang-format just doesn't get it... guess will have to sdjust manually before merging

simo5:
Not super-happy about how it monkeys with macros, and the whole bizarre things it does with struct initializers with arrays, unless I manually add a comma after the last initializer.

However this comes close enough ...
I am going to try to add a job that checks diffs on pull requests, so that it doesn't complain about intentionally manually arranged stuff.

src/exchange.c
oerdnj: Remove the last comma after `0`?

simo5: uh, could be I guess

simo5:
Ok, this looks reasonably ok.

The Makefile part is not stellar but gets the job done.

What do you think?

simo5:
Obviously it shows a failure because of the commit where I fixed some clang-format "mistakes".

oerdnj:
> Obviously it shows a failure because of the commit where I fixed some clang-format "mistakes".

There's a command you can put in the comment to enable/disable formatting of the part of the code.

My experience is that it's simply easier to accept that the tool is not perfect and just let it go...

simo5:
> My experience is that it's simply easier to accept that the tool is not perfect and just let it go...

Ok, I guess those macros can live as is ...
When clang will grow some support for better initializer and macro formatting we can revisit this.

PR #22, opened 2022-07-25
Include <strings.h> to get strcasecmp() declaration

oerdnj:


simo5:
Ah this needs to be fixed to use OPENSSL_strcasecmp() instead.

Just using strcasecmp is not ok as it runs afoul of issues with locales that have been fixed in the openssl version.

oerdnj:
Ah, ok, replaced with #26

PR #24, opened 2022-07-25
Add GitHub Action to build with clang

oerdnj:
The LLVM/Clang has different checks, so it's often useful to compile the project also with non-GCC compiler.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,17 +10,20 @@ jobs:
   build:
     name: CI with softoken
     runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        compiler: [gcc, clang]
     steps:
     - uses: actions/checkout@v3
       name: Checkout Repository
     - name: Install Dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y autoconf-archive libnss3 libnss3-tools libnss3-dev
+        sudo apt-get install -y autoconf-archive libnss3 libnss3-tools libnss3-dev clang
     - name: Setup
       run: |
         autoreconf -fiv
-        ./configure
+        CC=${{ matrix.compiler }} ./configure
     - name: Build and Test
       run: |
         make


simo5:
Please just add another job or a matrix to the main build.yml file.
See this example where I use a different python compiler for each job, everything else is the same:
https://github.com/latchset/jwcrypto/blob/master/.github/workflows/build.yml

Otherwise we have high maintenance cost of changing many different .yml files every time we need to change some build/configure option.

oerdnj:
> Please just add another job or a matrix to the main build.yml file. See this example where I use a different python compiler for each job, everything else is the same: https://github.com/latchset/jwcrypto/blob/master/.github/workflows/build.yml

Done. I've been mostly using GitLab for the past 10 years, so this GitHub Actions thing is quite new to me.

simo5:
You should drop all changes where you add <strings.h> include, I thin those are remnants of the strcasecmp business.

.github/workflows/build.yml
```diff
@@ -10,17 +10,20 @@ jobs:
...
     - name: Install Dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y autoconf-archive libnss3 libnss3-tools libnss3-dev
+        sudo apt-get install -y autoconf-archive libnss3 libnss3-tools libnss3-dev clang
     - name: Setup
       run: |
         autoreconf -fiv
-        ./configure
+        CC=${{ matrix.compiler }} ./configure
```
simo5, line 26: perfect!

simo5:
> > Please just add another job or a matrix to the main build.yml file. See this example where I use a different python compiler for each job, everything else is the same: https://github.com/latchset/jwcrypto/blob/master/.github/workflows/build.yml
> 
> Done. I've been mostly using GitLab for the past 10 years, so this GitHub Actions thing is quite new to me.

I am no expert either, I just recently succumbed to use yaml instead of json, because not all constructs are supported there ...
I just happened to fin the matrix thing super-useful in the past.

PR #25, opened 2022-07-25
Enable all (most) of the warnings as errors

oerdnj:
Enable -Wall -Wextra and couple of additional warnings and treat
them as errors.

Because the implementation is not complete, disable unused parameter
warning.

And fix all issues found.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -19,6 +19,26 @@ AX_CHECK_COMPILE_FLAG([-std=c11],
                       AC_MSG_ERROR([C compiler must support at least C11 standard])
 )
 
+STD_CFLAGS="-Wall -Wextra -Wwrite-strings -Wpointer-arith -Wno-missing-field-initializers -Wformat -Wshadow"
+
+# Temporarily disable unused parameter until the implementation is complete
+STD_CFLAGS="$STD_CFLAGS -Wno-unused-parameter"
+
+# These should be always errors
+STD_CFLAGS="$STD_CFLAGS -Werror=implicit-function-declaration -Werror=missing-prototypes -Werror=format-security -Werror=parentheses -Werror=implicit -Werror=strict-prototypes"
+
+# Don't enable warnings on VLA yet, but I would avoid using VLAs
+# STD_CFLAGS"$STD_CFLAGS -Werror=vla"
+
+AX_CHECK_COMPILE_FLAG([-fno-strict-aliasing],
+		      [STD_CFLAGS="$STD_CFLAGS -fno-strict-aliasing"])
+AX_CHECK_COMPILE_FLAG([-Werror -fno-delete-null-pointer-checks],
+		      [STD_CFLAGS="$STD_CFLAGS -fno-delete-null-pointer-checks"])
+AX_CHECK_COMPILE_FLAG([-fdiagnostics-show-option],
+		      [STD_CFLAGS="$STD_CFLAGS -fdiagnostics-show-option"])
+
+AC_SUBST([STD_CFLAGS])
+
 # Checks for libraries.
 PKG_CHECK_MODULES(
 	[OPENSSL],
--- a/src/asymmetric_cipher.c
+++ b/src/asymmetric_cipher.c
@@ -61,7 +61,6 @@ static void *p11prov_rsaenc_dupctx(void *ctx)
 {
     struct p11prov_rsaenc_ctx *encctx = (struct p11prov_rsaenc_ctx *)ctx;
     struct p11prov_rsaenc_ctx *newctx;
-    int ret;
 
     if (encctx == NULL) {
         return NULL;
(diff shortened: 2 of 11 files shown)


configure.ac
```diff
@@ -54,6 +54,26 @@ AC_TYPE_SIZE_T
...
+
+AX_CHECK_COMPILE_FLAG([-fno-strict-aliasing],
+		      [STD_CFLAGS="$STD_CFLAGS -fno-strict-aliasing"])
+AX_CHECK_COMPILE_FLAG([-Werror -fno-delete-null-pointer-checks],
+		      [STD_CFLAGS="$STD_CFLAGS -fno-delete-null-pointer-checks"])
+AX_CHECK_COMPILE_FLAG([-fdiagnostics-show-option],
+		      [STD_CFLAGS="$STD_CFLAGS -fdiagnostics-show-option"])
+
+AC_SUBST([STD_CFLAGS])
+
```
simo5, line 76: This should stay near the check for c11, either all up to it, or move the c11 cheack down here...

src/provider.h
```diff
@@ -18,7 +18,7 @@
 #include <openssl/proverr.h>
 #include <openssl/core_names.h>
 
-#define UNUSED __attribute__((unused))
+#define UNUSED __attribute__((__unused__))
```
simo5, line 21: Why this change ?

src/asymmetric_cipher.c
```diff
@@ -3,6 +3,7 @@
 
 #include "provider.h"
 #include <string.h>
+#include <strings.h>
```
simo5, line 6: this came from the strcasecmp) usage right?
If so drop from here and the other two files

src/kdf.c
```diff
@@ -298,7 +298,7 @@ static int p11prov_hkdf_set_ctx_params(void *ctx, const OSSL_PARAM params[])
     /* can be multiple paramaters, which wil be all concatenated */
     for (p = OSSL_PARAM_locate_const(params, OSSL_KDF_PARAM_INFO); p != NULL;
          p = OSSL_PARAM_locate_const(p + 1, OSSL_KDF_PARAM_INFO)) {
-        void *ptr;
+        uint8_t *ptr;
```
simo5, line 301: Why this change ?

src/provider.h
oerdnj: There was issue in the past with the standard `noreturn` (from `#include <stdnoreturn.h>`) which mangled system headers that use `__attribute__((noreturn))` and mangled it to `__attribute__((_Noreturn))` which in turn `clang` ignored, so it's kind of habit to use the underscored variants of attributes, especially in headers.

There's no other reason here, so this could be dropped, I did this on autopilot.

src/kdf.c
simo5: Might be better to cast later if it is a type issue.

simo5: Thinking again, it is probably ok, nvm

src/keymgmt.c
```diff
@@ -165,7 +165,8 @@ static int p11prov_rsakm_secbits(int bits)
     if (bits > 4096) return 152;
     if (bits > 3072) return 128;
     if (bits > 2048) return 112;
-    if (bits <= 2048) return 0;
```
simo5, line 168: I think I will have to fix these comparisons, they probably should all be >=, returning 0 for 2048 is wrong ...

src/kdf.c
oerdnj: Pointer arithmetics on `void *` is GNU extension and not defined by C standard.

configure.ac
oerdnj: Done.

src/provider.h
oerdnj: Removed.

src/kdf.c
simo5: Yeah, I realized that when I expanded the code below, I already forgotten this code :-D

src/provider.h
simo5: As lon as I know the reason I am fin with it

src/provider.c
```diff
@@ -312,9 +312,14 @@ static int p11prov_get_capabilities(void *provctx, const char *capability,
...
 
+struct reason_string {
+    unsigned int id;
+    const char *ptr;
+};
+
 static const OSSL_ITEM *p11prov_get_reason_strings(void *provctx)
 {
-    static const OSSL_ITEM reason_strings[] = {
+    static const struct reason_string reason_strings[] = {
```
simo5, line 322: Why this change?

oerdnj: The `OSSL_ITEM` is `struct { int, void * }` and the compiler complains about dropping the `const` modifier when assigning `const char *` to `void *`.

It was either this or typecasting all strings in the array to `char *` (which also would be a valid fix). I am fine with either, but this is less chatty.

src/keymgmt.c
oerdnj: I did this while rebasing and fixing the conflicts, but if you want this as a separate commit, that's fine too.

src/asymmetric_cipher.c
oerdnj: Yeah, it had the extra commits from other branches which are now merged, so I dropped this.

src/provider.c
simo5: yeah lets do a
#define C(str) (void *)(str)
before the struct and call them all as C("string")

I rather not shadow structures defined in headers.

oerdnj: Done

simo5: perfect

PR #30, opened 2022-07-25
Add GitHub Action for Coverity Scan

oerdnj:
This requires setting two secrets, but I've tested it on my fork, and it works as it should.

### DIFF
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -1,0 +1,27 @@
+name: Coverity Scan
+
+on:
+  schedule:
+    - cron:  '41 3 * * 1-5'
+
+jobs:
+  coverity:
+    name: Coverity Scan
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v3
+      name: Checkout Repository
+    - name: Install Dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y autoconf-archive libnss3 libnss3-tools libnss3-dev clang
+    - name: Setup
+      run: |
+        autoreconf -fiv
+        ./configure
+    - uses: vapier/coverity-scan-action@v1
+      name: Coverity Scan
+      with:
+        project: "PKCS%2311+Provider"
+        email: ${{ secrets.COVERITY_SCAN_EMAIL }}
+        token: ${{ secrets.COVERITY_SCAN_TOKEN }}


oerdnj:
Here's the test build on my fork: https://github.com/oerdnj/pkcs11-provider/runs/7497965977?check_suite_focus=true

simo5:
And how do I set up the secrets?

oerdnj:
Just a remark: I think I am done with the basic infrastructure setup. There are couple of things that we can do, but that's for later:
* test coverage - coveralls.io, codecov.io, or just even running gcov and recording the output
* LGTM.com - this is a quite helpful static analysis tool, but their infrastructure is still running on Ubuntu 20.04, so neither Autoconf 2.71 nor OpenSSL 3.0

FTR: I would be happy to help with the reviews if needed.

oerdnj:
> And how do I set up the secrets?

Settings -> Secrets -> Actions and click on [New repository secret]

oerdnj:
Screenshot 2022-07-25 at 13 01 30

simo5:
> > And how do I set up the secrets?
> 
> Settings -> Secrets -> Actions and click on [New repository secret]

Yes I know how to technically set the secrets in github :-D

The questions was more like:
- where do they come from?
- how do I get to know what to put in there ?

oerdnj:
Oh, it's in the commit message:
```
    This adds new action that runs the Coverity Scan on every push to main
    branch.  This requires to set two GitHub Secrets:

      * COVERITY_SCAN_EMAIL: one of the maintainers
      * COVERITY_SCAN_TOKEN: can be found in the project settings in the
        https://scan.coverity.com interface
```

simo5:
So I assume it means I need to create an account with coverity?
Yet another one ... I'll have to read the ToC...

oerdnj:
Here's the token:

```
-----BEGIN PGP MESSAGE-----

hQEMA+RgWMrBh3CJAQgAirNJ8K58wyHpBCykGhUES0p8Py87RW45T4kI1mw8gg+I
fQfE4EhQeP8P4lyX+OOgKL2XERytMHdXa/9R2j5TAEixO3KjbfSMPqs9wIiWPpww
TDVs2S8eU1egl7qPeefnwmSMvR79MsnfFqxevZVSIjWyPoT3qCszcYT1+cPs20uW
C/ADJgDA8yqHovpiiuYZJ/HDkl0RT6oUOB2idfvA2j+yW57VqoKz9JK4MbpQEEUz
T78llVNyBY7v/FB7433qUnLXXlXuvSCT0XXfuXoUsTM2r6scO2IpzBgn8DMYcrEo
vojwZId9+wclpH71RBk2Z4dgJ7zT8qZDV9kMzc+mN9JSAeE9p/y7Hc+F+xcmeQRA
ZhGTzZKLpGV/1HxnDOnchtfPCRsD6bHHS0QVZU8xIDoeCCll3u+dDdKymc6AtCtB
JwYq1EBEePylWhmIqGGJLdnEhg==
=gTGF
-----END PGP MESSAGE-----
```

simo5:
Pretty strict limits (4 builds/day, 28/month), but I see that you added it to the build only when we push a PR to main.

Is there a way to distinguish between issues found and rejected build?
Is there a way to trigger this only when files in src/*.c are changed?
No point in consuming builds when I change a .md file in the root or similar.

oerdnj:
You don't have to create an account - it's possible just to receive the emails from the tool, but the feature seems to be broken right now.

simo5:
Turns out I already had an old account :-D
The secrets are set up, now we need to find out how to minimize builds to only when needed

oerdnj:
> Is there a way to distinguish between issues found and rejected build?

It's asynchronous, so it does not affect the GitHub Action success.  The people in the project will receive an email with new issue found / fixed.

> Is there a way to trigger this only when files in src/*.c are changed?
> No point in consuming builds when I change a .md file in the root or similar.

I've changed the `on:` to run the scan on every Sunday.

simo5:
Uhm I think we can run it once per each working day actually.
That sounds like a decent pace, and given the limit is 28 per month it should be fine

oerdnj:
> Uhm I think we can run it once per each working day actually.
> That sounds like a decent pace, and given the limit is 28 per month it should be fine

Sure, that's probably reasonable now and when the project is (semi-)stable switch back to the push on the `main` branch because that will be probably less frequent than once per day.

simo5:
I wonder if it make sense to run a git log --since=1day before the invcation of the scan action and bail out if there were no changes ...

simo5:
Something around:
CHANGES=`git log --since=1day`
if [ $CHANGES -eq 0 ]; then ~skip~; fi

PR #33, opened 2022-07-26
Fix issues found by the last Coverity Scan check

oerdnj:


oerdnj:
> [Modify the key pointers in find_keys() only on success](https://github.com/latchset/pkcs11-provider/pull/33/commits/8bad77826a8dbdc0887af3179e3509c37f02873f)

This could be also fixed by clearing the public and private key pointers on error in the `p11prov_store_load()`, but I think it's cleaner to only touch the dereferenced pointers only on success.

PR #41, opened 2022-07-29
Fix PIN wiping in few places

fabled:


src/util.c
```diff
@@ -493,10 +493,13 @@ static CK_RV token_login(P11PROV_CTX *provctx, CK_SLOT_ID slotid,
...
             P11PROV_raise(provctx, retc, "Failed to close session %lu", s);
         }
-        return ret;
+        goto done;
     }
 
-    return p11prov_ctx_set_login_session(provctx, s);
+    ret = p11prov_ctx_set_login_session(provctx, s);
+done:
+    OPENSSL_cleanse(pin, pinlen);
```
simo5, line 501: This will destroy the pin stored in the parsed_uri object or the main configuration context.
You should call this within the if (pw_cb) { } section only. Instead of in a common "done" at the end.

src/provider.c
```diff
@@ -139,6 +140,10 @@ static void p11prov_ctx_free(P11PROV_CTX *ctx)
         }
     }
 
+    if (ctx->pin) {
```
simo5, line 143: no need for the if here, openssl handles NULL pointers correctly

simo5 (changes requested):
Need to fix the pin destroying issue.

src/util.c
fabled: Fixed.

src/provider.c
fabled: Yes, but it calls `strlen` which is not guaranteed to handle NULL.

simo5: uh?

simo5: oh, doh!

simo5 (approved):

PR #47, opened 2022-08-12
Fix issues found by valgrinding test suite

fabled:
Various issues fixed found by valgrind during test suite running.

### DIFF
--- a/src/provider.c
+++ b/src/provider.c
@@ -607,7 +607,7 @@ static int refresh_slots(P11PROV_CTX *ctx)
         goto done;
     }
 
-    slots = OPENSSL_malloc(nslots * sizeof(struct p11prov_slot));
+    slots = OPENSSL_zalloc(nslots * sizeof(struct p11prov_slot));
     if (slots == NULL) {
         OPENSSL_free(slotid);
         ret = -ENOMEM;
--- a/src/store.c
+++ b/src/store.c
@@ -237,7 +237,8 @@ static CK_RV p11prov_store_ctx_add_key(void *pctx, CK_OBJECT_CLASS class,
 
     if ((ctx->num_objs % OBJS_ALLOC_SIZE) == 0) {
         P11PROV_OBJ **tmp =
-            OPENSSL_realloc(ctx->objects, ctx->num_objs + OBJS_ALLOC_SIZE);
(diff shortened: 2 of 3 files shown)


src/store.c
```diff
@@ -237,7 +237,7 @@ static CK_RV p11prov_store_ctx_add_key(void *pctx, CK_OBJECT_CLASS class,
 
     if ((ctx->num_objs % OBJS_ALLOC_SIZE) == 0) {
         P11PROV_OBJ **tmp =
-            OPENSSL_realloc(ctx->objects, ctx->num_objs + OBJS_ALLOC_SIZE);
+            OPENSSL_realloc(ctx->objects, sizeof(P11PROV_OBJ*[ctx->num_objs + OBJS_ALLOC_SIZE]));
```
simo5, line 240: I do not fully understand this, we are just creating an array of pointers here.
I never seen this casting within a sizeof, it confuses me.

Sounds to me this would be more understandable:
 (ctx->num_objs + OBJS_ALLOC_SIZE) * sizeof(P11PROV_OBJ *)
Ie:
OPENSSL_realloc(ctx->objects, (ctx->num_objs + OBJS_ALLOC_SIZE) * sizeof(P11PROV_OBJ *));

simo5 (changes requested):
Thanks, these are good fixes

src/store.c
fabled: `sizeof` takes size of the type. And the cast looking thing is type definition of the array we are instantiating. For me this is more readable and safe (no parenthesis required, taking size of type instead of calculating it), than doing the multiplication manually, but I'll change it.

src/provider.c
```diff
@@ -607,7 +607,7 @@ static int refresh_slots(P11PROV_CTX *ctx)
         goto done;
     }
 
-    slots = OPENSSL_malloc(nslots * sizeof(struct p11prov_slot));
+    slots = OPENSSL_zalloc(nslots * sizeof(struct p11prov_slot));
```
fabled, line 610: On second look, this might not be enough? Seems `refresh_slot_profiles` can be called again on same slot. I'll fix this also.

fabled: Actually no code paths call it again. The name was just slightly misleading. I'll leave it as it.

simo5 (approved):

PR #48, opened 2022-08-26
Session object pooling

### LINKED ISSUE #32: Implement session pooling or caching
Current strategy is to create PKCS11 sessions on the spot when needed. However is not ideal from interoperability and performance:
 - Some modules may have hard limit on few or even one concurrent session. In this case the first thread gets the sessions, and other threads will fail with failure to create session.
 - Other modules may support indefinite sessions, but have large CPU/time overhead on creation the session (e.g. session maps directly to a TCP connection). Thus using cache or pool of sessions could remove this overhead.
 - Also, as further optimization, if logging in to session is needed to use the key/object, we can remove need for extra logins for consecutive use of object if session is kept in logged-in state.

Additional potential PKCS11 complication to account for is that all sessions share the logged-in user and state. Thus some operations might be possible if there are sessions using from logged-in user (e.g. some thread doing crypto as normal user, but another thread wanting to create/delete objects needing SO user).

This ticket is to discuss the strategy to manage sessions and concurrent access.

Two options to consider:
 - Implement session pooling similar to [libp11's pooling](https://github.com/OpenSC/libp11/commit/4bd92da60a670e3c25d9cb69fca3363f47a225c7). Down side is the need for locking even in single threaded use cases.
 - Using per-thread cached session in pthread "setspecific" area. These can have a destructor associated so sessions can be cleaned up on thread exit. Additionally C_CloseAllSessions can be used to flush them globally if needed.

Perhaps additional options brew up during discussion of the problem.

simo5:
This PR implement a session object as well as pooling of session objects.

The design reuses allocated objects to minimize memory allocation on reuse, it also permits referencing of session objects multiple times. No code uses this currently but I can see it done when the code currently tries to duplicate sessions perhaps. This is a detail that will need discussion.

freeing a session object causes locking to be used in order to move a session object from the busy_sessions list to the free_sessions list. This is something we may want to avoid, so needs discussion.

Fixes #32 

- [x] tests for concurrent use of sessions

### DIFF
--- a/src/provider.c
+++ b/src/provider.c
@@ -31,7 +31,7 @@ struct p11prov_ctx {
     void *dlhandle;
     CK_FUNCTION_LIST *fns;
 
-    CK_SESSION_HANDLE login_session;
+    P11PROV_SESSION *login_session;
 
     int nslots;
     struct p11prov_slot *slots;
@@ -103,8 +103,7 @@ CK_UTF8CHAR_PTR p11prov_ctx_pin(P11PROV_CTX *ctx)
 }
 
 /* the login_session functions must be called under lock */
-CK_RV p11prov_ctx_get_login_session(P11PROV_CTX *ctx,
-                                    CK_SESSION_HANDLE *session)
+CK_RV p11prov_ctx_get_login_session(P11PROV_CTX *ctx, P11PROV_SESSION **session)
 {
     if (!ctx->is_locked) {
         /* called without mutex held */
@@ -117,7 +116,7 @@ CK_RV p11prov_ctx_get_login_session(P11PROV_CTX *ctx,
     return CKR_OK;
 }
 
-CK_RV p11prov_ctx_set_login_session(P11PROV_CTX *ctx, CK_SESSION_HANDLE session)
+CK_RV p11prov_ctx_set_login_session(P11PROV_CTX *ctx, P11PROV_SESSION *session)
 {
     if (!ctx->is_locked) {
         /* called without mutex held */
@@ -126,8 +125,8 @@ CK_RV p11prov_ctx_set_login_session(P11PROV_CTX *ctx, CK_SESSION_HANDLE session)
         return CKR_MUTEX_NOT_LOCKED;
     }
 
-    if (ctx->login_session != CK_INVALID_HANDLE) {
-        (void)ctx->fns->C_CloseSession(ctx->login_session);
+    if (ctx->login_session) {
+        p11prov_session_free(ctx->login_session);
     }
 
     ctx->login_session = session;
@@ -144,8 +143,8 @@ static void p11prov_ctx_free(P11PROV_CTX *ctx)
     OSSL_LIB_CTX_free(ctx->libctx);
 
     /* TODO: C_CloseAllSessions ? */
-    if (ctx->login_session != CK_INVALID_HANDLE) {
-        (void)ctx->fns->C_CloseSession(ctx->login_session);
+    if (ctx->login_session) {
+        p11prov_session_free(ctx->login_session);
     }
 
     if (ctx->dlhandle) {
--- a/src/provider.c
+++ b/src/provider.c
@@ -31,35 +31,18 @@ struct p11prov_ctx {
     void *dlhandle;
     CK_FUNCTION_LIST *fns;
 
-    P11PROV_SESSION *login_session;
-
     int nslots;
     struct p11prov_slot *slots;
 };
 
-int p11prov_ctx_lock_slots(P11PROV_CTX *ctx, struct p11prov_slot **slots)
+int p11prov_ctx_get_slots(P11PROV_CTX *ctx, struct p11prov_slot **slots)
 {
     if (ctx->status != P11PROV_INITIALIZED) {
         return RET_OSSL_ERR;
     }
 
-    pthread_mutex_lock(&ctx->lock);
-    ctx->is_locked = true;
-
     *slots = ctx->slots;
     return ctx->nslots;
-}
-
-void p11prov_ctx_unlock_slots(P11PROV_CTX *ctx, struct p11prov_slot **slots)
-{
-    if (ctx->status != P11PROV_INITIALIZED) {
-        return;
-    }
-
-    *slots = NULL;
-
-    ctx->is_locked = false;
-    pthread_mutex_unlock(&ctx->lock);
 }
 
 OSSL_LIB_CTX *p11prov_ctx_get_libctx(P11PROV_CTX *ctx)
@@ -102,37 +85,6 @@ CK_UTF8CHAR_PTR p11prov_ctx_pin(P11PROV_CTX *ctx)
     return (CK_UTF8CHAR_PTR)ctx->pin;
 }
 
-/* the login_session functions must be called under lock */
-CK_RV p11prov_ctx_get_login_session(P11PROV_CTX *ctx, P11PROV_SESSION **session)
-{
-    if (!ctx->is_locked) {
-        /* called without mutex held */
-        P11PROV_raise(ctx, CKR_MUTEX_NOT_LOCKED,
-                      "Tried to get login session without lock");
-        return CKR_MUTEX_NOT_LOCKED;
-    }
-
-    *session = ctx->login_session;
-    return CKR_OK;
-}
-
-CK_RV p11prov_ctx_set_login_session(P11PROV_CTX *ctx, P11PROV_SESSION *session)
-{
-    if (!ctx->is_locked) {
-        /* called without mutex held */
-        P11PROV_raise(ctx, CKR_MUTEX_NOT_LOCKED,
-                      "Tried to get login session without lock");
-        return CKR_MUTEX_NOT_LOCKED;
-    }
-
-    if (ctx->login_session) {
-        p11prov_session_free(ctx->login_session);
-    }
-
-    ctx->login_session = session;
-    return CKR_OK;
-}
-
 static void p11prov_ctx_free(P11PROV_CTX *ctx)
 {
     if (ctx->status != P11PROV_UNINITIALIZED) {
@@ -142,20 +94,18 @@ static void p11prov_ctx_free(P11PROV_CTX *ctx)
 
     OSSL_LIB_CTX_free(ctx->libctx);
 
-    /* TODO: C_CloseAllSessions ? */
-    if (ctx->login_session) {
-        p11prov_session_free(ctx->login_session);
-    }
-
     if (ctx->dlhandle) {
-        ctx->fns->C_Finalize(NULL);
-        dlclose(ctx->dlhandle);
-
         if (ctx->slots) {
+            for (int i = 0; i < ctx->nslots; i++) {
+                (void)p11prov_session_pool_free(ctx->slots[i].pool);
+            }
             OPENSSL_free(ctx->slots);
             ctx->slots = NULL;
             ctx->nslots = 0;
         }
+
+        ctx->fns->C_Finalize(NULL);
+        dlclose(ctx->dlhandle);
     }
 
     if (ctx->pin) {
@@ -529,7 +479,7 @@ static const OSSL_DISPATCH p11prov_dispatch_table[] = {
     { 0, NULL },
 };
 
-static int refresh_slot_profiles(P11PROV_CTX *ctx, struct p11prov_slot *slot)
+static int get_slot_profiles(P11PROV_CTX *ctx, struct p11prov_slot *slot)
 {
     CK_SESSION_HANDLE session;
     CK_BBOOL token = CK_TRUE;
@@ -590,46 +540,38 @@ done:
     return ret;
 }
 
-static int refresh_slots(P11PROV_CTX *ctx)
+static CK_RV get_slots(P11PROV_CTX *ctx)
 {
     CK_ULONG nslots;
     CK_SLOT_ID *slotid;
     struct p11prov_slot *slots;
     int ret;
 
-    if (ctx->status == P11PROV_INITIALIZED) {
-        pthread_mutex_lock(&ctx->lock);
-        ctx->is_locked = true;
-    }
-
     ret = ctx->fns->C_GetSlotList(CK_FALSE, NULL, &nslots);
     if (ret) {
-        goto done;
+        return ret;
     }
 
     /* arbitrary number from libp11 */
     if (nslots > 0x10000) {
-        ret = -E2BIG;
-        goto done;
+        return CKR_GENERAL_ERROR;
     }
 
     slotid = OPENSSL_malloc(nslots * sizeof(CK_SLOT_ID));
     if (slotid == NULL) {
-        ret = -ENOMEM;
-        goto done;
+        return CKR_HOST_MEMORY;
     }
 
     ret = ctx->fns->C_GetSlotList(CK_FALSE, slotid, &nslots);
     if (ret) {
         OPENSSL_free(slotid);
-        goto done;
+        return ret;
     }
 
     slots = OPENSSL_zalloc(nslots * sizeof(struct p11prov_slot));
     if (slots == NULL) {
         OPENSSL_free(slotid);
-        ret = -ENOMEM;
-        goto done;
+        return CKR_HOST_MEMORY;
     }
 
     for (size_t i = 0; i < nslots; i++) {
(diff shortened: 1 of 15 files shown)


simo5:
@fabled please take a good look at this one and provide feedback

fabled (commented):
I would have probably structured some things slightly differently as discussed earlier. However, I have no objection against this approach either.
I did add few review comments to places with some specific issues.

src/session.c
```diff
@@ -0,0 +1,449 @@
...
+        case CKR_OK:
+            break;
+        case CKR_SESSION_CLOSED:
+        case CKR_SESSION_HANDLE_INVALID:
+            /* try to open a session again */
+            sess->session = CK_INVALID_HANDLE;
+            ret = p11prov_session_open(sess, false, NULL, 0);
+            if (ret == CKR_OK) {
+                pthread_mutex_lock(&slot->sessions->lock);
+                LIST_INSERT(slot->sessions->busy_sessions, sess);
```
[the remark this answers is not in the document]
simo5: Good point, I do not particularly like this whole list management, I wonder if there is a better way...

src/session.c
```diff
@@ -0,0 +1,449 @@
...
+        check_session = true;
+    } else {
+        sess = p11prov_session_new(provctx, slot);
+        if (sess == NULL) {
+            ret = CKR_HOST_MEMORY;
+            goto unlock;
+        }
+    }
+    if (sess->session == CK_INVALID_HANDLE) {
+        ret = p11prov_session_open(sess, false, NULL, 0);
```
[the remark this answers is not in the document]
simo5: My concern is some form of deadlock if we do this, perhaps we should sleep for a "reasonable" time in case of session starvation and then still fail if a session does not come free in time.
The question is, what's a "reasonable" time?

src/session.c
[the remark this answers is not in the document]
simo5: worse that that, on failure it still returns the session to the caller without an error ...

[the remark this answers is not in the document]
simo5: However I do not see how 2 happens, given that in 2, if we do open a session we also set check_session to false. And w/o 2 there is no problem with LIST_INSERT() ...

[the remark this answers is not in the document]
simo5: on error paths we may also "lose" sessions as they are removed from the free_sessions list and not added back.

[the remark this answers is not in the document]
simo5: I really do not like this 2 lists business, I am going to change this whole thing using a free/busy int on the session and a linear list scan I think. It may be slightly less efficient when searching for a free session from time to time, but manipulations will not need locks, as a simple atomic xompare_exchange operation can be used to flip a session from free to busy and back.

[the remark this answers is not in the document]
fabled: This happens when trying to reuse an existing session handle, which has been closed/invalidated by the pkcs11 library for some reason.

src/session.c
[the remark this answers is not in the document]
simo5: I am thinking we could wait but only for a "reasonable" time, what is reasonable TBD.
I do not want to risk deadlocks, where one operation holds the last free session and waits on another that can't get a session. Probably rare, but could happen. So we need a way to timeout eventually. In any case it makes no sense to wait indefinitely on a single thread application that just happens to have exhausted all its sessions (via a resource leak for example).

simo5:
Ok, so I overhauled how sessions store session objects and how they are managed.
I think this version is much better than my v1, but there may still be gaps.

Please let me know what you think @fabled

simo5:
Btw I only changed the last 2 commits, previous stuff was not changed.

simo5:
@fabled any chance you can re-review the new connection pool code and comment?

src/session.c
```diff
@@ -7,33 +7,168 @@
...
+    /* deref login_session first */
+    p11prov_session_free(pool->login_session);
+
+    /* LOCKED SECTION ------------- */
+    pthread_mutex_lock(&pool->lock);
+    for (int i = 0; i < pool->num_sessions; i++) {
+        session = pool->sessions[i];
+        if (session->refcnt > 1) {
+            P11PROV_raise(pool->provctx, CKR_GENERAL_ERROR,
+                          "Can't free multiply-referenced session (%u)", i);
```
fabled, line 86: This still frees the object. Either add `continue`, or maybe adjust error to `Freeing still referenced session` to reflect the action taken better?

src/session.c
```diff
@@ -7,33 +7,168 @@
...
+    pthread_mutex_unlock(&pool->lock);
+    /* ------------- LOCKED SECTION */
+
+    pthread_mutex_destroy(&pool->lock);
+    OPENSSL_clear_free(pool, sizeof(P11PROV_SESSION_POOL));
+
+    return CKR_OK;
+}
+
+#define SESS_ALLOC_SIZE 8
```
fabled, line 118: Pointers are cheap in size, reallocation is slow. I'd make this at least 32 or 64.

src/session.c
```diff
@@ -43,99 +178,160 @@ P11PROV_SESSION *p11prov_session_ref(P11PROV_SESSION *session)
...
+        cur_sessions = __atomic_load_n(&p->cur_sessions, __ATOMIC_SEQ_CST);
+        if (cur_sessions > p->limit_sessions) {
+            CK_FUNCTION_LIST *f;
+            int ret;
+
+            f = p11prov_ctx_fns(session->provctx);
+            if (f == NULL) {
+                /* FIXME: should we abort, crash and burn here? */
+                return;
+            }
```
fabled, line 353: For me `p11prov_ctx_fns` should always succeed. `p11prov_ctx_fns` should assert / crash / panic if it can't return valid function list. This would clean up the code in various places.

src/session.c
```diff
@@ -178,56 +180,129 @@ static P11PROV_SESSION *p11prov_session_ref(P11PROV_SESSION *session)
...
+                ret = CKR_SESSION_COUNT;
+                cs = 0;
+                wait_ok = cyclewait_with_timeout(MAX_WAIT, SLEEP, &startime);
+                continue;
+            }
+        }
+
+        wait_ok = false;
+        ret = f->C_OpenSession(slot, CKF_SERIAL_SESSION, NULL, NULL, session);
+        if (ret == CKR_SESSION_COUNT) {
```
fabled, line 256: Should this also adjust the session pool limits if maximum was reached?
Or is this intentionally left not adjusting it, in case some other library is using the same token?

src/session.c
```diff
@@ -178,56 +180,129 @@ static P11PROV_SESSION *p11prov_session_ref(P11PROV_SESSION *session)
...
+        if (cs == 0) {
+            cs = __atomic_add_fetch(&p->cur_sessions, 1, __ATOMIC_SEQ_CST);
+            if (cs > p->max_sessions) {
+                (void)__atomic_sub_fetch(&p->cur_sessions, 1, __ATOMIC_SEQ_CST);
+                ret = CKR_SESSION_COUNT;
+                cs = 0;
+                wait_ok = cyclewait_with_timeout(MAX_WAIT, SLEEP, &startime);
+                continue;
+            }
+        }
```
fabled, line 252: Doing custom semaphores with sleep polling sounds prone to race conditions and hard to validate. It also means there's a level of busy polling with the process waking up a lot. So sleeping with wakup sounds better fit here.

Maybe consider using `pthread_cond_wait` and `pthread_cond_signal` together with mutex.
Alternatively this could be implemented as semaphore with `sem_post` and `sem_timedwait`.

src/session.c
```diff
@@ -232,20 +435,93 @@ CK_RV p11prov_get_session(P11PROV_CTX *provctx, CK_SLOT_ID *slotid,
...
+    }
+
+    if (session->session != CK_INVALID_HANDLE) {
+        /* check that the pkcs11 session is still ok */
+        CK_FUNCTION_LIST *f;
+        CK_SESSION_INFO session_info;
+
+        f = p11prov_ctx_fns(provctx);
+        if (f == NULL) {
+            ret = CKR_GENERAL_ERROR;
```
fabled, line 570: Most other code paths use `CKR_FUNCTION_FAILED` instead. But consider making it `p11prov_ctx_fns` contract  to return valid pointer or abort.

src/session.c
```diff
@@ -7,33 +7,168 @@
...
-P11PROV_SESSION *p11prov_session_new(P11PROV_CTX *ctx, CK_SLOT_ID slotid)
+struct p11prov_session_pool {
+    P11PROV_CTX *provctx;
+
+    CK_ULONG max_sessions;
+    CK_ULONG limit_sessions;
+    CK_ULONG cur_sessions;
+
+    P11PROV_SESSION **sessions;
+    int num_sessions;
```
fabled, line 29: `num_sessions` is slightly confusing with `cur_sessions`. The other `*_sessions` seem to reflect PKCS#11 session counts. But this seems to reflect the number of `P11PROV_SESSION` structs allocated. Maybe call it `num_p11sessions` or somehow differentiate it?

src/session.c
```diff
@@ -7,33 +7,168 @@
...
-    sess->session = CK_INVALID_HANDLE;
+    session->provctx = ctx;
+    session->slotid = slot->id;
+    session->session = CK_INVALID_HANDLE;
+    session->pool = p;
+    session->refcnt = 1;
+
+    /* LOCKED SECTION ------------- */
+    pthread_mutex_lock(&p->lock);
+
```
fabled, line 149: Maybe check here for `p->num_sessions >= p->max_sessions`. Otherwise due to races, it's possible allocate more `P11PROV_SESSION` objects than allowed.

src/session.c
```diff
@@ -7,33 +7,168 @@
...
 {
-    P11PROV_SESSION *sess;
+    P11PROV_SESSION_POOL *p = slot->pool;
+    P11PROV_SESSION *session;
+    CK_ULONG cur_sessions;
+
+    /* this is just to quit early and avoid work.
+     * Passing this check does not guarantee another thread is not
+     * racing to open the last avaliable session */
+    cur_sessions = __atomic_load_n(&p->cur_sessions, __ATOMIC_SEQ_CST);
```
fabled, line 129: Should this be `num_sessions` instead? This checks against pkcs11 session count which can get adjusted in various places. `num_sessions` is more monotonic on how many internal session objects we have.

src/session.c
```diff
@@ -178,56 +180,129 @@ static P11PROV_SESSION *p11prov_session_ref(P11PROV_SESSION *session)
...
+    return ret;
+}
+
+static CK_RV p11prov_session_open(P11PROV_SESSION *session, bool login,
+                                  CK_UTF8CHAR_PTR pin, CK_ULONG pinlen)
+{
+    P11PROV_SESSION_POOL *p = session->pool;
+    CK_RV ret;
+
+    ret = internal_session_open(p, session->slotid, &session->session);
```
fabled, line 273: I think this is wrong place for the waiting. It is possible that other threads exhausted the internal P11PROV_SESSION objects, and we don't even get here.
The right place for synchronizing/waiting would be `p11prov_get_session`.

src/session.c
```diff
@@ -232,20 +435,93 @@ CK_RV p11prov_get_session(P11PROV_CTX *provctx, CK_SLOT_ID *slotid,
...
     }
-    *session = sess;
-    return CKR_OK;
+
+    if (session == NULL) {
+        session = p11prov_session_new(provctx, slot);
+        if (session == NULL) {
+            return CKR_GENERAL_ERROR;
+        }
+    }
```
fabled, line 551: Somehow this feels racy. Walking the `session` list is done using the lock held, but it is released before `p11prov_session_new` call. I'd try to make the walking and inserting new object into one batch operation. Especially considering that a potential sleeping would be needed in this place. See the review message for next commit.

fabled (commented):
Looks already a lot better. I added more detailed commit messages this time. There's also additional points with the new synchronization functionality.

src/session.c
simo5: I thought about this a few time, but I *really* do not like when nested down in the bowel somewhere l;ibraries crash my application because they have a bug.
This can lave applications in bad states.
However what I could do is to mark the provider failed if there is ever an error like this and then refuse any further operation with it.

src/session.c
simo5: ok

src/session.c
```diff
@@ -7,33 +7,168 @@
...
+    pool = OPENSSL_zalloc(sizeof(P11PROV_SESSION_POOL));
+    if (!pool) {
+        return CKR_HOST_MEMORY;
+    }
+    pool->provctx = ctx;
+
+    pthread_mutex_init(&pool->lock, 0);
+
+    if (token->ulMaxSessionCount != 0) {
+        pool->max_sessions = token->ulMaxSessionCount;
```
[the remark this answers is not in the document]
simo5: makes sense

src/session.c
simo5: Yeah I do not want to unnecessarily reduce maximums, because the token may be used concurrently by multiple applications at some point in time so the failure may be spurious.
Thinking of a network HSM that have some quota on concurrent sessions based on source IP or something, and similar other HW limitations for directly attached HSMs.

We should never really hit this, because we also have a soft limit to try to avoid to hog all available sessions anyway.

src/session.c
simo5: I considered using pthread_cond_wait() and similar, but that only works when the application itself is the cause of the issue.
If the problem is a separate process temporarily keeping sessions busy, we wouldn't be able to handle it.

The problem with sem_post/sem_timedwait is that you need a semaphore, but cur_sessions is a counter with a variable maximum. I would have to revers cur_sessions to be something like cur_free_sessions, then perhaps we could use it as a semaphore ... I will look into it after I address the basic issues in this review.

I think it is ok to busy-poll in this case anyway.

src/session.c
simo5: See the other comment, I want to try to make the provider stop working entirely and return a CKR_GENERAL_ERROR everywhere, in fact I think I will change the prototype of the function to return an error and get a pointer to fns as an argument to fill in instead.

src/session.c
simo5: Sounds good

src/session.c
simo5: Not sure it is really worth it, at most we waste 64/256 bytes of data anyway.

src/session.c
simo5: No because internal object may not have sessions associated, we really want to check against the actually open sessions with the token.

src/session.c
simo5: This is safe p11prov_session_new() locks the pool on its own, and is done only if no free sessions were found.
It is true that a session may come free just as we exit the locked section and before entering p11prov_session_new(), but that changes nothing because if we moved p11prov_session_new() inside the locked block [hard because then we need to detect that and avoid locking inside_new()] then we'd also have created a new session regardless as nothing could free anything in the meanwhile. So bottom line we end up with a new session no matter what we do.

By keeping p11prov_session_new() outside of the locked block we simplify the locked sections and keep them tidy. That was one of guiding principle in changing this code. Reducing locked sections (which means reducing contention) is very important here if people are using Hardware Acceleration or HSMs.

src/provider.c
```diff
@@ -620,26 +562,30 @@ static int refresh_slots(P11PROV_CTX *ctx)
...
-        ctx->is_locked = false;
-        pthread_mutex_unlock(&ctx->lock);
+    if (ret != CKR_OK) {
+        for (size_t i = 0; i < nslots; i++) {
+            p11prov_session_pool_free(slots[i].pool);
+        }
+        OPENSSL_free(slots);
+    } else {
+        ctx->slots = slots;
+        ctx->nslots = nslots;
```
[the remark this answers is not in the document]
simo5: Yeah at the moment I chnged the code to only ever pull slots once at initialization.
Feel free to create an issue if we ever think we may have a way to reload slot info.
I am not sure we'll ever fully reload slot info.

src/session.c
simo5: Will add a way to orphan the entry and move on

src/session.c
simo5: See #53 for my take on this.

src/session.c
simo5: See #53

simo5:
Rebase on top of the new p11prov_ctx_status() change and added a test program to check that multiple sessions are kept open properly and concurrently.

Manual inspection of the debug file also confirms we close sessions when we have more than soft-limit open, and stop closing them once we reach again the limit.

src/session.c
simo5: So after some more thinking and seeing if I can use sem_post and sem_timedwait I realized that sem_timedwait does not let you chose a monotonic clock, which makes it vulnerable to clock changes.
I think I will stick with my crude implementation for now, given there are tests that prove it works.

simo5:
@fabled can you provide a final review?
I think this code is good to be merged at this point.

fabled (commented):
Looks pretty good. I'd still prefer if the synchronization location is changed. But that's perhaps something that could be addressed later too.

src/session.c
```diff
@@ -0,0 +1,575 @@
...
+        /* keep a max of 10% of the sessions */
+        pool->limit_sessions = pool->max_sessions / 10;
+        if (pool->limit_sessions == 0) {
+            pool->limit_sessions = 3;
+        }
+    } else {
+        /* arbitrary max concurrent open sessions */
+        pool->max_sessions = 1024;
+        pool->limit_sessions = 8;
```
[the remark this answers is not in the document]
simo5: I am not sure why we want to tie that to the number of CPUs?
TBH, I pulled this values more or less at random, I think 1024 is way too high as a default maximum, but 8 sounded like a decent number of concurrent sessions to have, enough to have a login session, a couple of keys and perform 2,3 operations in parallel and still have a session to spare.
If we want to tie this to the CPU I would think we'd have to double the number of CPUs + 1 (login session), but then again I am not sure why we'd do that, I have no data to know how many concurrent session a single application would normally use.

I thought that in future we could keep some statistics and self tune.
This code *needs* to be improved because on a test run I did, reading the logs, I discovered we are not using the session objects with an actually cached session in preference, instead we just pick the first free and opena  new session if one is not available.

So this code definitely needs more smarts when it is looking for a free session.

src/session.c
simo5: Uhmm, I guess we can avoid the check in session_new then, however should we have an upper limit to p11prov_sessions too ?
I do not think it make sense to have too many of them way beyond the max number of possible token sessions. they will give a lot of "free" sessions that just keep failing with a timeout later if no sessions are being released, or the token somehow malfunctioned and decided to not give out sessions anymore.

src/session.c
simo5: Yeah this ties in to the other comment I just left.
Should we have a limit for P11PROV_SESSIONS? Or should we allow a runaway thread to just allocate as many as it wants?
Perhaps we should allow just a handful more than max_sessions ?

simo5:
@fabled can you check the last commit, I think it should address your concerns as well as my finding about reusing actual cached token sessions.

simo5:
I think any additional comment can be addressed in a followup, so merging.

PR #49, opened 2022-08-31
Add required macros file

pemensik:
Fedora Rawhide requires this file to configure inside repository.
Otherwise steps in BUILD.md are not enough.


simo5:
This is actually not needed specifically for Fedora, you just need to:
# dnf install autoconf-archive

The file is then available as such:
$ rpm -ql autoconf-archive |grep ax_check_compile
/usr/share/aclocal/ax_check_compile_flag.m4

I think Debian has a similar archive.

pemensik:
ah, okay. It is even mentioned in BUILD.md. My mistake, sorry. Closing this.

simo5:
No problem,
thanks for sending patches instead of just opening issues.

Always appreciated :-)

PR #50, opened 2022-09-01
Fail hard make check if nss-softokn devel files were not found

pemensik:
The test would fail if softokn has wrong path detected. It is required
just for testing, not a normal build. Add a simple check to makefile to
indicate a working path is expected.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -8,6 +8,8 @@ testsdir=@abs_top_builddir@/tests
 dist_check_SCRIPTS = \
 	test.sh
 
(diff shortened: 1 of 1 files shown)


simo5 (commented):
The approach is fine, but it would be nice if the test was silent on success, currently the whole line is printed on output ...

pemensik:
Modified it to actually pass just make all without it, but complain on make check. Checks just file presence by make as dependency.

pemensik:
Ok, modified

simo5 (approved):

PR #56, opened 2022-09-11
Rsa keygen

simo5:
Try to implement RSA key generation function as best we can.

openssl genpkey assumes that you *always* have key data to return, which is obviously not the case when generateing a key on a token.

We provide a function to print the -text option, but errors are reported from the missing pem/der encoding functions although public and private keys are generated and stored in the DB.

A separate test binary checks that we can correctly generate keys and apply ID and Label to them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ tests/openssl.cnf
 tests/tokens
 tests/tmp
 tests/tsession
+tests/tgenkey
 tests/*.log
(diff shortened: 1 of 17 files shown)


simo5:
@fabled I gave another spin to the old rsa_keygen patch, this one is based on the session_obj PR

simo5:
I think this is good enough for now, merging

PR #59, opened 2022-09-16
Add code to list and debug token mechanisms

simo5:
This provides a complete list of all the supported mechanisms.

It also select the available operations based on the token stated mechanism availability.


simo5:
@fabled do we need fallback code for this?
I am not sure all actual tokens export mechanism info ...

simo5:
@fabled ok then, should we just merge as is?

fabled:
Yes, looks good to me. Thanks!

PR #61, opened 2022-09-21
Add RSA-PSS support

### LINKED ISSUE #60: Add RSA PSS support
Some infrastructure is already present but there are no keymgmt or signature operations exported.

simo5:
Fixes #60 

Also fixes an off by one bug in the store.

- [x] partial restriction checks on parametrers that openssl expects
Not all parameters can be handled, and even those we can most tokens will not be able to for now.

### DIFF
--- a/src/asymmetric_cipher.c
+++ b/src/asymmetric_cipher.c
@@ -121,7 +121,7 @@ static int p11prov_rsaenc_encrypt_init(void *ctx, void *provkey,
         return RET_OSSL_ERR;
     }
 
-    encctx->key = p11prov_object_get_key(obj, 0);
+    encctx->key = p11prov_object_get_key(obj);
     if (encctx->key == NULL) {
         return RET_OSSL_ERR;
     }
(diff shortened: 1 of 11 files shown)


simo5:
So unfortunately NSS softoken currently does not support any PSS related restriction.
Keys are all generated equal for RSA.

There is a way to set attributes on key objects that I will implement but keep disabled until we can test.

The CKA_ALLOWED_MECHANISMS attribute on a key can specify an array of mechanism that are allowed for a key, using this we would be able to specify only _PSS mechanism and even only just specific digests. Unfortunately we cannot use this by default as it would caue a key template to be rejected with invalid arguments (or in some cases simply ignored, depending on pkcs11 driver/token implementation).

I will add basically dead code that can set a list of algorithm as an additional operation after key generation, but will probably end up ifdeffing it out for now to avoid unnecessary errors in logs. Unfortunately I am not aware of a way to detect if this attribute is supported on key objects by a token, so it cannot be easily conditionalized.

simo5:
@fabled in this PR I am adding a generic mechanism to register quirks, let me know what you think.
Otherwise this PR is good to be merged as I implemented all I could muster for now.

simo5:
I am happy with the status of this work, merging.

PR #67, opened 2022-10-11
WIP: Remove the use of custom operation names

simo5:
This is  in conjunction with experimental code I have in a custom OpenSSL build that prefers to stick to the provider that originates keys, countrary to current OpenSSL behavior that ends up always selecting the 'default' provider even if keys came from a completely different one.

~~This depends on OpenSSL changes (https://github.com/openssl/openssl/pull/19395) in current form. If those are declined we'll have to change the test-suite to alwyas pass "provider=pkcs11" to each operation.~~

Thanks to the discussion on the openssl PR I found out a single line change (first commit now) that, single-handedly rectifies the situation


simo5:
Note that CI failure is expected because CI does not have a patched OpenSSL.

simo5:
Funnily enough this works locally but now fails in CI w/o enough info, awesome :-)

simo5:
CI now failing because the code we depend on was broken until OpenSSL 3.0.5

PR #77, opened 2022-10-24
Fedora package and requirements clarifications

Jakuje:
While getting familiar with this package, I prepared a spec file and build the packages in copr:

https://copr.fedorainfracloud.org/coprs/jjelen/pkcs11-provider/monitor/

From here, it should be easy to integrate the packit to automate the builds.

The tests now fail on s390x on signature verification test on all Fedora versions, but I was not able to investigate this further yet:

```
## Sign and Verify with provided Hash and RSA
openssl dgst -sha256 -binary -out ${TMPPDIR}/sha256.bin ${SEEDFILE}
openssl pkeyutl -sign -inkey "${PRIURI}" -in ${TMPPDIR}/sha256.bin -out ${TMPPDIR}/sha256-sig.bin
openssl pkeyutl -verify -inkey "${PUBURI}" -pubin -in ${TMPPDIR}/sha256.bin -sigfile ${TMPPDIR}/sha256-sig.bin
000003FFB0772720:error:0200008A:rsa routines:RSA_padding_check_PKCS1_type_1:invalid padding:crypto/rsa/rsa_pk1.c:75:
000003FFB0772720:error:02000072:rsa routines:rsa_ossl_public_decrypt:padding check failed:crypto/rsa/rsa_ossl.c:599:
Signature Verification Failure
FAIL test.sh (exit status: 1)
```
If anyone has some ideas where to look, feel free to point me in the right direction or fix that. Otherwise I will open an issue later.

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -20,6 +20,7 @@ Files: **/Makefile.am
        tests/openssl.cnf.in
        .clang-format
        .clang-format-ignore
+       packaging/pkcs11-provider.spec
 Copyright: (C) 2022 Simo Sorce <simo@redhat.com>
 License: Apache-2.0
 
--- a/BUILD.md
+++ b/BUILD.md
@@ -2,8 +2,9 @@
 
 This package requires the following:
 - OpenSSL 3.0+ libraries and development headers
-- autoconf-archives packages for some m4 macros
-- NSS softoken and development headers (for testing)
+- OpenSSL tools (for testing)
+- autoconf-archive packages for some m4 macros
+- NSS softoken, tools and development headers (for testing)
 - a C compiler that supports at least C11 semantics
 
 The usual command to build are:
--- a/packaging/pkcs11-provider.spec
+++ b/packaging/pkcs11-provider.spec
@@ -1,0 +1,58 @@
+Name:          pkcs11-provider
+Version:       0.1
+Release:       1%{dist}
+Summary:       A PKCS#11 provider for OpenSSL 3.0+
+License:       Apache-2.0
+URL:           https://github.com/latchset/pkcs11-provider
+Source:        pkcs11-provider-%{version}.tar.gz
+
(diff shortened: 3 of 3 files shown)


simo5:
@Jakuje you may want to check if byteswap_buf() is doing the right thing on s390x ...

simo5:
To fix the REUSE check just add the file in the appropriate list in .reuse/dep5

Jakuje:
> @Jakuje you may want to check if byteswap_buf() is doing the right thing on s390x ...

Can you clarify where did you learn that the `Tokens return data in bigendian order` for RSA keys? I do not see any note about this in PKCS#11 specs.

> To fix the REUSE check just add the file in the appropriate list in .reuse/dep5

Added. Thanks.

One more concern I have is that the filename in `%{_libdir}/ossl-modules/` should probably be just `pkcs11.so` to allow loading it from inside of OpenSSL as a provider with name `pkcs11` when this shortcut is not made inside of the `openssl.cnf`, what do you think?

simo5:
> > @Jakuje you may want to check if byteswap_buf() is doing the right thing on s390x ...
> 
> Can you clarify where did you learn that the `Tokens return data in bigendian order` for RSA keys? I do not see any note about this in PKCS#11 specs.

I think it was just experimenatation with OpenSSL and the NSS softoken, I do not know if what I did is the right thing, or just happened to work on little endian.

> 
> > To fix the REUSE check just add the file in the appropriate list in .reuse/dep5
> 
> Added. Thanks.
> 
> One more concern I have is that the filename in `%{_libdir}/ossl-modules/` should probably be just `pkcs11.so` to allow loading it from inside of OpenSSL as a provider with name `pkcs11` when this shortcut is not made inside of the `openssl.cnf`, what do you think?

I think this makes sense, if we stick the provider in a specific directory there is no point to carry the "provider" part of the name.

PR #79, opened 2022-10-26
Add different ways for specifying PKCS#11 module to use

Jakuje:
Configuration of the provider through the openssl configuration file is cumbersome and it prevents also dynamically loading the provider (or makes it very complicated, see openssl/openssl#19496). Especially for testing, it would be very useful to have a way to pass environment variable with the testing pkcs11 module to use.

For generic use and integration in the consistent pkcs11 in os, the provider should try to pick up the default system pkcs11 module, which is in most of the cases p11-kit. This adds it as a default fallback option through `--with-default-pkcs11-module=yes`, which will ask p11-kit for this path through the pkg_config.

This also adds a implicit failure in case of no module was specified in any of the above means, which leads to dlopen call with NULL argument opening the calling so file, which does not have any `C_GetFunctionList`, leading to hard-to-parse errors like:

    prog: undefined symbol: C_GetFunctionList

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -48,6 +48,13 @@ PKG_CHECK_MODULES(
 )
 AC_SUBST([SHARED_EXT], $(eval echo "${shrext_cmds}"))
 
+# Check whether we have a p11-kit to use as a default PKCS#11 module
+PKG_CHECK_EXISTS([p11-kit-1],
+                 [PKG_CHECK_VAR([DEFAULT_PKCS11_MODULE],
+                                [p11-kit-1],
+                                [proxy_module])],
+                 [AC_MSG_WARN([The P11-kit proxy is not available. No fallback PKCS11 module used.])])
+
 # Try nss-softoken first as used on Fedora,
 # fallback to "nss" as used on Debian
 PKG_CHECK_EXISTS(
--- a/src/provider.c
+++ b/src/provider.c
@@ -23,7 +23,7 @@ struct p11prov_ctx {
     OSSL_LIB_CTX *libctx;
 
     /* Configuration */
-    const char *module;
+    char *module;
     const char *init_args;
     char *pin;
     /* TODO: ui_method */
@@ -299,6 +299,11 @@ static void p11prov_ctx_free(P11PROV_CTX *ctx)
 
     if (ctx->pin) {
         OPENSSL_clear_free(ctx->pin, strlen(ctx->pin));
+    }
+
+    if (ctx->module) {
+        OPENSSL_free(ctx->module);
+        ctx->module = NULL;
     }
 
     if (ctx->quirks) {
@@ -1045,10 +1050,18 @@ static int p11prov_module_init(P11PROV_CTX *ctx)
         .pReserved = (void *)ctx->init_args,
     };
     CK_INFO ck_info = { 0 };
+    const char *env_module;
     int ret;
 
     if (ctx->status != P11PROV_UNINITIALIZED) {
         return 0;
+    }
+
+    /* The environment variable has the highest precedence */
+    env_module = getenv("PKCS11_PROVIDER_MODULE");
+    if (env_module && *env_module) {
+        OPENSSL_free(ctx->module);
+        ctx->module = OPENSSL_strdup(env_module);
     }
 
     P11PROV_debug("PKCS#11: Initializing the module: %s", ctx->module);
@@ -1117,6 +1130,7 @@ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
                        const OSSL_DISPATCH **out, void **provctx)
 {
     OSSL_PARAM core_params[4] = { 0 };
+    const char *module = NULL;
     char *pin = NULL;
     P11PROV_CTX *ctx;
     int ret;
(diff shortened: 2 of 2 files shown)


src/provider.c
```diff
@@ -1139,7 +1147,7 @@ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
 
     /* get module path */
     core_params[0] = OSSL_PARAM_construct_utf8_ptr(
-        P11PROV_PKCS11_MODULE_PATH, (char **)&ctx->module, sizeof(ctx->module));
+        P11PROV_PKCS11_MODULE_PATH, &ctx->module, sizeof(ctx->module));
```
simo5, line 1160: This will not copy the path I think, only reference the internal storage, so you need to copy it after the fact I think, otherwise the free() at line 1058 may break things

src/provider.c
```diff
@@ -1045,12 +1045,20 @@ static int p11prov_module_init(P11PROV_CTX *ctx)
...
 
     if (ctx->status != P11PROV_UNINITIALIZED) {
         return 0;
     }
 
+    /* The environment variable has the highest precedence */
+    env_module = getenv("PKCS11_PROVIDER_MODULE");
+    if (env_module && *env_module) {
+        free(ctx->module);
+        ctx->module = strdup(env_module);
```
simo5, line 1059: please use OPENSSL_strdup() and add an OPENSSL_free() of the module parameter in the providered cleanup function.

configure.ac
```diff
@@ -8,6 +8,37 @@ AC_CONFIG_HEADERS([src/config.h])
...
+               AC_CHECK_FILE("$PKCS11_PATH",
+                       [ AC_DEFINE_UNQUOTED([DEFAULT_PKCS11_MODULE],
+                                            ["$PKCS11_PATH"],
+                                            [Path to default PKCS#11 module])
+                       ],
+                       [ AC_MSG_ERROR([Requested PKCS11 provided not found]) ]
+               )
+       fi ]
+)
+
```
simo5, line 41: Looks over complicated, what is wrong in using directly something like:
```
PKG_CHECK_EXISTS(
        [p11-kit-1],
        [PKG_CHECK_VAR([...
```

See how this is done for nss-sotoken please.

src/provider.c
```diff
@@ -1059,6 +1059,16 @@ static int p11prov_module_init(P11PROV_CTX *ctx)
         ctx->module = strdup(env_module);
     }
 
+    /* If the module is not specified in the configuration file, use the p11-kit proxy  */
+    if (ctx->module == NULL) {
+#ifdef DEFAULT_PKCS11_MODULE
+        ctx->module = strdup(DEFAULT_PKCS11_MODULE);
```
simo5, line 1065: OPENSSL_strdup() here too.

src/provider.c
```diff
@@ -1059,6 +1059,16 @@ static int p11prov_module_init(P11PROV_CTX *ctx)
...
+#else
+        P11PROV_debug("No PKCS#11 module specified.");
```
simo5, line 1067: Perhaps this should be a P11PROV_raise() so that we return a meaningful error to the openssl stack as well.

simo5 (changes requested):

configure.ac
Jakuje: I wanted to give a chance to enable/disable this, but I think this should be elected by the build system by providing p11-kit or not so lets simplify this.

src/provider.c
Jakuje: Thanks. Changed.

src/provider.c
Jakuje: Good idea. Updated.

src/provider.c
Jakuje: You are probably right. This API is quite opaque so it is hard to guess what does it actually do, but this would make sense.

src/provider.c
```diff
@@ -1045,12 +1049,30 @@ static int p11prov_module_init(P11PROV_CTX *ctx)
...
+        OPENSSL_free(ctx->module);
+        ctx->module = OPENSSL_strdup(env_module);
+    }
+
+    /* If the module is not specified in the configuration file, use the p11-kit proxy  */
+    if (ctx->module == NULL) {
+#ifdef DEFAULT_PKCS11_MODULE
+        ctx->module = OPENSSL_strdup(DEFAULT_PKCS11_MODULE);
+#else
+        P11PROV_raise(ctx, -ENOENT, "No PKCS#11 module specified.");
```
simo5, line 1071: P!!PROV_raise take a CK_RV error which is then converted to txt via p11prov_get_reason_strings() by internal openssl dispatching.
So a (negative) errno is not ok, you should probably use  CKR_GENERAL_ERROR ... unless we want to start definining vendor specific errors and add them to the translation table.

src/provider.c
```diff
@@ -1152,6 +1174,14 @@ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
         p11prov_ctx_free(ctx);
         return ret;
     }
+    /* Copy the returned module path if there was one */
+    if (ctx->module != NULL) {
```
simo5, line 1178: Wonder if it would be more clear to use a const char *modconf local variable, and then copy it to ctx->module.

I am afraid the current code may still lead to an attempt to free memory we shouldn;t when openssl tries to unload the provider after this fatal error.

simo5 (changes requested):

src/provider.c
Jakuje: I am not a huge fan of const chars. In this case, it is not clear where it is stored and when it will stop existing. Is it bound to the openssl context? Provider context? Something else?

Lets use temporary variable, which should cover all the corner cases.

src/provider.c
Jakuje: Thanks. Overlooked this. Fixed with the current version.

simo5:
@Jakuje there is only one style check error, once you fix that I'll merge this.

Jakuje:
> @Jakuje there is only one style check error, once you fix that I'll merge this.

My bad. Should be fixed now.

simo5:
Thanks!

PR #90, opened 2022-11-07
Implement callback for tls group capabilities

simo5:
Fixes #89

PR #97, opened 2022-11-09
run tests also using SoftHSM

Jakuje:
This is mostly POC of running the tests against different PKCS#11 module with different issues/restrictions. There was more than I anticipated, but it looks like it works.

There are some quirks:
* The SoftHSM is built against the OpenSSL and some of the initialization or cleanup functions are called in wrong order, we can not use SoftHSM directly, but only through the remoting functionality of p11-kit to shield away these issues.
* The SoftHSM does not support all tested mechanisms now, which include HKDF and some key operations with SHA2 (OAEP/PSS, ECC). These are now skipped.


tests/Makefile.am
```diff
@@ -42,6 +47,7 @@ MAINTAINERCLEANFILES = \
 
 edit_cmd = $(SED) \
     -e 's|@softokenpath[@]|$(softokenpath)|g' \
+    -e 's|@p11kitclientpath[@]|$(p11kitclientpath)|g' \
```
simo5, line 50: if we use the same .in file we just need a "tokenpath" variable in the .in file and the edit script will replace it wth the "right" token in each oif the config files.

tests/test-softhsm.sh
```diff
@@ -0,0 +1,171 @@
...
+    export NOPSS="1"
+    export NOECCSHA2="1"
+    export NOOAEPSHA2="1"
+    export NODERIVE="1"
+
+
+    title LINE "Export variables to ${TMPPDIR}/debugvars for easy debugging"
+#TODO
+    BASEDIR=$(pwd)
+    cat > ${TMPPDIR}/debugvars <<DBGSCRIPT
```
simo5, line 140: probably better to have debugvars.softhsm ?
(and debugvars.softokn on the other file)

tests/test.sh
```diff
@@ -203,12 +11,10 @@ ossl()
...
-
-title PARA "Export Public key to a file"
-ossl 'pkey -in $BASEURI -pubin -pubout -out ${TSTCRT}.pub'
-title LINE "Export Public key to a file (pub-uri)"
-ossl 'pkey -in $PUBURI -pubin -pubout -out ${TSTCRT}.pub'
+#title PARA "Export Public key to a file"
+#ossl 'pkey -in $BASEURI -pubin -pubout -out ${TSTCRT}.pub'
+#title LINE "Export Public key to a file (pub-uri)"
+#ossl 'pkey -in $PUBURI -pubin -pubout -out ${TSTCRT}.pub'
 title LINE "Export Public key to a file (with pin)"
```
simo5, line 18: Why are these commented out now ?

tests/test-softhsm.sh
```diff
@@ -0,0 +1,171 @@
...
+    fi
+    eval $(p11-kit server --provider "$P11LIB" "pkcs11:")
+    echo " ----------------------------------------------------------------------------------------------------"
+    pkcs11-tool -O --login --pin=$PINVALUE --module="$P11KITCLIENTPATH"
+    echo " ----------------------------------------------------------------------------------------------------"
+
+    export PKCS11_PROVIDER_MODULE=$P11KITCLIENTPATH
+    #export PKCS11_PROVIDER_MODULE=/usr/lib64/pkcs11-spy.so
+    #export PKCS11SPY=$P11KITCLIENTPATH
+    export NOPSS="1"
```
simo5, line 131: I think positively named variables are more understable.

Can you add the following to the common.sh file:
TEST_RSAPSS="1"
TEST_ECC_SHA2="1"
TEST_OAEP_SHA2="1"
TEST_HKDF="1"

then override them in this file by setting them to "0" ?

tests/test.sh
```diff
@@ -287,71 +93,77 @@ pkeyutl -verify -inkey "${PUBURI}"
...
-pkeyutl -verify -inkey "${PUBURI}"
-                -pubin
-                -digest sha256
-                -pkeyopt pad-mode:pss
-                -pkeyopt mgf1-digest:sha256
-                -pkeyopt saltlen:digest
-                -in ${TMPPDIR}/64krandom.bin
-                -rawin
-                -sigfile ${TMPPDIR}/sha256-dgstsig.bin'
+if [ "$NOPSS" != "1" ]; then
```
simo5, line 96: From previous comment on test controlling variables this would become:
if [ "$TEST_RSAPSS" != "0" ]; then

Which I find a lot more readable and understandable then a "double negation".

.github/workflows/build.yml
```diff
@@ -38,13 +38,15 @@ jobs:
           if [ -f /etc/fedora-release ]; then
             dnf -y install git ${{ matrix.compiler }} automake libtool \
               pkgconf-pkg-config autoconf-archive openssl-devel openssl \
-              nss-softokn nss-tools nss-softokn-devel
+              nss-softokn nss-tools nss-softokn-devel softhsm opensc \
```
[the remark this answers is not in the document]
simo5: But the advantage is that tess will show immediately if one fails and the other not, without skipping further tests if the first of the two fails.

tests/test-softhsm.sh
Jakuje: Fine for me. I wanted more to keep this on an exception basis that by default everything works and there would be some exceptions to be skipped, but it turned out the softhsm has problems with more stuff.

tests/test-softhsm.sh
Jakuje: certainly good idea if they will run at the same time

.github/workflows/build.yml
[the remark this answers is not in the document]
Jakuje: afaik existing from the script with 77 will mark the test as skipped, which I tried to do on the softhsm side. Modifying the softokn side to do the same might allow us to do this.

src/exchange.c
```diff
@@ -182,6 +182,8 @@ static int p11prov_ecdh_init(void *ctx, void *provkey,
     P11PROV_OBJ *obj = (P11PROV_OBJ *)provkey;
     CK_RV ret;
 
+    P11PROV_debug("%S: called", __func__);
+
```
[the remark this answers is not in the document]
simo5: ok let's remove

tests/Makefile.am
```diff
@@ -1,7 +1,8 @@
-EXTRA_DIST = openssl.cnf.in
-noinst_DATA = openssl.cnf
+EXTRA_DIST = openssl.cnf.in openssl-softhsm.cnf.in test.sh common.sh
+noinst_DATA = openssl.cnf openssl-softhsm.cnf
```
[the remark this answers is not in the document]
simo5: that is also ok.

tests/test-softhsm.sh
simo5: In the end it seem reasonable to be able to skip any test based on what a token support, I envision adding test for opencryptoki and HW tokens once we have a way to test them, so fine control will be needed anyway I think.

tests/test.sh
Jakuje: I was debugging something and forgot to clean this up and it was hidden in the move from one file to the other. Sorry.

simo5:
@Jakuje you need to fix all the github yaml build scripts to pull out the correct log and debug files as you have changed places.
Currnt failures are missing from the uploaded artifacts because of that.

Jakuje:
> @Jakuje you need to fix all the github yaml build scripts to pull out the correct log and debug files as you have changed places.
Currnt failures are missing from the uploaded artifacts because of that.

Done as well. I did a bit reshuffling of the tmp directories as the tests can not share the temporary files in case they run in parallel.
 * this moved the `tokens` directory under the `tmp.softhsm` and `tmp.softokn` as it is indeed temporary and different for both software tokens
 * moved the pinfile directly into the `tests/` directory as it is "shared" now and referenced from the shared `openssl.cnf`
 * the path to the pkcs11 module is now set through environment variable. By default to softokn, but softhsm tests override this variable
 * 

What I was not sure if we agreed on I should change or not:
 * separate test matrix for sofhsm, softokn x all the currently existing configurations (fedora, debian, gcc, clang)
 * using the positive or negative names for skipping some tests

tests/test-softhsm.sh
holger-dengler: I think it is sufficient to use different tmp-dirs. I don't see any benefit to also use a module-specific ending for the debugvars file (or any other file).

holger-dengler:
As far as I can see, the generic `tests/test.sh`  script will only run properly, if it is called from either `test-softhsm.sh` or `test-softokn.sh`. For debugging purposes, it might be valuable, if it would be able to run the generic tests without do all the setup stuff. Or, if you want to switch between the two modules, just do a `source tests/tmp.<module>/debugvars; tests/test.sh`. But maybe we should cover this in a separate PR.

simo5:
> > @Jakuje you need to fix all the github yaml build scripts to pull out the correct log and debug files as you have changed places.
> > Currnt failures are missing from the uploaded artifacts because of that.
> 
> Done as well. I did a bit reshuffling of the tmp directories as the tests can not share the temporary files in case they run in parallel.
> 
>     * this moved the `tokens` directory under the `tmp.softhsm` and `tmp.softokn` as it is indeed temporary and different for both software tokens

This is fine.

>     * moved the pinfile directly into the `tests/` directory as it is "shared" now and referenced from the shared `openssl.cnf`

ok

>     * the path to the pkcs11 module is now set through environment variable. By default to softokn, but softhsm tests override this variable

Uhmmm, I guess this works, but I wonder if we shouldn't create two openssl.<module>.cnf files instead...

> 
> What I was not sure if we agreed on I should change or not:
> 
>     * separate test matrix for sofhsm, softokn x all the currently existing configurations (fedora, debian, gcc, clang)

Yes please.

>     * using the positive or negative names for skipping some tests

Positives please.

simo5:
> As far as I can see, the generic `tests/test.sh` script will only run properly, if it is called from either `test-softhsm.sh` or `test-softokn.sh`. For debugging purposes, it might be valuable, if it would be able to run the generic tests without do all the setup stuff. Or, if you want to switch between the two modules, just do a `source tests/tmp.<module>/debugvars; tests/test.sh`. But maybe we should cover this in a separate PR.

This would be nice indeed, what's missing to be able to do it?

holger-dengler:
> This would be nice indeed, what's missing to be able to do it?

I've not yet tested it, but at least the import of the `title` function. Maybe it is sufficient to do a source of `common.sh` in the beginning of test.sh as well.

holger-dengler:
Although the setting of the variables should not interfere with the tests, I would move the function definition of `title` to a separate script (e.g. `func.sh`, maybe together with the `ossl` function) and source `func.sh` in the beginning of `test.sh` script as well.

holger-dengler:
But, as I said, let's do that in a separate PR.

Jakuje:
> > * the path to the pkcs11 module is now set through environment variable. By default to softokn, but softhsm tests override this variable
> 
> Uhmmm, I guess this works, but I wonder if we shouldn't create two openssl..cnf files instead...

This would either require different replace commands in the Makefile, or some more replacements done in the tests themselves, and probably moving that away to the tmp directories.

Right now, it we run only the `test.sh` with different soft modules and the `tgenkey` and `ttls` is executed only against NSS swtok so I am not sure if there is some simple way to skip these when swtokn will not be available. But let me give it a try.

>>    * separate test matrix for sofhsm, softokn x all the currently existing configurations (fedora, debian, gcc, clang)
> Yes please.

Done (hope it will work correctly)

>>    * using the positive or negative names for skipping some tests
> Positives please.

Done

Jakuje:
The issue is with the `tgenkey`, which depends on something setting up the NSS db. Not sure how to handle this best. Everything else but this test works (or is skipped) in the matrix now.

simo5:
tgenkey should only needs openssl.cnf to point to a token with a pin in the config and nothing else.
The token needs to support key generation of course.

Does softhsm support key generation ?

simo5:
If softhsm can't handle tgenkey, we can wrap that test in a shell script named test_genkey.sh that will return 77 if softhsm is detected, I guess.

.github/workflows/build.yml
```diff
@@ -13,38 +13,71 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        name: [fedora-gcc, fedora-clang, debian-clang, fedora-distcheck]
+        name: [fedora-softokn-gcc, fedora-softokn-clang, debian-softokn-clang, fedora-softokn-distcheck,
```
simo5, line 16: You should probably use a multidimensional matrix here instead of manually expanding the names:

Define:
token: [softokn, softhsm]
name: [fedora-gcc, fedora-clang, debian-clang, fedora-distcheck]

This will automatically create all 8 jobs

then in use if [ "${{ matrix.token }}" in the tests to know what to install

.github/workflows/build.yml
```diff
@@ -59,8 +92,10 @@ jobs:
...
         name: Test logs ${{ matrix.name }}
         path: |
-          tests/test.sh.log
+          tests/*.log
           tests/openssl.cnf
-          tests/tmp/p11prov-debug.log
-          tests/tmp/debugvars
+          tests/tmp.softokn/p11prov-debug.log
+          tests/tmp.softokn/debugvars
+          tests/tmp.softhsm/p11prov-debug.log
```
simo5, line 99: I think you can use tests/tmp.${{ matrix.name }}/ patterns here to pull the logs

simo5 (changes requested):
I think this is mostly ok, we just need tio streamline the build yaml and resolve the tgenkey issue then we can merge.

Jakuje:
> tgenkey should only needs openssl.cnf to point to a token with a pin in the config and nothing else.
> The token needs to support key generation of course.

It needs at least initialized software token. But I kept that pointing to the softokn DB. Let me have a look if there is a way around

> Does softhsm support key generation ?

Yes, it does.

> If softhsm can't handle tgenkey, we can wrap that test in a shell script named test_genkey.sh that will return 77 if softhsm is detected, I guess.

It should, but it requires at least some initialization for the DB. Let me try to wrap it it some basic shell script to handle that.

simo5:
Well tgenkey runs after test.sh, so the idea is that it already has an initialized db, perhaps it now lacks the env var for the module given you moved it put of openssl.cnf

Jakuje:
Sorry for a delay. It took a bit more than I expected.

I reworked the tests a bit as well as the CI matrix according to your recommendations and it looks like it finally works. Few more notes:
 * Running the `tgenkey` tests works directly with softhsm but does not work through p11-kit remote (RSA goes through ok, but ECC does not for some reason) so I implemented a way to skip the p11-kit remoting dance.
 * The softhsm (when invoked directly) does not like the init arguments used for NSS so I incorporated one more sed to get rid of these and create temporary `openssl.cnf` in the end
 * I split the tests into separate `setup-*.sh` and `test*.sh` to make it easier to run the matrix when we will have the tests with opencryptoki or more tests. There is still need for the wrapper around the pairs though. This looked more readable than doing some magic using the autotools
 * The tests should be possible to invoke directly in the test directory as `./test-softhsm.sh` or just run the setup phase with `./setup-softokn.sh` (given there are provided some env variables)
 * The `ttls` looks like does not invoke the pkcs11 provider in any way (it passes regardless any of the software token is configured) so I did not touch that yet, but it should go in the same direction as the `tgenkey` once it will do something

simo5:
> Sorry for a delay. It took a bit more than I expected.
> 
> I reworked the tests a bit as well as the CI matrix according to your recommendations and it looks like it finally works. Few more notes:
> 
>     * Running the `tgenkey` tests works directly with softhsm but does not work through p11-kit remote (RSA goes through ok, but ECC does not for some reason) so I implemented a way to skip the p11-kit remoting dance.

ok, but if it works for this test I am surprised the whole dance is even necessary for the others, what am I missing?

>     * The softhsm (when invoked directly) does not like the init arguments used for NSS so I incorporated one more sed to get rid of these and create temporary `openssl.cnf` in the end

Wouldn't it make sense to always create two separate openssl.cnf files in tests.whatever/tmp ?
 
>     * I split the tests into separate `setup-*.sh` and `test*.sh` to make it easier to run the matrix when we will have the tests with opencryptoki or more tests. There is still need for the wrapper around the pairs though. This looked more readable than doing some magic using the autotools

ok

>     * The tests should be possible to invoke directly in the test directory as `./test-softhsm.sh` or just run the setup phase with `./setup-softokn.sh` (given there are provided some env variables)

Is it possible to run them as ./tests/test-whatever? or do they need to be run from within the tests directory ?

>     * The `ttls` looks like does not invoke the pkcs11 provider in any way (it passes regardless any of the software token is configured) so I did not touch that yet, but it should go in the same direction as the `tgenkey` once it will do something

The ttls binary used the provider indirectly in that all providers are queried foir tls group capabilities for some reason, see #90

Jakuje:
> > Sorry for a delay. It took a bit more than I expected.
> > I reworked the tests a bit as well as the CI matrix according to your recommendations and it looks like it finally works. Few more notes:
> >
> > * Running the `tgenkey` tests works directly with softhsm but does not work through p11-kit remote (RSA goes through ok, but ECC does not for some reason) so I implemented a way to skip the p11-kit remoting dance.
> 
> ok, but if it works for this test I am surprised the whole dance is even necessary for the others, what am I missing?

AFAIK the problems start only after the softhsm starts decrypting keys from the token store. If they are just generated and written, the cleanup issue does not surface.

> > * The softhsm (when invoked directly) does not like the init arguments used for NSS so I incorporated one more sed to get rid of these and create temporary `openssl.cnf` in the end
> 
> Wouldn't it make sense to always create two separate openssl.cnf files in tests.whatever/tmp ?

Right now, probably yes, but for now for example the test `ttls` needs this file outside of these directories as the environment variable is configured in the autotools.

> > * The tests should be possible to invoke directly in the test directory as `./test-softhsm.sh` or just run the setup phase with `./setup-softokn.sh` (given there are provided some env variables)
> 
> Is it possible to run them as ./tests/test-whatever? or do they need to be run from within the tests directory ?

The only problem I can see might be that we somehow need to detect the project root directory to reference the other scripts that are opened/executed/loaded from the tests. It might be either absolute path `$PWD` or relative path `..` or something more clever to find the root. For now, there is  `SOURCE_PATH=${SOURCE_PATH:-..}` in the scripts. But there are still some corner case that will not work. Let me try to address this tomorrow.

simo5:
> AFAIK the problems start only after the softhsm starts decrypting keys from the token store. If they are just generated and written, the cleanup issue does not surface.

ok

> > Wouldn't it make sense to always create two separate openssl.cnf files in tests.whatever/tmp ?
> 
> Right now, probably yes, but for now for example the test `ttls` needs this file outside of these directories as the environment variable is configured in the autotools.

ok, we could symlink the first or last created, but it is not a big deal.

> Let me try to address this tomorrow.

No worries, I think this PR is good as is, we'll refine it as we go.

simo5:
Uhmm I have some issues with the way things source each other:
$ ./test-digests-softhsm.sh 
Missing P11KITCLIENTPATH env variable

and that is right because debugvars does not include that variable.

And in my new PR the test segfaults with softhsm for some reason, but I can't reproduce if I run the test by hand, so it means other variables are probably amiss...

Jakuje:
> Uhmm I have some issues with the way things source each other: $ ./test-digests-softhsm.sh Missing P11KITCLIENTPATH env variable

Yeah, this one is needed to be passed from autotools for now.

> and that is right because debugvars does not include that variable.
> 
> And in my new PR the test segfaults with softhsm for some reason, but I can't reproduce if I run the test by hand, so it means other variables are probably amiss...

Let me have a look tomorrow.

PR #105, opened 2022-11-14
Try to trigger covscans via labels

simo5:


simo5:
Forgot that this does not work because on pull requests the action works within the context of the user and that context does not have access to th covscan secrets ...

PR #106, opened 2022-11-15
Refactor test suite

simo5:
Thanks to @jakuje we have now support for multiple tokens.
However that work revealed how annoyingly monolithic the test suite was.

This refactoring has multiple goals.

- Being able to run tests directly (or via a simple proxy script) and any amount of times without rerunning the setup scripts every time by simply sourcing the testvars generated
- Split tests in smaller units
- Use a wrapper script in the makefile so that the same tests can be run with multiple tokens without having to create individual shell script wrappers


tests/test-wrapper
```diff
@@ -0,0 +1,33 @@
...
+
+if [ -f "./tmp.${TOKEN_DRIVER}/testvars" ];  then
+    source ./tmp.${TOKEN_DRIVER}/testvars
+else
+    exit 77 # token not configured, skip
+fi
+
+# some tests are compiled, others are just distributed scripts
+# so we need to check both the current tests build dir and the
+# source tests dir in the make distcheck case
```
Jakuje, line 19: nit: I would adjust the comment to mention out-of-source build instead of distcheck as you can build out-of-source also outside of distcheck (but noone really does it except for `distcheck` though).

Jakuje (approved):
looks good to me.

PR #108, opened 2022-11-17
Rename the module binary to just pkcs11.so

simo5:
It is a little bit silly to name a provider something-provider, and other openssl providers just use their name.so (for examnple tpm2.so), so let's do the same for the pkcs11 provider and install it propoerly under /libdir/ossl-modules/pkcs11.so as expected by openssl provider distribution standards.

Also add a dstribution check CI tests, that runs make distcheck and builds an rpm from the spec file

### DIFF
--- a/packaging/pkcs11-provider.spec
+++ b/packaging/pkcs11-provider.spec
@@ -41,8 +41,6 @@ autoreconf -fi
 
 %install
 %make_install
-mkdir %{buildroot}%{_libdir}/ossl-modules/
-ln -s ../pkcs11_provider.so %{buildroot}%{_libdir}/ossl-modules/pkcs11.so
 
 
 %check
@@ -53,7 +51,6 @@ make check || if [ $? -ne 0 ]; then cat tests/*.log; exit 1; fi;
 %files
 %license COPYING
 %doc README
-%{_libdir}/pkcs11_provider.so
 %{_libdir}/ossl-modules/pkcs11.so
 
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,32 +13,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        name: [fedora-gcc, fedora-clang, debian-clang, fedora-distcheck]
+        name: [fedora, debian]
+        compiler: [gcc, clang]
         token: [softokn, softhsm]
         include:
-          - name: fedora-gcc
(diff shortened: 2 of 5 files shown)


packaging/pkcs11-provider.spec
```diff
@@ -42,7 +42,6 @@ autoreconf -fi
 %install
 %make_install
 mkdir %{buildroot}%{_libdir}/ossl-modules/
-ln -s ../pkcs11_provider.so %{buildroot}%{_libdir}/ossl-modules/pkcs11.so
```
holger-dengler, line 45: I'm not that familiar with spec files, but is the `mkdir` line still necessary? I would expect, that the previous `%make_install` already created the ossl-modules directory. Although it won't hurt, I would prefer to remove it.

holger-dengler (approved):
Looks good to me.

packaging/pkcs11-provider.spec
simo5: @jakuje Given I need to fix the PR I guess I can add a job to build the RPM after distcheck generates the tarball, right ?

Jakuje: There are different ways but certainly it should go into separate PR.

The packit usually just makes tarball from the currently committed files and creates SRPM from that and can build that either locally or in copr. AFAIK there is no way to pass it the tarball from make dist. 

Default integration requires installing Github application to the organization: https://packit.dev/docs/guide/#via-fedora-copr but for opensc we went with simple container running the packit and building the package locally: https://github.com/OpenSC/OpenSC/pull/2625/ (and https://github.com/OpenSC/OpenSC/pull/2528 for the packit configuration).

simo5: I just added a rpmbuild step to this PR.
It is just a small change, I do not think we need a whole new PR
It may fail in CI because I probably ned to pull in some more packages to have rpmbuild work, but you can look at the distcheck job change and let me know what you think.

simo5:
@jakuje rpmbuild seem to work to, can you approve (or disapprove if you think the test is not good) this PR ?

Jakuje (approved):

packaging/pkcs11-provider.spec
Jakuje: Right. Packit does not do much more fancy sftuff than calling rpmbuild internally so using the rpmbuild like this is totally ok.

Issue #109, opened 2022-11-18
Tests fail when building outside of source 


dengert:
**Describe the bug**
git HEAD is  47e1c475a4f6da14f7028208514ee888b3db305a
When configuring in a build directory (build) outside of the  source directory `make check` fails:
```
doug@XU-22:/a/appl/pkcs11-provider-git/build$ make check
Making check in src
make[1]: Entering directory '/a/appl/pkcs11-provider-git/build/src'
make[1]: Leaving directory '/a/appl/pkcs11-provider-git/build/src'
Making check in tests
make[1]: Entering directory '/a/appl/pkcs11-provider-git/build/tests'
make  tsession tgenkey ttls tdigests \
  helpers.sh setup-softhsm.sh setup-softokn.sh softhsm-proxy.sh test-wrapper tbasic teccsha2 thkdf toaepsha2 trsapss
make[2]: Entering directory '/a/appl/pkcs11-provider-git/build/tests'
make[2]: 'tsession' is up to date.
make[2]: 'tgenkey' is up to date.
make[2]: 'ttls' is up to date.
make[2]: 'tdigests' is up to date.
make[2]: Nothing to be done for '../../pkcs11-provider/tests/helpers.sh'.
make[2]: Nothing to be done for '../../pkcs11-provider/tests/setup-softhsm.sh'.
make[2]: Nothing to be done for '../../pkcs11-provider/tests/setup-softokn.sh'.
make[2]: Nothing to be done for '../../pkcs11-provider/tests/softhsm-proxy.sh'.
make[2]: Nothing to be done for '../../pkcs11-provider/tests/test-wrapper'.
make[2]: Nothing to be done for '../../pkcs11-provider/tests/tbasic'.
make[2]: Nothing to be done for '../../pkcs11-provider/tests/teccsha2'.
make[2]: Nothing to be done for '../../pkcs11-provider/tests/thkdf'.
make[2]: Nothing to be done for '../../pkcs11-provider/tests/toaepsha2'.
make[2]: Nothing to be done for '../../pkcs11-provider/tests/trsapss'.
make[2]: Leaving directory '/a/appl/pkcs11-provider-git/build/tests'
make  check-TESTS
make[2]: Entering directory '/a/appl/pkcs11-provider-git/build/tests'
make[3]: Entering directory '/a/appl/pkcs11-provider-git/build/tests'
SKIP: basic-softokn
SKIP: basic-softhsm-proxy
SKIP: eccsha2-softokn
[...]

PR #111, opened 2022-11-19
Fix "tests" when built outside source directory

dengert:
The test now work when built in the source or some other directory and configured like ../pkcs11-provider/configure

Fixes: #109

Define "testblddir" and "TESTBLDDIR" to point to the build/test directory i.e. "@abs_top_builddir@/tests" and correct several places where original code made some assumptions that the source directory was the same as the build directory.

`tests/test-wrapper` was patched to allow a file path name to include "-" It would fail to run some of the tests because it would not find the command an options.

The export PINFILE="${BASEDIR}/${PINFILE}" work when basename was null as previously in code it was already defined as PINFILE="${PWD}/pinfile.txt"

 Changes to be committed:
	modified:   tests/Makefile.am
	modified:   tests/openssl.cnf.in
	modified:   tests/setup-softhsm.sh
	modified:   tests/setup-softokn.sh
	modified:   tests/test-wrapper

### DIFF
--- a/tests/setup-softokn.sh
+++ b/tests/setup-softokn.sh
@@ -2,7 +2,7 @@
 # Copyright (C) 2022 Simo Sorce <simo@redhat.com>
 # SPDX-License-Identifier: Apache-2.0
 
-source ${TESTSDIR}/helpers.sh
+source ${TESTSSRCDIR}/helpers.sh
 
 if ! command -v certutil &> /dev/null
 then
@@ -146,20 +146,21 @@ OPENSSL_CONF=${BASEDIR}/${TMPPDIR}/openssl.cnf
 
 title LINE "Generate openssl config file"
 sed -e "s|@libtoollibs[@]|${LIBSPATH}|g" \
-    -e "s|@testsdir[@]|${BASEDIR}|g" \
-    ${TESTSDIR}/openssl.cnf.in > ${OPENSSL_CONF}
+    -e "s|@testssrcdir[@]|${BASEDIR}|g" \
+    -e "s|@testsblddir@|${TESTBLDDIR}|g" \
+    ${TESTSSRCDIR}/openssl.cnf.in > ${OPENSSL_CONF}
 
 title LINE "Export tests variables to ${TMPPDIR}/testvars"
 cat > ${TMPPDIR}/testvars <<DBGSCRIPT
 export PKCS11_PROVIDER_DEBUG="file:${BASEDIR}/${TMPPDIR}/p11prov-debug.log"
 export PKCS11_PROVIDER_MODULE="${SOFTOKNPATH}/libsoftokn3.so"
 export OPENSSL_CONF="${OPENSSL_CONF}"
-export TESTSDIR="${TESTSDIR}"
+export TESTSSRCDIR="${TESTSSRCDIR}"
+export TESTBLDDIR="${TESTBLDDIR}"
 
 export TOKDIR="${BASEDIR}/${TOKDIR}"
 export TMPPDIR="${BASEDIR}/${TMPPDIR}"
 export PINVALUE="${PINVALUE}"
-export PINFILE="${BASEDIR}/${PINFILE}"
 export SEEDFILE="${BASEDIR}/${TMPPDIR}/noisefile.bin"
 export RAND64FILE="${BASEDIR}/${TMPPDIR}/64krandom.bin"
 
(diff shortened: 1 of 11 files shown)


tests/setup-softokn.sh
```diff
@@ -155,11 +156,12 @@ export PKCS11_PROVIDER_DEBUG="file:${BASEDIR}/${TMPPDIR}/p11prov-debug.log"
 export PKCS11_PROVIDER_MODULE="${SOFTOKNPATH}/libsoftokn3.so"
 export OPENSSL_CONF="${OPENSSL_CONF}"
 export TESTSDIR="${TESTSDIR}"
+export TESTBLDDIR="${TESTBLDDIR}"
 
 export TOKDIR="${BASEDIR}/${TOKDIR}"
 export TMPPDIR="${BASEDIR}/${TMPPDIR}"
 export PINVALUE="${PINVALUE}"
-export PINFILE="${BASEDIR}/${PINFILE}"
+export PINFILE="${PINFILE}"
```
simo5, line 164: I do not see how changing this PINFILE export makes a difference or is correct.
If you change the openssl.cnf var to use testsblddir then you should also export this as `${TESTSBLDDIR}/${PINFILE}` I think and same for TMPDIR and TOKDIR.

simo5 (commented):
The intent looks good, but I am not sure about some of the substitution in the testvars files.

Also please can you add your signef-off-by to the commit ?

tests/setup-softokn.sh
simo5: Actually looking at tests we do not need to export PINFILE at all, nothing but openssl.cnf references it anyway. And I checked that it's path is now correct in testvars, but it is unesed, so would you mind just removing it ?
Also you did not correct it in setup-softhsm.sh, just remove it from exports there too

simo5:
Can you also make me a favor and rename TESTSDIR to TESTSSRCDIR while there?
This way it will be much clearer which dir *that* one refers to.

simo5:
Oh and the - in the paths is a great catch, I was very confused about what test-wrapper was trying to run ...

tests/setup-softokn.sh
dengert: OK, will remove "export PINFILE=..."

simo5:
LGTM, can you please rebase and squash the commits and add your signed-off-by line?

dengert:
both setup-softhsm.sh and setup-softokn.sh
- create an openssl.cnf file in their respective tests/tmp.softhsm and tests/tmp.softokn directories
- create the same pinfile in tests
- Both happen to set the same pin of 12345678

`openssl.cnf` has a `pkcs11-module-token-pin = file:/a/appl/pkcs11-provider-git/pkcs11-provider/tests/pinfile.txt`

If they really need a pinfile, I would suggest each create their own pinfile in their respective tests/tmp.softhsm and tests/tmp.softokn directories to avoid future problems if they set different pins. I could see other tests such as a setup-opensc.sh where a real test card is used, which may have a different pin or prompt for the pin.  

Is the `pkcs11-module-token-pin` actually used? Was it just added  during early development?  pkcs11 URI allows for passing a pin value.

simo5:
> both setup-softhsm.sh and setup-softokn.sh
> 
>     * create an openssl.cnf file in their respective tests/tmp.softhsm and tests/tmp.softokn directories
> 
>     * create the same pinfile in tests
> 
>     * Both happen to set the same pin of 12345678
> 
> 
> `openssl.cnf` has a `pkcs11-module-token-pin = file:/a/appl/pkcs11-provider-git/pkcs11-provider/tests/pinfile.txt`
> 
> If they really need a pinfile, I would suggest each create their own pinfile in their respective tests/tmp.softhsm and tests/tmp.softokn directories to avoid future problems if they set different pins. I could see other tests such as a setup-opensc.sh where a real test card is used, which may have a different pin or prompt for the pin.

We should probably also add tests that exercise the pkcs11 URI pin support as well as the configuration one. Mind creating an issue about this?


> Is the `pkcs11-module-token-pin` actually used? Was it just added during early development? pkcs11 URI allows for passing a pin value.

It is totally used, and a key feature to be able to specify the pinfile in the provider configuration rather than passing a naked pin around with keys as it is easier to leak that in log files.

Besides with things like p11-kit it is possible to obtain privilege separation with some applications via the client/server model (you can see it used in the softhsm tests) and keep the pinfile secret from the application, that is not possible if you pass the pinvalue through via pkcs11 URIs.

tests/setup-softokn.sh
dengert: done

dengert:
OK, rebased and signoff

PR #114, opened 2022-11-23
Some code simpifications by generalizing helper functions

fabled:
- Use OSSL_PARAM_get_utf8_string_ptr() when possible
- Refactor accessing p11prov_digest_get_* API to return digest struct
- Add p11prov_mech_by_mechanism() helper and use it

```
$ git diff --stat main...HEAD
 src/asymmetric_cipher.c |  36 ++++++++--------
 src/digests.c           |  97 ++++++++++++++++++-------------------------
 src/exchange.c          |  22 ++++------
 src/kdf.c               |  37 ++++++-----------
 src/provider.h          |  19 ++++++---
 src/signature.c         | 210 ++++++++++++++++++++++++++++++++++++++++-----------------------------------------------------
 6 files changed, 180 insertions(+), 241 deletions(-)
```

### DIFF
--- a/src/asymmetric_cipher.c
+++ b/src/asymmetric_cipher.c
@@ -505,11 +505,10 @@ static int p11prov_rsaenc_set_ctx_params(void *ctx, const OSSL_PARAM params[])
 
     p = OSSL_PARAM_locate_const(params, OSSL_ASYM_CIPHER_PARAM_OAEP_DIGEST);
     if (p) {
-        char digest[256];
-        char *ptr = digest;
+        const char *digest = NULL;
         CK_RV rv;
 
-        ret = OSSL_PARAM_get_utf8_string(p, &ptr, 256);
+        ret = OSSL_PARAM_get_utf8_string_ptr(p, &digest);
         if (ret != RET_OSSL_OK) {
             return ret;
         }
@@ -523,9 +522,8 @@ static int p11prov_rsaenc_set_ctx_params(void *ctx, const OSSL_PARAM params[])
 
     p = OSSL_PARAM_locate_const(params, OSSL_ASYM_CIPHER_PARAM_MGF1_DIGEST);
     if (p) {
-        char digest[256];
-        char *ptr = digest;
-        ret = OSSL_PARAM_get_utf8_string(p, &ptr, 256);
+        const char *digest = NULL;
+        ret = OSSL_PARAM_get_utf8_string_ptr(p, &digest);
         if (ret != RET_OSSL_OK) {
             return ret;
         }
--- a/src/asymmetric_cipher.c
+++ b/src/asymmetric_cipher.c
@@ -348,14 +348,14 @@ static const char *p11prov_rsaenc_mgf_name(CK_RSA_PKCS_MGF_TYPE mgf)
 {
     for (int i = 0; mfg_map[i].digest != CK_UNAVAILABLE_INFORMATION; i++) {
         if (mfg_map[i].mgf == mgf) {
-            const char *name;
+            const struct p11prov_digest *digest;
             CK_RV rv;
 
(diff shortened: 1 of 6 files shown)


src/asymmetric_cipher.c
```diff
@@ -505,11 +505,10 @@ static int p11prov_rsaenc_set_ctx_params(void *ctx, const OSSL_PARAM params[])
 
     p = OSSL_PARAM_locate_const(params, OSSL_ASYM_CIPHER_PARAM_OAEP_DIGEST);
     if (p) {
-        char digest[256];
-        char *ptr = digest;
+        const char *digest = NULL;
         CK_RV rv;
 
-        ret = OSSL_PARAM_get_utf8_string(p, &ptr, 256);
+        ret = OSSL_PARAM_get_utf8_string_ptr(p, &digest);
```
simo5: This was done to avoid allocations, what is the reason to change it?
Do you expect digest names a lot longer than 256 chars?

simo5 (changes requested):
I am sorry but I do not agree with the architecture of commits one and two.

I intentionally want to keep data private between files and use accessors instead. Making p11prov_degist public will not be accepted for now.

I also do not think the p11prov_mech change is really necessary, but I am not adamant about it. I still prefer the original code that pulls out specific algids because it makes the code very clear about what we are asking for. sO it is more readable. Returning the structure does not improve readability nor efficiency of the code.

fabled:
> I intentionally want to keep data private between files and use accessors instead. Making p11prov_digest public will not be accepted for now.

I thought this is not an issue as `struct p11prov_slot` is shared. But makes sense, even if it adds some boilerplate code for doing the accessors.

> I also do not think the p11prov_mech change is really necessary, but I am not adamant about it. I still prefer the original code that pulls out specific algids because it makes the code very clear about what we are asking for. sO it is more readable. Returning the structure does not improve readability nor efficiency of the code.

I try to splitting functions like this because you are now writing specialized functions for `WHAT_by_INDEX`. IT's generally less code to write searching functions `get_by_INDEX` and either add `WHAT` accessors, or use the struct directly. This actually is shown by the diffstat for the `p11prov_mech` and the way it removes the search loop from multiple places.

I'm happy to refactor this changes, but more specific direction and guidelines on what is acceptable would be appreciated.

simo5:
> > I intentionally want to keep data private between files and use accessors instead. Making p11prov_digest public will not be accepted for now.
> 
> I thought this is not an issue as `struct p11prov_slot` is shared. But makes sense, even if it adds some boilerplate code for doing the accessors.

I have not had the time to "fix" struct p11prov_slot yet indeed, there isn't a good reason why it is directly accessible, it is on the chopping list thoguh.

> > I also do not think the p11prov_mech change is really necessary, but I am not adamant about it. I still prefer the original code that pulls out specific algids because it makes the code very clear about what we are asking for. So it is more readable. Returning the structure does not improve readability nor efficiency of the code.
> 
> I try to splitting functions like this because you are now writing specialized functions for `WHAT_by_INDEX`. IT's generally less code to write searching functions `get_by_INDEX` and either add `WHAT` accessors, or use the struct directly. This actually is shown by the diffstat for the `p11prov_mech` and the way it removes the search loop from multiple places.
> 
> I'm happy to refactor this changes, but more specific direction and guidelines on what is acceptable would be appreciated.

I think a get_by_INDEX that has multiple return arguments is acceptable if there is a user for it. Internally (within a file or structure-as-object set of functions) code can be reorganized to reduce duplication as well, no arguments against that.

src/asymmetric_cipher.c
simo5: I see, I think this is fine then, thanks.

fabled:
Since this is many commits in one, I'll close this and submit separate PRs per commit as needed.

simo5:
Sounds good thanks

PR #116, opened 2022-11-23
Add basic support to load certificates from tokens

simo5:
This PR refactors how objects are stored.

In particular it merges the P11PROV_KEY structure into P11PROV_OBJ and adds support for representing both keys and certificates from a single object.

All code from keys.c and some from store.c is moved to objects.c


fabled (approved):
LGTM

PR #117, opened 2022-11-24
Simplify and fix signature DER AlgorithmInfo

fabled:
Changes to data:
 - DER_ECSHA2_SEQ_LEN was incorrectly 0x10 (should've been 0x0A)
 - DER_SEQ_ECDSA_SHA3_* incorrectly contained the NULL tag

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -174,103 +174,68 @@ static void p11prov_sig_freectx(void *ctx)
 
 #define DER_SEQUENCE 0x30
 #define DER_OBJECT 0x06
-#define DER_NULL 0x05, 0x00
+#define DER_NULL 0x05
 
-#define DER_RSAID_SEQ_LEN 0x0D
-#define DER_RSAID_LEN 0x09
-/* 1.2.840.113549.1.1 */
+/* iso(1) member-body(2) us(840) rsadsi(113549) pkcs(1) pkcs-1(1) */
 #define DER_RSADSI_PKCS1 0x2A, 0x86, 0x48, 0x86, 0xF7, 0x0D, 0x01, 0x01
+#define DER_RSADSI_PKCS1_LEN 0x08
 
-#define DER_SEQ_RSA_SHA512 \
-    DER_SEQUENCE, DER_RSAID_SEQ_LEN, DER_OBJECT, DER_RSAID_LEN, \
-        DER_RSADSI_PKCS1, 0x0D, DER_NULL
-const unsigned char der_rsa_sha512[] = { DER_SEQ_RSA_SHA512 };
-#define DER_SEQ_RSA_SHA384 \
-    DER_SEQUENCE, DER_RSAID_SEQ_LEN, DER_OBJECT, DER_RSAID_LEN, \
-        DER_RSADSI_PKCS1, 0x0C, DER_NULL
-const unsigned char der_rsa_sha384[] = { DER_SEQ_RSA_SHA384 };
-#define DER_SEQ_RSA_SHA256 \
-    DER_SEQUENCE, DER_RSAID_SEQ_LEN, DER_OBJECT, DER_RSAID_LEN, \
-        DER_RSADSI_PKCS1, 0x0B, DER_NULL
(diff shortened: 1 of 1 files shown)


fabled:
The style check fails, but it would be better IMHO this way to make the DER structure obvious. It helps validating the length fields.

simo5:
> The style check fails, but it would be better IMHO this way to make the DER structure obvious. It helps validating the length fields.

I had similar feelings when we introduced style checks, but ultimately decided that it is better to be consistent by following what the bot wants than creating exceptions, overall consistency is better even if sometimes the bot makes odd choices.

*of course* as the tool improves and upstream adds more tweaks we can change the tool to improve those areas that look odd, and change the style accordingly.

So if you find a flag in the tool that improves a specific area without compromising others it is ok to submit a PR to change options and sjust style of the impacted files.

src/signature.c
```diff
@@ -174,103 +174,68 @@ static void p11prov_sig_freectx(void *ctx)
 
 #define DER_SEQUENCE 0x30
 #define DER_OBJECT 0x06
-#define DER_NULL 0x05, 0x00
+#define DER_NULL 0x05
```
simo5, line 177: Do not understand this change, the ASN.1 for DER Null is type 5 followed by a 0 byte.

In fact everywhere you used DER_NULL later on was followed by a ", 0" addition.

So why change this and manually add zeros in the definitions ?

src/signature.c
```diff
@@ -174,103 +174,68 @@ static void p11prov_sig_freectx(void *ctx)
...
+        DER_SEQUENCE, DER_NIST_SIGALGS_LEN+5, \
+            DER_OBJECT, DER_NIST_SIGALGS_LEN+1, DER_NIST_SIGALGS, rsa_algid, \
+            DER_NULL, 0 \
+    }; \
+    static const unsigned char der_ecdsa_sha3_##bits[] = { \
+        DER_SEQUENCE, DER_NIST_SIGALGS_LEN+3, \
+            DER_OBJECT, DER_NIST_SIGALGS_LEN+1, DER_NIST_SIGALGS, ecdsa_algid \
+    };
+
+static const unsigned char der_rsa_sha1[] = {
```
simo5, line 219: In theory we could declare this via the ugly:
DEFINE_DER_SEQ_SHA(1, 0x05, 0x01);

Or am I missing something?

simo5 (changes requested):
Could you add a test that creates a signature with one of the incorrect ones you identified and then verifies it via OpenSSL default provider so we can catch if we are breaking this in the future?

src/signature.c
fabled: The following 0 byte is the Length field of the TLV. I changed this because:
 - `DER_SEQUENCE` and `DER_OBJECT` are Tag bytes only. So this makes `DER_NULL` consistent and be the Tag byte only.
 - It makes the Length fields explicit in the actual definition, and it's easier to validate the DER data on the byte array.
 - And it does not hide two bytes inside a macro. This enables validating the length calculations in the byte arrays.

For the length calculation reasons, I also changed the OID lengths to be the length of the named OID prefix variable. Instead of the expected sub entry.

src/signature.c
fabled: The ecdsa entry has different OID prefix, `DER_ANSIX962_SHA2_SIG` vs `DER_ANSIX962_SIG`. 

Though, this could be parametrized in the macros, by e.g.  adding `#define DEFINE_DER_ALGID(name, oid, oid_len)` and `#define DEFINE_DER_ALGID_WITH_NULL(name, oid, oid_len)` helpers to construct the individual DER arrays.

src/signature.c
simo5: Ah, thinking of it this way it makes a lot of sense.
Thanks!

src/signature.c
simo5: Nah, I was just amused as it seemed possible, but it doesn't make sense to  waste time on this.
We should probably just remove SHA-1, we are blocking it in newest distributions anyways ... but that will be in the future.

simo5 (approved):

PR #118, opened 2022-11-24
Add p11prov_mech_by_mechanism() helper and use it

fabled:
This generalizes the code quite a bit and removes copied code.


simo5:
LGTM

PR #121, opened 2022-11-24
Fix ECDSA signatures and improve tests

fabled:
- Add test for ECDSA verify using openssl algorithm
- Add test for EC/SHA2 CSR
- Add verify for RSA CSR

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -5,6 +5,7 @@
 #include <string.h>
 #include "openssl/evp.h"
 #include "openssl/rsa.h"
+#include "openssl/ec.h"
 #include "openssl/sha.h"
 
 struct p11prov_sig_ctx {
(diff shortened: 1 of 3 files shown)


fabled:
@simo5 Any ideas why the softhsm tests fail now?

fabled:
Fixed test to verify CSR instead. Turns out the ECDSA signature is likely wrong.

fabled:
Ok, updated the earlier EC signing test to use also openssl software verify of the signature. This fails.

PKCS#11 ECDSA returns `2nLen` blob. OpenSSL wants DER encoded ECDSASignature. So the Sign/Verify needs to convert back and forth these formats for `CKM_ECDSA`.

simo5:
To be honest I am surprised that we have to manually encode this.
I would assume the pkcs#11 tokens should do the encoding themselves, unless someone intentionally uses a raw signature mechanism and handrolls the rest.

fabled:
> To be honest I am surprised that we have to manually encode this.
> I would assume the pkcs#11 tokens should do the encoding themselves, unless someone intentionally uses a raw signature mechanism and handrolls the rest.

The PKCS11 current mechanisms specifies that the signature is in raw format, see:
http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/errata01/os/pkcs11-curr-v2.40-errata01-os-complete.html#_Toc441850447

simo5 (approved):
I see this is only format conversion in the end, pity that pkcs#11 does not do ASN.1, otoh it is also a good thing it doesn't.

LGTM.

PR #122, opened 2022-11-24
Fix RSA signatures with pre-calculated hash

### LINKED ISSUE #113: RSA signatures with pre-calculated hash, but digest set are calculated wrong
**Describe the bug**
Doing RSA PKCS1.5 signature with pre-calculated hash with hash information is calculated wrong. A DigestInfo struct should be included in the data to be signed.

**To Reproduce**
Patch to add for the test suite:
``` diff
diff --git a/tests/tbasic b/tests/tbasic
index 7153cb3..07015e9 100755
--- a/tests/tbasic
+++ b/tests/tbasic
@@ -54,6 +54,19 @@ pkeyutl -verify -inkey "${PUBURI}"
                 -in ${TMPPDIR}/sha256.bin
                 -sigfile ${TMPPDIR}/sha256-sig.bin'
 
+title PARA "Sign and Verify with provided Hash and RSA with DigestInfo struct"
+ossl 'dgst -sha256 -binary -out ${TMPPDIR}/sha256.bin ${SEEDFILE}'
+ossl '
+pkeyutl -sign -inkey "${PRIURI}" -pkeyopt digest:sha256
+              -in ${TMPPDIR}/sha256.bin
+              -out ${TMPPDIR}/sha256-sig.bin'
+
+ossl '
+pkeyutl -verify -inkey "${PUBURI}" -pkeyopt digest:sha256
+                -pubin
+                -in ${TMPPDIR}/sha256.bin
+                -sigfile ${TMPPDIR}/sha256-sig.bin'
+
 title PARA "Sign and Verify with provided Hash and EC"
 ossl '
 pkeyutl -sign -inkey "${ECBASEURI}"
```

1. Apply above patch
2. `make check`

**Expected behavior**
Modified test suite passes.

**Additional context**
The OpenSSL core will properly set rsasig context, and result in log:
> Set OSSL_SIGNATURE_PARAM_DIGEST to sha256

However, the sequence to call sign/verify will end up in a call to `p11prov_sig_set_mechanism(digest_sign=false)` and the digest gets completely ignored there.

It seems PKCS11 does not support signature with any of the `CKM_SHA256_RSA_PKCS` etc., so it is likely that the provider code should calculate and produce the `DigestInfo` structure with the hash info and use `CKM_RSA_PKCS` sign/verify mechanism.

fabled:
A DigestInfo DER structure should be constructred and signed when signing a pre-calculated hash with known digest.

fixes #113

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -175,6 +175,7 @@ static void p11prov_sig_freectx(void *ctx)
 #define DER_SEQUENCE 0x30
 #define DER_OBJECT 0x06
 #define DER_NULL 0x05
+#define DER_OCTET_STRING 0x04
 
 /* iso(1) member-body(2) us(840) rsadsi(113549) pkcs(1) pkcs-1(1) */
 #define DER_RSADSI_PKCS1 0x2A, 0x86, 0x48, 0x86, 0xF7, 0x0D, 0x01, 0x01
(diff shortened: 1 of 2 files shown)


src/signature.c
```diff
@@ -636,6 +673,31 @@ static int p11prov_sig_operate(P11PROV_SIG_CTX *sigctx, unsigned char *sig,
         }
     }
 
+    if (sigctx->mechtype == CKM_RSA_PKCS && sigctx->digest != 0) {
```
fabled, line 676: Wondering if this should be instead a `if (sigctx->digest != 0) { switch (sigctx->mechtype)` and whitelist the mechanisms where we know the proper padding is done, or `p11prov_sig_set_mechanism` can setup the p11 to match openssl.

simo5 (approved):
LGTM

Issue #124, opened 2022-11-25
URI matching issues


Jakuje:
**Describe the bug**
When parsing parts of the PKCS#11 URI, the function `parse_attr()` allocates just the `len + 1` bytes for the members (for example `uri->object`, but when these are compared against the token data, the hardcoded length from the PKCS#11 specification (32B) is used in `p11prov_uri_match_token()`. This leads to the buffer overruns (probably not reported because the openssl's allocs are used) and failures to match the keys.

**To Reproduce**
1. Try to select key by URI
2. See the matching failed

**Expected behavior**
Matching works

**Operating environment (please complete the following information):**
 - OS: Fedora
 - Version 36

**Token and application used (please complete the following information):**
 - Device: SoftHSM, p11-kit
 - PKCS11 Driver version: latest
 - Application libssh
 - Version master

**Additional context**
The `parse_attr` either needs to allocate the full length of the buffers it is checked against in `p11prov_uri_match_token()` or the `p11prov_uri_match_token()` needs to be adjusted to not compare behind the end of the buffer, ignoring trailing whitespace.

fabled:
This was fixed with #119 and can be closed?

Jakuje:
> This was fixed with #119 and can be closed?

Right. Took me too long to report the issue until you fixed that.

PR #125, opened 2022-11-25
Fix uri matching logic

Jakuje:
The URI matching assumes that all the tokeninfo fields are of fixed length. This is the case in the PKCS#11 layer, but the spaces are stripped with `trim()` in `provider.c:get_slots()`. The URI fields that are compared to them do not have fixed length as they are dynamically allocated so the matching should always fail (probably works because of custom openssl allocators stuff the memory with some more zeroes?).

This is removing the unused uri member, which was confusing while reading the code, and updates the matching to match as a C strings (with explicit length checks). I think this still can break in case of the field is full (without any space padding) though.

Fixes: #124

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -109,7 +109,7 @@ struct p11prov_uri {
     char *token;
     char *serial;
     CK_ATTRIBUTE id;
-    CK_ATTRIBUTE label;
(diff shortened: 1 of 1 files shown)


Jakuje:
Looks like I was testing without #119 so this is no longer needed.

PR #129, opened 2022-11-28
Add fallback to cert to get public key attributes

### LINKED ISSUE #127: Add support for getting public key from a certificate object
We have fallback code to load an associated public key from a private key object.
We need to extend this to also try to use a certificate object if a public key object is not available as some tokens store only certificate and private key and depend on software to retrieve the public key from the certificate.

simo5:
Allow the code to try not only to find an associated public key, but to also fallback further to try to find a certificate object and extract the public key attributes from that object.

Fixes #127

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -4,6 +4,7 @@
 #include "provider.h"
 #include <string.h>
 #include <openssl/ec.h>
+#include <openssl/x509.h>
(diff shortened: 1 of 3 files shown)


simo5:
I do not see a way to test this right now, but my manual testing says it works.
I opened #133 to track this

Issue #42, opened 2022-07-29
Implement support for keys with CKA_ALWAYS_AUTHENTICATE=CK_TRUE


fabled:
If a private key has `CKA_ALWAYS_AUTHENTICATE` set `CK_TRUE` to do any crypto operations on it, a secondary `C_Login(CKU_CONTEXT_SPECIFIC)` is needed in connection with the crypto operation.

simo5:
So reading the spec it says:

CKA_ALWAYS_AUTHENTICATE | CK_BBOOL | If   CK_TRUE, the user has to supply the PIN for each use (sign or decrypt) with   the key. Default is CK_FALSE.
-- | -- | --

Now, this is problematic because the sign or decrypt operations do not pass any password prompter to the provider.
The only chance of decrypting a key is when it is loaded from the store.
One way to handle this is to explicitly prompt the user at that time (unless the pkcs11_uri is 
specific to a key object and provides a pin), and then store the pin in the key object.

Then whenever we have a sign or decrypt operation we can call the special C_Login with the stored pin. This does not exactly meet the spirit of CKA_ALWAYS_AUTHENTICATE ... but I see no way around it.

Alternatively we store it and allow its use for exactly one operation then refuse to do any other requiring the program to load the key again via the store...

Jakuje:
There is some controversy around this attribute as this did not exist in older versions of the specs and when the private key operation failed, the caller should have tried to login and retry the operation in the past.

> Alternatively we store it and allow its use for exactly one operation then refuse to do any other requiring the program to load the key again via the store...

I think this won't work as the login operation needs to be just the one before the sign/decrypt operation on PKCS#11 layer. This might be enforced in the library or in the card that any other APDU sent to the card (or command to the HSM) might reset the state and the operation might not go through.

I think this is something we can live without for now, but as soon as this will get into real applications with PIV cards enforcing this, we will need to address this somehow.

simo5:
I think the main issue about this requirement is the need to suddenly serialize all operations.
I agree we should defer dealing with this requirement until it is really needed, and then we'll need to decide whether we want to deal with synchronization at the library level, or push it to the application level via documentation.

dengert:
This could be deferred and PIV could still be useable, if caching the PIN is acceptable.  

[...]

PR #130, opened 2022-11-29
Defer requiring login to token

### LINKED ISSUE #126: Readonly sessions without PIN fail unconditionally preventing even the key listing
**Describe the bug**
The PKCS#11 modules should be generally usable before login at least for listing of the keys. The NSS is taking this to the extreme, asking pin before any operation, which is OK, according to the standards, but painful for the users [[1](https://bugzilla.redhat.com/show_bug.cgi?id=1892137)][[2](https://bugzilla.redhat.com/show_bug.cgi?id=2016724)].
For this, the profile objects were invented in PKCS#11 3.0, but there are many PKCS#11 modules not supporting this yet a lot of software that is using OpenSSL and will be expecting this behavior [libssh].

**To Reproduce**
1. Remove the pkcs11-module-token-pin directive from `openssl.cnf`
2. Run search for a public key: `pkey -in $PUBURI -pubin -pubout -out ${TSTOUT}.pub`
3. This fails on cryptic message somewhere deep in creating sessions while attempting to login

**Expected behavior**
The applications should be allowed to list public objects without being prompted for a PIN and without requiring the pin in the configuration file.

 - OS: Fedora
 - Version 36

 - Device: SoftHSM, NSS Softokn from testsuite
 - PKCS11 Driver version: *
 - Application: libssh, openssl
 - Version any

**Additional context**
The pkcs11 provider should try to list the objects on the card even without login. The `CKF_LOGIN_REQUIRED` flag says only

> True if there are some cryptographic functions that a user MUST be logged in to perform

so this does not mean that the user needs to be logged in to do ANY operation on the module.

An reproduced and ugly solution that solved the issue for me is available in the follwing branch (but it will probably need some more work to work in all corner cases
https://github.com/Jakuje/pkcs11-provider/tree/login (https://github.com/latchset/pkcs11-provider/compare/main...Jakuje:login)

simo5:
Really require a login only if the calling code says a token is required. The calling code generally know when it is using a private key or performing other operation that generally requires logins.

- [x] Add test to check this does not regress

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -360,7 +360,7 @@ CK_SLOT_ID p11prov_session_slotid(P11PROV_SESSION *session)
 
 static CK_RV token_login(P11PROV_CTX *provctx, struct p11prov_slot *slot,
                          P11PROV_URI *uri, OSSL_PASSPHRASE_CALLBACK *pw_cb,
-                         void *pw_cbarg)
+                         void *pw_cbarg, bool reqlogin)
 {
     P11PROV_SESSION *session = NULL;
     char cb_pin[MAX_PIN_LENGTH + 1] = { 0 };
@@ -394,17 +394,22 @@ static CK_RV token_login(P11PROV_CTX *provctx, struct p11prov_slot *slot,
             ret = CKR_GENERAL_ERROR;
             goto done;
         }
+        if (cb_pin_len == 0) {
+            ret = reqlogin ? CKR_CANCEL : CKR_OK;
+            goto done;
+        }
 
         pin = (CK_UTF8CHAR_PTR)cb_pin;
         pinlen = cb_pin_len;
     } else {
-        ret = CKR_GENERAL_ERROR;
+        ret = reqlogin ? CKR_CANCEL : CKR_OK;
         goto done;
     }
 
     session = p11prov_session_new(provctx, slot);
     if (session == NULL) {
-        return CKR_GENERAL_ERROR;
+        ret = CKR_GENERAL_ERROR;
+        goto done;
     }
 
     session = p11prov_session_ref(session);
@@ -413,17 +418,18 @@ static CK_RV token_login(P11PROV_CTX *provctx, struct p11prov_slot *slot,
                       "Failed to ref count session");
         /* intentionally leave this broken session busy so it won't be
          * used anymore */
-        return CKR_GENERAL_ERROR;
+        ret = CKR_GENERAL_ERROR;
+        goto done;
     }
 
     ret = p11prov_session_open(session, true, pin, pinlen);
-
-done:
     if (ret == CKR_OK) {
         session->pool->login_session = session;
     } else {
         p11prov_session_free(session);
     }
+
+done:
     OPENSSL_cleanse(cb_pin, cb_pin_len);
     return ret;
 }
@@ -443,17 +449,16 @@ static CK_RV check_slot(P11PROV_CTX *provctx, struct p11prov_slot *provslot,
     if (uri) {
         /* skip slots that do not match */
         ret = p11prov_uri_match_token(uri, &provslot->token);
-    }
-    if (ret == CKR_OK
-        && ((provslot->token.flags & CKF_LOGIN_REQUIRED) || reqlogin)) {
-        ret = token_login(provctx, provslot, uri, pw_cb, pw_cbarg);
-    }
-
-    if (ret == CKR_OK && reqlogin && !provslot->pool->login_session) {
-        ret = CKR_USER_NOT_LOGGED_IN;
+        if (ret != CKR_OK) {
+            return ret;
+        }
     }
 
-    return ret;
+    if (!(provslot->token.flags & CKF_LOGIN_REQUIRED) && !reqlogin) {
+        return CKR_OK;
+    }
+
+    return token_login(provctx, provslot, uri, pw_cb, pw_cbarg, reqlogin);
 }
 
 /* There are three possible ways to call this function.
--- a/tests/tbasic
+++ b/tests/tbasic
@@ -138,4 +138,13 @@ req -new -batch -key "${PRIURI}" -out ${TMPPDIR}/rsa_csr.pem'
 ossl '
 req -in ${TMPPDIR}/rsa_csr.pem -verify -noout'
 
+title PARA "Test fetching public keys without PIN"
+ORIG_OPENSSL_CONF=${OPENSSL_CONF}
+sed "s/^pkcs11-module-token-pin.*$/##nopin/" ${OPENSSL_CONF} > ${OPENSSL_CONF}.nopin
+OPENSSL_CONF=${OPENSSL_CONF}.nopin
+ossl 'pkey -in $PUBURI -pubin -pubout -out ${TMPPDIR}/rsa.pub.nopin.pem'
+ossl 'pkey -in $ECPUBURI -pubin -pubout -out ${TMPPDIR}/ec.pub.nopin.pem'
+OPENSSL_CONF=${ORIG_OPENSSL_CONF}
+rm -f ${OPENSSL_CONF}.nopin
+
 exit 0


src/session.c
```diff
@@ -409,11 +409,15 @@ static CK_RV token_login(P11PROV_CTX *provctx, struct p11prov_slot *slot,
...
+        if (cb_pin_len == 0) {
+            ret = CKR_USER_NOT_LOGGED_IN;
+            goto done;
+        }
 
         pin = (CK_UTF8CHAR_PTR)cb_pin;
         pinlen = cb_pin_len;
     } else {
-        ret = CKR_GENERAL_ERROR;
+        ret = CKR_USER_NOT_LOGGED_IN;
```
Jakuje, line 420: my only concern is if the `CKR_USER_NOT_LOGGED_IN` is the correct error to return here (I used it mostly as a fast hack to try if it will do something in my patch set). The [PKCS#11 specs](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/pkcs11-base-v2.40-errata01-os-complete.html#_Toc441755792) say

> ·         `CKR_USER_NOT_LOGGED_IN`: The desired action cannot be performed because the appropriate user (or an appropriate user) is not logged in.  One example is that a session cannot be logged out unless it is logged in.  Another example is that a private object cannot be created on a token unless the session attempting to create it is logged in as the normal user.  A final example is that cryptographic operations on certain tokens cannot be performed unless the normal user is logged in.

which really does not match what we are trying to do here. But good is that the C_Login should not return this value if I see right.

Reading through other return codes, using something more "unique" would be better, for example "CKR_CANCEL"?  Or `CKR_FUNCTION_CANCELED`

> ·         `CKR_FUNCTION_CANCELED`: The function was canceled in mid-execution.  This happens to a cryptographic function if the function makes a CKN_SURRENDER application callback which returns CKR_CANCEL (see CKR_CANCEL). **It also happens to a function that performs PIN entry through a protected path.** The method used to cancel a protected path PIN entry operation is device dependent.

Or just use some `CKR_VENDOR_DEFINED + X` to avoid any conflicts altogether.

src/session.c
```diff
@@ -450,10 +454,13 @@ static CK_RV check_slot(P11PROV_CTX *provctx, struct p11prov_slot *provslot,
         ret = token_login(provctx, provslot, uri, pw_cb, pw_cbarg);
     }
 
-    if (ret == CKR_OK && reqlogin && !provslot->pool->login_session) {
+    if (ret == CKR_OK && !provslot->pool->login_session) {
         ret = CKR_USER_NOT_LOGGED_IN;
```
Jakuje, line 458: I am wondering in which case we go through this code path and if this is needed or if it can be somehow simplifiled by merging the code with the next condition.

The `ret` is `CKR_OK` and `provslot->pool->login_session` is `NULL` only if either the `token_login` was not called at all, isn't it?

tests/tbasic
```diff
@@ -138,4 +138,12 @@ req -new -batch -key "${PRIURI}" -out ${TMPPDIR}/rsa_csr.pem'
 ossl '
 req -in ${TMPPDIR}/rsa_csr.pem -verify -noout'
 
+title PARA "Test fetching public keys without PIN"
+ORIG_OPENSSL_CONF=${OPENSSL_CONF}
+sed "s/^pkcs11-module-token-pin.*$/##nopin/" ${OPENSSL_CONF} > ${OPENSSL_CONF}.nopin
+OPENSSL_CONF=${OPENSSL_CONF}.nopin
+ossl 'pkey -in $BASEURI -pubin -pubout -out ${TSTOUT}.pub.nopin'
```
Jakuje, line 145: In the original test I had also test with EC keys. Not a big deal and should not matter much, but would be good to have better coverage.

https://github.com/latchset/pkcs11-provider/compare/main...Jakuje:login

src/session.c
simo5: We could define our own error space, I've been tempted a few times already, but I see no problem in reusing CKR ones for our purpose, given we are never returning these errors to an application anyway (we are the consumer, not a pkcs#11 module).

I use CKR_CANCEL in other places, but I wanted an error that very specifically is about "login", and this makes it clear what happened even though it is slightly misused.

tests/tbasic
simo5: Alright, I can add also the EC one

src/session.c
Jakuje: This chain of several conditions on various variables setting just return value are hard to read and follow, especially if the code will go through more of these. I can test it works, but I can not say why, especially without any comments. Not blocking the merge of this, but I would love to see this somehow improved either now or later.

dengert:
Keep in mind that if CKF_PROTECTED_AUTHENTICATION_PATH is set in CK_TOKEN_INFO, this may mean a PINPAD reader is to be used. OpenSC supports this and can use the UI to send to tell user to look/use  at the PINPAD to enter pin.  See pkcs11-tool.c 

https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/csprd01/pkcs11-base-v3.0-csprd01.html#_Toc10472658
"To log into a token with a protected authentication path, the pPin parameter to C_Login should be NULL_PTR."

src/session.c
simo5: Ok, I'll change it then.

simo5:
> Keep in mind that if CKF_PROTECTED_AUTHENTICATION_PATH is set in CK_TOKEN_INFO, this may mean a PINPAD reader is to be used. OpenSC supports this and can use the UI to send to tell user to look/use at the PINPAD to enter pin. See pkcs11-tool.c
> 
> https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/csprd01/pkcs11-base-v3.0-csprd01.html#_Toc10472658 "To log into a token with a protected authentication path, the pPin parameter to C_Login should be NULL_PTR."

We are aare and have a bug for it, but we are not particularly interested in this option at this time.
It is useful for the tool OpenSC supports to manipulate cards which we are not going to replicate in this project.
But it is not very useful for the common use of the provider which is unattended access to HSMs from server programs.

If it turns out pkcs11-provider will come useful to more interactive programs like login processes we'll revisit and add support at that time.

simo5:
@Jakuje so I fixed the obvious things, what remains is whethere you think we should really introduce a custom error code for token_login()

I see that an option would be to just return a boolean variable as well together with CKR_CANCEL, to sidestep the issue of creating new codes.

simo5:
> @Jakuje so I fixed the obvious things, what remains is whethere you think we should really introduce a custom error code for token_login()
> 
> I see that an option would be to just return a boolean variable as well together with CKR_CANCEL, to sidestep the issue of creating new codes.

Actually, I think I found a better way to handle this within login_token, see the fixup commit.

dengert:
Please don't assume the provider will be used only:  "But it is not very useful for the common use of the provider which is unattended access to HSMs from server programs". It should be work when used  interactively by all the OpenSSL utilities.

simo5:
> Please don't assume the provider will be used only: "But it is not very useful for the common use of the provider which is unattended access to HSMs from server programs". It should be work when used interactively by all the OpenSSL utilities.

The main issue is that I have no way to test it right now.

But if you want to contribute code or ideas on how to handle it, I think a PR, or comments to #42 are the best places to do it.

simo5:
Lovely now the public test I added segfaults in make check, but if I run it manually it doesn't ...

simo5:
Ah because of #131 we now have a null dereference issue accessing a NULL session on the error path because now I goto done with CKR_OK early:
https://github.com/latchset/pkcs11-provider/blob/ac73f04e0b65093a5484d86782251b609692f565/src/session.c#L422-L423

simo5:
[btw the deleted comment above was nothing crazy, just a lost/broken comment with half a quote and nothing else, the actual comment from @dengert was then entered again here: https://github.com/latchset/pkcs11-provider/pull/130#issuecomment-1332508906]

tests/tbasic
```diff
@@ -138,4 +138,13 @@ req -new -batch -key "${PRIURI}" -out ${TMPPDIR}/rsa_csr.pem'
 ossl '
 req -in ${TMPPDIR}/rsa_csr.pem -verify -noout'
 
+title PARA "Test fetching public keys without PIN"
+ORIG_OPENSSL_CONF=${OPENSSL_CONF}
+sed "s/^pkcs11-module-token-pin.*$/##nopin/" ${OPENSSL_CONF} > ${OPENSSL_CONF}.nopin
+OPENSSL_CONF=${OPENSSL_CONF}.nopin
+ossl 'pkey -in $BASEURI -pubin -pubout -out ${TMPPDIR}/rsa.pub.nopin.pem'
+ossl 'pkey -in $ECPUBURI -pubin -pubout -out ${TMPPDIR}/ec.pub.nopin.pem'
```
Jakuje, line 146: nit: the `BASEURI` and `ECPUBURI` are not the same counterparts. I think the first should be `PUBURI`. The `BASEURI` probably works because of the type is added somehow along the way or the public key is first (and visible without login?)

Jakuje (approved):
looks good, aside the one last nit in test

Jakuje (approved):

PR #136, opened 2022-12-04
We can reduce this requirement

beldmit:
Debian stable still have 2.69, and everything works fine

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,7 +1,7 @@
 #                                               -*- Autoconf -*-
 # Process this file with autoconf to produce a configure script.
 
-AC_PREREQ([2.71])
+AC_PREREQ([2.69])
 AC_INIT([pkcs11-provider], [0.1], [simo@redhat.com])
 AC_CONFIG_SRCDIR([src/provider.c])
 AC_CONFIG_HEADERS([src/config.h])


beldmit:
I have several habits and I think I'm not the unique one :)

- having a custom build of openssl to test with (e.g. it can be built without optimization and patch if necessary) without installation
- don't raise the requirements until you really need it

It would be also great to specify the path to custom OpenSSL without dance with PKG_CONFIG_DIR

simo5:
At the moment I test with custom openssl this way (after setting LD_LIBRARY_PATH and PATH to point to openssl library and binaries path):
```
./Configure --prefix=$HOME/tmp/openssl --openssldir=$HOME/tmp/openssl/etc/pki/tls \
    enable-rfc3779 enable-md2 enable-rc5 no-mdc2 no-ec2m no-sm2 no-sm4 \
    enable-buildtest-c++ shared -ggdb3 '-DDEVRANDOM="\"/dev/urandom\""'

make all
make install
```

Are you thinking we need something different?
(note that setting PATH and LD_LIBRARY_PATH is needed anyway to be able to urn make check)

simo5:
In order to build against a custom openssl w/o setting PATH and LD_LIBRARY_PATH we would need two changes:
ability to set rpath at link time
change all our tests to refer to openssl binaries by absolute path

beldmit:
Yes, I use the similar setup. I'm currently speaking about the CFLAGS/LDPLAGS for configure

simo5:
@beldmit I actually pass them to configure:
`CFLAGS=-I/balh/includes LDFLAGS=blah ./configure`

simo5:
@beldmit pretty sure it works with any system, but it is prone to mistakes
you need to make sure to always have also PATH and LD_LIBRARY_PATH set

But if you want to contribute build system changes to  make this easier in this PR I am all for it.

simo5:
At the moment I do not recall a pressing need to require 2.71 strictly, so I'll merge.
However should something come up where 2.71 or higehr becomes required we will raise it again w/o appeal :-)

beldmit:
> 

sure, thanks!

PR #141, opened 2022-12-05
Skip slots that are not usable and check mechanism before selecting slot

simo5:
This PR handles optimizing iteration and usage of slots.
We skip completely slots that are at fault (can't fetch information) or where the token is not present.
We also now check that the token can handle the operation we are trying to perform at the time we request a session. This helps selecting the correct token on some operation when multiple tokens are available, and in general fails earlier if the token does not support the requested mechanism.

### DIFF
--- a/src/provider.c
+++ b/src/provider.c
@@ -1042,75 +1042,91 @@ static void trim_padded_field(CK_UTF8CHAR *field, ssize_t n)
 
 static CK_RV get_slots(P11PROV_CTX *ctx)
 {
-    CK_ULONG nslots;
+    CK_ULONG ns;
+    CK_ULONG ts;
     CK_SLOT_ID *slotid;
     struct p11prov_slot *slots;
     int ret;
 
-    ret = ctx->fns->C_GetSlotList(CK_FALSE, NULL, &nslots);
+    ret = ctx->fns->C_GetSlotList(CK_FALSE, NULL, &ns);
     if (ret) {
         return ret;
     }
 
     /* arbitrary number from libp11 */
-    if (nslots > 0x10000) {
+    if (ns > 0x10000) {
         return CKR_GENERAL_ERROR;
     }
 
-    slotid = OPENSSL_malloc(nslots * sizeof(CK_SLOT_ID));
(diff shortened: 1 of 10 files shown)


simo5:
@beldmit FYI

src/provider.c
```diff
@@ -1042,75 +1042,91 @@ static void trim_padded_field(CK_UTF8CHAR *field, ssize_t n)
...
-            ret = ctx->fns->C_GetTokenInfo(slotid[i], &slots[i].token);
+    /* keep separate counter for slots we'll actually have at the end,
+     * because we will skip slots that do not have any token present */
+    ts = 0;
+    for (size_t i = 0; i < ns; i++) {
+        CK_SLOT_INFO slot;
+        CK_TOKEN_INFO token;
+
+        ret = ctx->fns->C_GetSlotInfo(slotid[i], &slot);
+        if (ret != CKR_OK || (slot.flags & CKF_TOKEN_PRESENT) == 0) {
```
beldmit, line 1086: Does it work for 2.x?

simo5: C_getSlotInfo is available in 2.40 whcih is our current baseline:
http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/pkcs11-base-v2.40-errata01-os-complete.html#_Toc441755804

beldmit: `CKF_TOKEN_PRESENT` is also defined. Thanks!

simo5:
@beldmit added second patch to handle mechanism checks

If you could test this on your system it would be nice.

@holger-dengler you may want to test this on the setup where you had empty slot, as this PR should enable that environment now, which should reduce the amount of custom fiddling with the configurations for openCryptoKi

beldmit:
I confirm it works on a system with 2 different tokens

Jakuje:
nit: typo in the commit message (bogus 9), otherwise looks ok:

> 9in case multiple can match) that actually supports the operation.

simo5:
@Jakuje thanks I fixed the typo, merging.

PR #142, opened 2022-12-07
Wrap PKCS#11 interface calls

simo5:
This allows us a few things:
- Support multiple PKCS#11 version drivers
- Add automatic debugging/tracing and error reporting for each PKCS#11 function called
- Gain information on whether the token is modern enough to support forking processes

Additionally this allowed to remove a lot of cruft and some odd calls to p11prov_ctx_status() making some code a lot more readable.


Jakuje (approved):
looks ok on a fast review

PR #143, opened 2022-12-08
Add code that makes certs usable in OpenSSL

simo5:
A previous PR added code to allow loading certs.

This brings that work forward by correctly wiring the store api to return the cert objects to OpenSSL.

It can be exercised via the openssl storeutl command.

### DIFF
--- a/src/provider.h
+++ b/src/provider.h
@@ -474,7 +474,7 @@ CK_RV p11prov_fetch_attributes(P11PROV_CTX *ctx, P11PROV_SESSION *session,
                                unsigned long attrnums);
 
 #define MAX_PIN_LENGTH 32
-P11PROV_URI *p11prov_parse_uri(const char *uri);
+P11PROV_URI *p11prov_parse_uri(P11PROV_CTX *ctx, const char *uri);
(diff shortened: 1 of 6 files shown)


simo5:
@Jakuje when you created the softhsm setup it seem only public/private key pairs were created but no certs.

Can you help me with the necessary command to create/store certs there?

At the monebt this PR widens the gap adding a tcerts test only for softokn, so cert stuff is not tested at all with softhsm.

simo5:
btw it seems session-softhsm-proxy is flaky, sometimes fails but haven't looked into what is the cause.

simo5:
A simple rerun generally makes it pass

simo5:
Btw, found another OpenSSL bug while trying to test the latest two patches here:
https://github.com/openssl/openssl/pull/19856

fabled (approved):
LGTM

simo5:
Thanks!

Jakuje:
> @Jakuje when you created the softhsm setup it seem only public/private key pairs were created but no certs.

Yeah. At that time, the certs were not used for anything and I was striving for simplicity.

> Can you help me with the necessary command to create/store certs there?

Filled #144 for this.

PR #144, opened 2022-12-08
Generate certificates for SoftHSM tests

Jakuje:
This extends the setup similarly as it is done with softokn (in #143) now to have certificates on the smart cards.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,8 @@ jobs:
             if [ "${{ matrix.token }}" = "softokn" ]; then
               dnf -y install nss-softokn nss-tools nss-softokn-devel
             elif [ "${{ matrix.token }}" = "softhsm" ]; then
-              dnf -y install softhsm opensc  p11-kit-devel p11-kit-server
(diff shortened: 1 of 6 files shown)


simo5:
So in NSS softokn we have a CA cert and signed certs for a total of 4 certs, while this PR seem to generate 3 self-signed certs.
Additional the CNs are different.
I wonder if we should create a common configuration that all the drivers read in order to know what certs and keys to create, because otherwise we'll have to keep them manually in sync when we assume something in the tests.

Now what I do not get is why the last test in tcerts is not failing, given you do not have a cert with Subject CN=Issuer ...

Can you investigate that when you fix the failures on debian?

Jakuje:
Oh, I missed the issuer cert. Its not failing because of the missing `exit 1` in the case the cert is missing. I will fix that.

I used the self-sign certs as I had them at hand. It should not be complicated to rewrite it to use real issuer CA. Let me give it a try.

simo5:
Looks great, thanks!

simo5 (approved):

PR #147, opened 2022-12-19
Some housekeeping commits

simo5:
Some changes while I was evaluating next steps:
- Move p11prov_slot structure and assorted functions to session.c where it is mostly used.
- Reorganize provider.h in smaller header files, it was becoming unwieldy.
- Suppress some stderr from setup scripts (redirected to the log file)


Jakuje (approved):
looks ok on the fast review. I see most of the changes are moving stuff around.

src/digests.h
```diff
@@ -0,0 +1,64 @@
...
+    "SHA2-512/256:SHA-512/256:SHA512-256:2.16.840.1.101.3.4.2.6"
+#define P11PROV_DESCS_SHA2_512_256 "PKCS11 SHA2-512/224 Implementation"
+#define P11PROV_NAMES_SHA3_224 "SHA3-224:2.16.840.1.101.3.4.2.7"
+#define P11PROV_DESCS_SHA3_224 "PKCS11 SHA3-224 Implementation"
+#define P11PROV_NAMES_SHA3_256 "SHA3-256:2.16.840.1.101.3.4.2.8"
+#define P11PROV_DESCS_SHA3_256 "PKCS11 SHA3-256 Implementation"
+#define P11PROV_NAMES_SHA3_384 "SHA3-384:2.16.840.1.101.3.4.2.9"
+#define P11PROV_DESCS_SHA3_384 "PKCS11 SHA3-384 Implementation"
+#define P11PROV_NAMES_SHA3_512 "SHA3-512:2.16.840.1.101.3.4.2.10"
+#define P11PROV_DESCS_SHA3_512 "PKCS11 SHA3-512 Implementation"
```
beldmit, line 55: Looks like a half of these defines is the same with OpenSSL. Can we avoid duplication?

simo5: I do not think they are available in OpenSSL public headers, but if I am wrong, please let me know and I will gladly avoid re-defining them.

simo5:
> looks ok on the fast review. I see most of the changes are moving stuff around.

Thanks, there are indeed only minor adjustments in tests stuff on top of the moves, so I'll merge.

src/digests.h
beldmit: No, they aren't public :(

Thanks, will raise the upstream issue

Issue #149, opened 2022-12-22
RFC: Usefulness of debug messages


Jakuje:
Currently, the debug messages are very rough and not much verbose. Getting hang of what is going on there is sometimes hard for the people who did not write the code.

From debugging other projects for years, I was very happy with the verbosity of debug messages, that contain at least a function name in them so it is easy to figure out where the message comes from. For example, now messages can look like this:
```
Get session on slot 18
Error: 224
Failed to get session to load keys
```
The way to figure out which error is the above is mostly about grepping for some strings that look like they should be fixed (sometimes already in other function or in other file) and then guessing where the code went through. The least would be the function name, similarly like this which should already give rough estimate where the message comes from (this is something like OpenSSH):
```
p11prov_get_session: Get session on slot 18
p11prov_session_open: Error: 224
store_fetch: Failed to get session to load keys
```

Even better would be to have there exact location in the file, as sometimes functions are not always in the file where I would expect them to be. Adding pid or thread ID might also come very helpful when this will be used in multi-thread/multi-process applications in the future (this is similar what we have in OpenSC now):
```
[1234][T1] session.c:714 p11prov_get_session: Get session on slot 18
[1234][T1] session.c:471 p11prov_session_open: Error: 224
[1234][T1] store.c:106 store_fetch: Failed to get session to load keys
```

Additionally, I found useful in the past log messages at the beginning of any non-trivial function that record the function was called. This makes easy to trace the program and see where it went off the path I would expect.

The last minor note would be interpretation of the CKR values or printing these values in hexadecimal, as that is easier to find in the header files. Having the following log would already give much more information than the decimal number in the previous cases:
```
Error: 0xD0 (TEMPLATE_INCOMPLETE) 
```
This is mostly request for comments if somebody sees some additional information that would be useful to get logged or if you see some drawbacks of logging some of the information before I will start rewriting this code.

beldmit:
Do we need effectively duplicate pkcs11-spy functionality?

Jakuje:
> Do we need effectively duplicate pkcs11-spy functionality?

Its not about duplicating the pkcs11 spy functionality. This is about tracing what is going on in the provider itself and not on the pkcs11 interface only.

beldmit:
Thanks! Yes, it's worth doing and current OpenSSL mechanism can be used, I think. Unfortunately, it lacks the log-level conception :(

simo5:
[...]

PR #150, opened 2022-12-22
Simplify debug machinery and print line/function

simo5:
Fixes: #149

### DIFF
--- a/src/debug.h
+++ b/src/debug.h
@@ -5,38 +5,55 @@
 #define _DEBUG_H
 
 /* Debugging */
-extern int debug_lazy_init;
-#define P11PROV_debug_status(action) \
+extern int debug_level;
+#define P11PROV_debug(...) \
     do { \
-        int enabled = 0; \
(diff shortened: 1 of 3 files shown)


src/debug.h
```diff
@@ -5,38 +5,47 @@
...
+        } \
+    } while (0)
 
 #define P11PROV_debug_slot(...) \
-    P11PROV_debug_status(p11prov_debug_slot(__VA_ARGS__))
+    do { \
+        if (debug_level < 0) { \
+            p11prov_debug_init(); \
+        } else if (debug_level > 0) { \
+            p11prov_debug_slot(__VA_ARGS__); \
```
Jakuje, line 33: Should not this be separate `if` (similarly in the other macros here)? Isn't this going to ignore the first message and do just the initialization?

simo5: good catch!

simo5: Fixed in rebased patch

Jakuje (approved):
lgtm

PR #151, opened 2022-12-22
Use closed loop for FindObjectsInit/Final

### LINKED ISSUE #148: Interrupting the FindObjects sequence with other calls as C_GetAttributeValue can lead in incorrect results
The PKCS#11 WG has a page with common bugs and while reading one of the pkcs11 traces from pkcs11-provider, I noticed that it does this: Calling FindObjectsInit, then the GetAttributeValue and then moving on to other objects. 

> C_FindObjectsInit/Find/End can't be interrupted

> Interrupting the sequence C_FindObjectsInit/C_FindObjects/C_FindObjectsEnd with a different call results in inconsistent/erratic results. This makes it difficult to safely perform any general processing of multiple objects rather than just using the find-objects capability as a basic "find me this one object". 

https://wiki.oasis-open.org/pkcs11/CommonBugs

Last time I was writing something like this I think I was pointed to that page by somebody more experienced that this is wrong (even though the wording is not very clear). I changed the code to run the Find, store the handles and then read the objects attributes like this:

https://github.com/OpenSC/OpenSC/blob/master/src/tests/p11test/p11test_case_common.c#L800-L875

I could not find any similar wording in the PKCS#11 specification though and I believe any sane PKCS#11 module should handle this, but if there is somebody with more experience who could add more information, it would be appreciated

simo5:
Search all objects first and resolve them only after the FindObjectsInit / FindObjects / FindObjectsFinal loop is complete. This handles some tokens that may have issues handling more than one kind of operation per session, even though legal per spec.

Fixes #148

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -508,8 +508,8 @@ CK_RV p11prov_obj_from_handle(P11PROV_CTX *ctx, P11PROV_SESSION *session,
     return CKR_OK;
 }
 
-#define OBJS_IN_SEARCH 64
-#define MAX_OBJS_IN_STORE OBJS_IN_SEARCH * 16 /* 1024 */
+#define OBJS_PER_SEARCH 64
+#define MAX_OBJS_IN_STORE OBJS_PER_SEARCH * 16 /* 1024 */
 CK_RV p11prov_obj_find(P11PROV_CTX *provctx, P11PROV_SESSION *session,
                        CK_SLOT_ID slotid, P11PROV_URI *uri,
                        store_obj_callback cb, void *cb_ctx)
@@ -519,6 +519,7 @@ CK_RV p11prov_obj_find(P11PROV_CTX *provctx, P11PROV_SESSION *session,
     CK_ATTRIBUTE id = p11prov_uri_get_id(uri);
     CK_ATTRIBUTE label = p11prov_uri_get_label(uri);
     CK_ATTRIBUTE template[3] = { 0 };
+    CK_OBJECT_HANDLE *objects = NULL;
     CK_ULONG tsize = 0;
     CK_ULONG objcount = 0;
     CK_ULONG total = 0;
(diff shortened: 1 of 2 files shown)


src/objects.c
```diff
@@ -557,32 +558,26 @@ CK_RV p11prov_obj_find(P11PROV_CTX *provctx, P11PROV_SESSION *session,
...
+            return CKR_HOST_MEMORY;
+        }
+        ret = p11prov_FindObjects(provctx, sess, &objects[total],
+                                  OBJS_PER_SEARCH, &objcount);
         if (ret != CKR_OK || objcount == 0) {
             result = ret;
             break;
         }
         total += objcount;
+    } while (objcount != 0 && total < MAX_OBJS_IN_STORE);
```
Jakuje, line 576: the `objcount != 0` is already checked on line 571 so this condition is  not needed.

src/objects.c
```diff
@@ -557,32 +558,26 @@ CK_RV p11prov_obj_find(P11PROV_CTX *provctx, P11PROV_SESSION *session,
...
         return ret;
     }
-    while (total < MAX_OBJS_IN_STORE) {
-        CK_OBJECT_HANDLE object[OBJS_IN_SEARCH];
-        ret = p11prov_FindObjects(provctx, sess, object, OBJS_IN_SEARCH,
-                                  &objcount);
+    do {
+        objcount = 0;
+        objects = OPENSSL_realloc(objects, (total + OBJS_PER_SEARCH)
+                                               * sizeof(CK_OBJECT_HANDLE));
```
Jakuje, line 564: does the `OPENSSL_realloc` handle freeying of the old location in case of failure? My reading of the source of openssl is that it just defers to standard `realloc` and it has the following note in the manual pages:

> If realloc() fails, the original block is left untouched; it is not freed or moved.

so on error, the original location needs to be freed too.

src/objects.c
```diff
@@ -591,7 +586,25 @@ CK_RV p11prov_obj_find(P11PROV_CTX *provctx, P11PROV_SESSION *session,
...
+        P11PROV_OBJ *obj = NULL;
+        ret = p11prov_obj_from_handle(provctx, session, objects[k], &obj);
+        if (ret == CKR_CANCEL) {
+            /* unknown object or other recoverable error to ignore */
+            continue;
+        }
+        if (ret == CKR_OK) {
+            ret = cb(cb_ctx, obj);
+        }
+        if (ret != CKR_OK) {
```
Jakuje, line 599: this could use `else if` and `else` respectively, which might make the code slightly less complex.

src/objects.c
simo5: fair

src/objects.c
simo5: true, I generally do that but forgot in this case

src/objects.c
simo5: can't do a final else as the last check catches an error in either p11prov_obj_from_handle() or the cb() call. But I can do and if/else if for tje first two

Jakuje (approved):

PR #152, opened 2022-12-27
Allow pass custom OpenSSL dir

beldmit:
This patch allows building the provider against custom openssl build located aside. It's useful for debug purposes and is relevant for old systems having openssl 1.1.1 on board


simo5 (approved):
LGTM

Issue #55, opened 2022-09-09
Cache keys in sessions


simo5:
At least in some soft token a token key is really slow because it is stored encrypted on disk, so at each use it would have to be recovered with an expensive operation.

To speed up things for application that hold a private key structure we should copy the key to a session key object (this causes the soft toke to cache the private key) on the first operation that actually sues the private key and hold both handles on the key object. 

This means the key objects will need to hold a session.

simo5:
All we need it o C_CopyObject() with the token handle with the only attribute in the tempalte being CKA_TOKEN set to false.

PR #154, opened 2022-12-29
Caching of keys in session objects and refactoring of session management

simo5:
This code currently uses the login_session session as a stable, long term, open session.
Using any other session would require keeping another session open for as long as any key is cached.

Resolves: #55

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -95,17 +95,45 @@ static void destroy_key_cache(P11PROV_OBJ *obj, P11PROV_SESSION *session)
     }
 }
 
+static CK_RV supports_caching(P11PROV_CTX *ctx, CK_SLOT_ID id, int action,
+                              CK_BBOOL *data)
+{
+    CK_ULONG data_size = sizeof(CK_BBOOL);
+    void *data_ptr = &data;
+    const char *name = "Caching Supported";
+
+    switch (action) {
+    case GET_ATTR:
+        return p11prov_ctx_get_quirk(ctx, id, name, data_ptr, &data_size);
+    case SET_ATTR:
+        return p11prov_ctx_set_quirk(ctx, id, name, data, data_size);
+    default:
+        return CKR_ARGUMENTS_BAD;
+    }
+}
+
 static void cache_key(P11PROV_OBJ *obj)
 {
     P11PROV_SESSION *session = NULL;
     CK_BBOOL val_false = CK_FALSE;
     CK_ATTRIBUTE template[] = { { CKA_TOKEN, &val_false, sizeof(val_false) } };
     CK_SESSION_HANDLE sess;
+    CK_BBOOL can_cache = CK_TRUE;
     CK_RV ret;
 
     /* We cache only keys on the token */
     if ((obj->class != CKO_PRIVATE_KEY && obj->class != CKO_PUBLIC_KEY)
         || obj->cka_token != CK_TRUE || obj->cka_copyable != TRUE) {
+        return;
+    }
+
+    ret = supports_caching(obj->ctx, obj->slotid, GET_ATTR, &can_cache);
+    if (ret != CKR_OK) {
+        P11PROV_raise(obj->ctx, ret, "Failed to get quirk");
+    }
+    if (can_cache != CK_TRUE) {
+        /* switch copyable so we do not try again */
+        obj->cka_copyable = CK_FALSE;
         return;
     }
 
@@ -122,13 +150,20 @@ static void cache_key(P11PROV_OBJ *obj)
     ret = p11prov_CopyObject(obj->ctx, sess, obj->handle, template, 1,
                              &obj->cached);
     if (ret != CKR_OK) {
-        P11PROV_debug("Failed to cache key. Error %lx", ret);
+        P11PROV_raise(obj->ctx, ret, "Failed to cache key");
+        if (ret == CKR_FUNCTION_NOT_SUPPORTED) {
+            can_cache = CK_FALSE;
+            ret = supports_caching(obj->ctx, obj->slotid, SET_ATTR, &can_cache);
+            if (ret != CKR_OK) {
+                P11PROV_raise(obj->ctx, ret, "Failed to set quirk");
+            }
+        }
         /* switch copyable so we do not try again */
         obj->cka_copyable = CK_FALSE;
+    } else {
+        P11PROV_debug("Key %lu:%lu cached as %lu:%lu", obj->slotid, obj->handle,
+                      session, obj->cached);
     }
-
-    P11PROV_debug("Key %lu:%lu cached as %lu:%lu", obj->slotid, obj->handle,
-                  session, obj->cached);
 
     p11prov_return_session(session);
     return;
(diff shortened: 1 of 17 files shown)


simo5:
The initial commit does not check if the key is already cached, and can cause multiple copies of keys.

simo5:
Added code to cache only token keys, and not session keys.
Added also some minor refactoring in other areas as I tried to debug some failures in testing.

simo5:
This PR has grown a little bit beyond my initial intent.
But I think the last commit is worth it, as I found places where races were possible without the mutex locking.

beldmit:
> This PR has grown a little bit beyond my initial intent. But I think the last commit is worth it, as I found places where races were possible without the mutex locking.

I wonder to what extent the necessary locking should be in a provider itself and to what - in the OpenSSL EVP_PKEY code...

src/objects.c
```diff
@@ -51,24 +53,118 @@ P11PROV_OBJ *p11prov_obj_new(P11PROV_CTX *ctx, CK_SLOT_ID slotid,
...
+        P11PROV_debug("Failed to get login session. Error %lx", ret);
+        return;
+    }
+
+    /* If already cached, release and re-cache */
+    destroy_key_cache(obj, session);
+
+    sess = p11prov_session_handle(session);
+    ret = p11prov_CopyObject(obj->ctx, sess, obj->handle, template, 1,
+                             &obj->cached);
```
Jakuje, line 121: I think the usage of the CopyObject and DestroyObject for caching should not be used by default. It might work ok with the software modules, but other modules are probably going to choke on that.

src/objects.c
```diff
@@ -51,24 +53,118 @@ P11PROV_OBJ *p11prov_obj_new(P11PROV_CTX *ctx, CK_SLOT_ID slotid,
...
+    obj = p11prov_obj_ref_no_cache(obj);
+    if (!obj) {
+        return NULL;
+    }
+
+    /* When referenced it means we are likely going to try to use the key in
+     * some operation, let's try to cache it in the tokens volatile memory for
+     * those tokens that support the operation. This will result in much faster
+     * key operations with some tokens as the keys are unencrypted in volatile
+     * memory */
```
Jakuje, line 156: I would say this is very NSS-specific. See also the previous comment. I am not against adding this, but I would probably not enable doing this by default on all tokens and gate this behind some configuration option (or pkcs11 module detection/quirk)

src/signature.c
```diff
@@ -79,33 +90,21 @@ static void *p11prov_sig_dupctx(void *ctx)
...
         goto done;
     }
 
-    /* This is not really funny. OpenSSL by dfault assume contexts with
+    /* This is not really funny. OpenSSL by default assumes contexts with
      * operations in flight can be easily duplicated, with all the
-     * cryptographic status and then both context can keep going
+     * cryptographic status and then both contexts can keep going
      * independently. We'll try here, but on failure we just 'move' the
      * to the new token and hope for the best */
```
Jakuje, line 103:
```suggestion
     * independently. We'll try here, but on failure we just 'move'
     * to the new token and hope for the best */
```

src/objects.c
fabled: Agree. Most HSM drivers do not support session object (except perhaps derived keys). This will fail on any key where the COPYABLE flag is false, or likely fails if e.g. if SENSITIVE is true. In general it is not even required that a module supports Copy object operation. If this kind of caching is enabled, it should be a quirk enabled only for white listed modules. Or better, filed as a performance bug against the module.

fabled: I suppose the caching is optional as the error is mostly ignored with a debug level message. But attempting the copy always, may actually hit the device (behind a network) where it fails.
As minimum, failure to copy key should be cached and not attempted again.

simo5:
> > This PR has grown a little bit beyond my initial intent. But I think the last commit is worth it, as I found places where races were possible without the mutex locking.
> 
> I wonder to what extent the necessary locking should be in a provider itself and to what - in the OpenSSL EVP_PKEY code...

Not sure what the right answer is yet, but at the moment I cannot rely on OpenSSL to know when it is safe to lock. And whether locking is needed may really depend on the provider. Like I could imagine a (HW or SW) device that can only do one operation at a time with a key, while another can do multiple concurrently. OpenSSL can't know.

src/objects.c
simo5: Will softhsm or pcscd suffer from doing this? (Performance, other?)
If not and there is no harm I would leave it on. Unclear how I would configure this, and explicit option in openssl.cnf ?

src/objects.c
simo5: Yeah I can see adding a signal that caching doesn't work.
I wonder if this is something we can reliably detect per slot, so that we try at most once for devices that can't handle it.

simo5: Should I check and trust CKA_COPYABLE ?
And default to false if it is not returned ?
Seem like a good indication.

src/objects.c
Jakuje: not sure about softhsm, but this will end up in OpenSC here as unimplemented so some check to avoid calling this over and over again (or having quirk for opensc) would be needed anyway:
https://github.com/OpenSC/OpenSC/blob/70771735ae10180bb039043b9a1b00b66bf00fc1/src/pkcs11/pkcs11-object.c#L166-L173

src/objects.c
Jakuje: > I wonder if this is something we can reliably detect per slot, so that we try at most once for devices that can't handle it.

It will probably depend on the error. If you get UNSUPPORTED, its probably whole slot/module.  Other errors such as `CKR_ACTION_PROHIBITED` might mean this only for a object, but I never worked with this in pkcs11.

simo5: I added a patch, let me know what you think

src/objects.c
simo5: Alright, I will add a tunable to skip caching entirely if we get that

simo5: Added quirk

src/session.c
```diff
@@ -353,207 +516,157 @@ CK_RV p11prov_session_pool_init(P11PROV_CTX *ctx, CK_TOKEN_INFO *token,
...
-        P11PROV_SESSION **tmp =
-            OPENSSL_realloc(p->sessions, (p->num_p11sessions + SESS_ALLOC_SIZE)
-                                             * sizeof(P11PROV_SESSION *));
+    if ((pool->num_p11sessions % SESS_ALLOC_SIZE) == 0) {
+        P11PROV_SESSION **tmp = OPENSSL_realloc(
+            pool->sessions, (pool->num_p11sessions + SESS_ALLOC_SIZE)
+                                * sizeof(P11PROV_SESSION *));
         if (tmp == NULL) {
-            P11PROV_raise(ctx, CKR_HOST_MEMORY,
+            P11PROV_raise(pool->provctx, CKR_CANT_LOCK,
```
Jakuje, line 590: Should not this be the `CKR_HOST_MEMORY` error instead of the `CKR_CANT_LOCK`?

simo5: Good catch!

simo5: Pushed fix

Jakuje (approved):

PR #155, opened 2023-01-03
Streamlinbe fetch_attrs interface

simo5:
This allows for writing much shorter functions with a lot less auxiliary variables needed.
It also avoids many manual coding errors by providing more utility functions.

### DIFF
--- a/src/exchange.c
+++ b/src/exchange.c
@@ -227,6 +227,7 @@ static int p11prov_ecdh_derive(void *ctx, unsigned char *secret,
     CK_SLOT_ID slotid;
     unsigned long secret_len;
     struct fetch_attrs attrs[1];
+    int num = 0;
     CK_RV ret;
 
     if (ecdhctx->key == NULL || ecdhctx->peer_key == NULL) {
@@ -291,9 +292,9 @@ static int p11prov_ecdh_derive(void *ctx, unsigned char *secret,
     }
 
     P11PROV_debug("ECDH derived hey handle: %lu", secret_handle);
-    FA_ASSIGN_ALL(attrs[0], CKA_VALUE, &secret, &secret_len, false, true);
+    FA_SET_BUF_VAL(attrs, num, CKA_VALUE, secret, secret_len, false, true);
     ret = p11prov_fetch_attributes(ecdhctx->provctx, session, secret_handle,
-                                   attrs, 1);
+                                   attrs, num);
     p11prov_return_session(session);
     if (ret != CKR_OK) {
         P11PROV_debug("ecdh failed to retrieve secret %lu", ret);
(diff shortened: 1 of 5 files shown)


simo5:
@Jakuje you mentioned on a recent PR that we could improve attribute fetching for keyt objects, as there is a lot of repetition.
This patch goes in that direction.

src/exchange.c
```diff
@@ -291,7 +292,7 @@ static int p11prov_ecdh_derive(void *ctx, unsigned char *secret,
     }
 
     P11PROV_debug("ECDH derived hey handle: %lu", secret_handle);
-    FA_ASSIGN_ALL(attrs[0], CKA_VALUE, &secret, &secret_len, false, true);
+    FA_SET_BUF_VAL(attrs, num, CKA_VALUE, secret, secret_len, false, true);
     ret = p11prov_fetch_attributes(ecdhctx->provctx, session, secret_handle,
                                    attrs, 1);
```
Jakuje, line 297:
```suggestion
                                   attrs, num);
```

src/kdf.c
```diff
@@ -154,7 +155,7 @@ static int p11prov_hkdf_derive(void *ctx, unsigned char *key, size_t keylen,
     }
 
     P11PROV_debug("HKDF derived hey handle: %lu", dkey_handle);
-    FA_ASSIGN_ALL(attrs[0], CKA_VALUE, &key, &dkey_len, false, true);
+    FA_SET_BUF_VAL(attrs, num, CKA_VALUE, key, dkey_len, false, true);
     ret = p11prov_fetch_attributes(hkdfctx->provctx, hkdfctx->session,
                                    dkey_handle, attrs, 1);
```
Jakuje, line 160:
```suggestion
                                   dkey_handle, attrs, num);
```

Jakuje:
> @Jakuje you mentioned on a recent PR that we could improve attribute fetching for keyt objects, as there is a lot of repetition. This patch goes in that direction.

If you refer to https://github.com/latchset/pkcs11-provider/pull/153#discussion_r1060597349 it was about code repetition when printing ID and label for different key types and not about fetching attrs, but this generally looks ok. But I will have a better look tomorrow.

src/objects.c
```diff
@@ -37,7 +37,7 @@ struct p11prov_obj {
     } data;
 
     CK_ATTRIBUTE *attrs;
-    unsigned long numattrs;
+    int numattrs;
```
beldmit, line 40: Why this change?

src/kdf.c
simo5: Same as above :)

src/objects.c
simo5: because of p11prov_move_alloc_attrs(), it takes an int, and we really do not need unsigned long many attrs :-)

src/exchange.c
Jakuje: I would rather have larger diff than magic numbers around that you need to change when you add more attrs. But it might be just my preference.

Jakuje (approved):

simo5:
Thanks Jakub.

PR #157, opened 2023-01-04
Add a simple code generator

simo5:
Often we want to use macros to simplify writing boilerplate.

On one hand this is good as it avoid tedious, error-prone work.
OTOH it makes reading and debugging code harder when trying to simply traverse complex code bases either with code indexers/taggers or even just in debuggers that get often confused when the source code file does not match the compiled code.
    
This simple generator allows to write macros to auto generate parts of code and uses the compiler pre-processor to spit out actual readable code that can be formatted by clang-format and committed to the tree.


simo5:
@beldmit this is the kind of clever solution I prefer to the macro-madness problem :-)

Makefile.am
```diff
@@ -18,6 +18,19 @@ check-style-show:
 check-style-fix:
 	git diff -U0 --no-color --relative origin/main | clang-format-diff -i -p1
 
+generate-code:
+	for pfile in src/*.pre; do \
+		gfile=`echo $${pfile} | sed s/\.pre/\.gen\.c/`; \
+		echo "/* DO NOT EDIT, autogenerated from $${pfile} */" > "$${gfile}"; \
+		echo "/* Modify $${proto} then run make generate-code */" >> "$${gfile}"; \
```
Jakuje, line 25:
```suggestion
		echo "/* Modify $${pfile} then run make generate-code */" >> "$${gfile}"; \
```

src/encoder.pre
```diff
@@ -0,0 +1,24 @@
+/* Copyright (C) 2022 Simo Sorce <simo@redhat.com>
```
Jakuje, line 1:
```suggestion
/* Copyright (C) 2022 - 2023 Simo Sorce <simo@redhat.com>
```

beldmit:
I like this approach for PKCS#11 wrappers. I'm not sure we get any advantages for encoder/decoder stuff and wonder whether we can reuse OpenSSL native code here

src/encoder.pre
simo5: Heh .... In this case I just copied over code written in 2022 :-)

simo5:
> I like this approach for PKCS#11 wrappers. I'm not sure we get any advantages for encoder/decoder stuff and wonder whether we can reuse OpenSSL native code here

The sad part is that the stuff in encoder.pre *is* OpenSSL code :-)

beldmit:
> > I like this approach for PKCS#11 wrappers. I'm not sure we get any advantages for encoder/decoder stuff and wonder whether we can reuse OpenSSL native code here
> 
> The sad part is that the stuff in encoder.pre _is_ OpenSSL code :-)

Yes. And my question is whether we have to effectively duplicate it here instead using the base provider.
It's probably an architecture question to upstream - and if so, we should raise it.

simo5:
@beldmit for public keys we do not *have to*, I *chose* to do it so I could print additional information (CKA_ID and CKA_LABEL) in the text versions, or to prevent exporting public keys to OpenSSL.

OpenSSL always attemts to export public keys out of a provider and import them in the default provider if it has support for the algorithm (do not know why probably unintended consequence), and this means that any future computation that uses the public key is then done by OpenSSL's default provider. This is fine in many circumstances, and even desirable in some cases when the pkcs#11 driver is slow (example network HSM). But in other cases we may have a pkcs#11 driver for an accelarator and we want to keep public key operations on the token as well without having to change applications, so by denying public key export we achieve that, but as a consequence we need to implement some of the encoders ourselves (I guess I could call the base provider directly from this code by internally exporting the public key ... but it would probably be more code than just encoding directly).

Bottom-line I think OpenSSL gives you the option to avoid encoders (at least for public keys), so there is nothing really to report except perhaps the sorry state of provider developer documentation which is basically non-existant and forces me to read a ton of internal OpenSSL code to figure out how to do things :-)

beldmit (approved):
LGTM

PR #158, opened 2023-01-04
Allow retrieving group name for EC keys

Jakuje:
The openssl excersises different code paths when accessing OSSL_PARAMS of a
provided EVP_PKEY structures. While going with EVP_PKEY_todata() works as
expected, using EVP_PKEY_get_utf8_string_param() goes through different code
path to retrive the parameters and in current version fails.

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -1306,31 +1306,21 @@ int p11prov_obj_export_public_rsa_key(P11PROV_OBJ *obj, OSSL_CALLBACK *cb_fn,
     return ret;
 }
 
-#define EC_PUB_ATTRS 2
-int p11prov_obj_export_public_ec_key(P11PROV_OBJ *obj, OSSL_CALLBACK *cb_fn,
-                                     void *cb_arg)
+const char *p11prov_obj_get_ec_group_name(P11PROV_OBJ *obj)
 {
-    CK_ATTRIBUTE attrs[EC_PUB_ATTRS] = { 0 };
-    OSSL_PARAM params[EC_PUB_ATTRS + 1];
-    ASN1_OCTET_STRING *octet = NULL;
+    CK_ATTRIBUTE attrs[1] = { 0 };
     EC_GROUP *group = NULL;
     const unsigned char *val;
     const char *curve_name;
     int curve_nid;
     CK_RV rv;
-    int ret;
-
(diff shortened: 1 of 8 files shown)


Jakuje:
Tested with the reverted commit f50b772 the simple reproducer fails.

src/objects.c
```diff
@@ -1306,31 +1306,25 @@ int p11prov_obj_export_public_rsa_key(P11PROV_OBJ *obj, OSSL_CALLBACK *cb_fn,
...
     EC_GROUP *group = NULL;
     const unsigned char *val;
     const char *curve_name;
     int curve_nid;
     CK_RV rv;
-    int ret;
 
-    if (p11prov_obj_get_key_type(obj) != CKK_EC) {
-        return RET_OSSL_ERR;
+    if (p11prov_ctx_allow_export(obj->ctx) & DISALLOW_EXPORT_PUBLIC) {
```
simo5, line 1318: This is wrong here

src/objects.c
```diff
@@ -1340,28 +1334,64 @@ int p11prov_obj_export_public_ec_key(P11PROV_OBJ *obj, OSSL_CALLBACK *cb_fn,
     group = d2i_ECPKParameters(NULL, (const unsigned char **)&val,
```
simo5, line 1334: given we do this already in fetch_ec_key(), I wonder if we shouldn't simply get all this data at that time and store it in the object, d2i functions are not exactly light (many allocations and transformations) and the results never change, might be better to do the op once at fetch and not every time a param is requested.

simo5 (changes requested):
Except for the noted comment I think this is ok.
I can deal with saving the data later, it is just an optimization.

src/objects.c
Jakuje: oh, yeah. Part of the rebase pains. Will fix.

src/objects.c
Jakuje: Yeah. That would be even better. I think storing the variable in `p11prov_key` should do that, but I dont have the whole architecture in my mind now so I am ok if you want to do it.

Not sure if we have this issue with other parameters as libssh uses only this one in this way now.

simo5 (approved):
LGTM

PR #159, opened 2023-01-04
Detect if token is write protected

simo5:
In which case return early and do not even try to login and go through operations that will fail.

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -807,13 +807,16 @@ done:
 }
 
 static CK_RV check_slot(struct p11prov_slot *provslot, P11PROV_URI *uri,
-                        CK_MECHANISM_TYPE mechtype)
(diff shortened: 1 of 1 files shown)


simo5:
@rjrelyea can you review?

beldmit (approved):
LGTM, but wonder how we can test it

simo5:
@beldmit NSS softokn by default always exposes two slots and one is always write protected because it is  not intended for storing data, so this is implicitly tested in CI.

rjrelyea:
Yes, it looks good. This will also return read only in softoken if the the DB's are openned read only. Also smart card are often read only as they are provisioned outside of PKCS #11, so this is a good general solution.

simo5:
Thanks Bob

PR #162, opened 2023-01-05
Session locking fixes

simo5:
In PR #156 an attempt to fix some minor locking issues resulted in a much worse potential deadlock within the code that attempts to create a login session.

The code roughly did the following:
    token_login {

        /* A section */
        lock -> pool
        lock -> session
        unlock -> pool

        /* issue  */

        /* B section */
        lock->pool
        unlock->pool
        return
    }

After section A, we hold the session lock, but do not hold the pool lock, that means another thread could be calling token_login() before we enter section B, and take the pool lock.
Then attempt to take the session lock and be stuck waiting. At the same time the original thread will be stuck trying to get the pool lock in section B, and not releaseing the session lock. This is a classic deadlock situation and both threads will wait forever.

The basic solution is to use MUTEX_TRY_LOCK to get the session lock so that in case of deadlock we'll get an error instead of being stuck forever.

After taking a step back however I reeliazed there were many shortcomings in how the code was refactor to avoid waiting withing locked sections, so the solution is a bit larger than a minimal fix, and better streamlines some internal operations reducing the number of locking sections to the bare minimum needed.

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -412,12 +412,12 @@ struct p11prov_session_pool {
     P11PROV_CTX *provctx;
     CK_SLOT_ID slotid;
 
-    CK_ULONG max_sessions;
-    CK_ULONG limit_sessions;
-    CK_ULONG cur_sessions;
+    int num_sessions;
+    int max_sessions;
+    int open_sessions;
+    int max_cached_sessions;
 
     P11PROV_SESSION **sessions;
-    int num_p11sessions;
 
     P11PROV_SESSION *login_session;
 
@@ -429,32 +429,29 @@ struct p11prov_session_pool {
 /* sleep interval, 50 microseconds (max 20 attempts) */
 #define SLEEP 50000
 
-/* NOTE: to be called with Pool Lock held */
 static CK_RV token_session_open(P11PROV_SESSION *session, CK_FLAGS flags)
 {
-    bool wait_ok = true;
     uint64_t startime = 0;
     CK_RV ret;
 
-    if (session->pool->cur_sessions >= session->pool->max_sessions) {
-        P11PROV_debug_once("Max Session (%lu) exceeded!",
-                           session->pool->max_sessions);
-        return CKR_SESSION_COUNT;
-    }
-
-    while (wait_ok) {
+    do {
         ret = p11prov_OpenSession(session->provctx, session->slotid, flags,
                                   NULL, NULL, &session->session);
         P11PROV_debug("C_OpenSession ret:%lu (session: %lu)", ret,
                       session->session);
-        if (ret == CKR_SESSION_COUNT) {
-            wait_ok = cyclewait_with_timeout(MAX_WAIT, SLEEP, &startime);
-            continue;
+        if (ret != CKR_SESSION_COUNT) {
+            break;
         }
-        break;
+    } while (cyclewait_with_timeout(MAX_WAIT, SLEEP, &startime));
+
+    if (ret != CKR_OK) {
+        session->session = CK_INVALID_HANDLE;
+        session->flags = DEFLT_SESSION_FLAGS;
+        return ret;
     }
 
-    return ret;
+    session->flags = flags;
+    return CKR_OK;
 }
 
 static void token_session_close(P11PROV_SESSION *session)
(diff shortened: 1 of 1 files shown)


src/session.c
```diff
@@ -388,12 +402,13 @@ static CK_RV mutex_destroy(P11PROV_CTX *provctx, pthread_mutex_t *lock,
...
     }
     return ret;
 }
-#define MUTEX_DESTROY(obj) mutex_destroy((obj)->provctx, &(obj)->lock, #obj)
+#define MUTEX_DESTROY(obj) \
+    mutex_destroy((obj)->provctx, &(obj)->lock, #obj, OPENSSL_FILE, \
+                  OPENSSL_LINE, OPENSSL_FUNC)
 
 /* Session stuff */
 #define DEFLT_SESSION_FLAGS CKF_SERIAL_SESSION
```
Jakuje, line 414: should this say `DEFAULT_SESSION_FLAGS` or the abbreviation is intentional or more readable than expanding?

src/session.c
```diff
@@ -492,18 +504,15 @@ static CK_RV session_pool_init(P11PROV_CTX *ctx, CK_TOKEN_INFO *token,
...
+    if (pool->max_sessions <= 0 || pool->max_sessions > 1024) {
         pool->max_sessions = 1024;
-        pool->limit_sessions = 8;
     }
-    if (pool->limit_sessions > pool->max_sessions) {
-        pool->limit_sessions = pool->max_sessions;
+    /* arbitrary max concurrent cached sessions */
+    pool->max_cached_sessions = 8;
+    if (pool->max_cached_sessions > pool->max_sessions) {
+        pool->max_cached_sessions = pool->max_sessions << 8;
```
Jakuje, line 515: this makes max cached sessions 4 for max_sessions 1024, which sounds very few. Is this really the intention or should we keep something more similarly as prior this change?

src/session.c
simo5: I think I wanted to keep things shorter when I originally introduced it.
A bit silly, I know, but I would change it in another PR if it is an issue.

src/session.c
simo5: Not intentional, I was thinking that I would like to have 8 as the result in the default case, then probably got distracted and wrote 8 in the shifting instead of a more appropriate 7 ... or maybe even 6.

I think this code is just too naive tbh, I would love to have a much system where we keep some data on average and peak number of concurrent open operations and then make a decision that make sense given those values. Even 8 may just be a waste if it is just a peak, and the program normally have 2 concurrent operations at most.

I will change this value to 7 now so that defaults and behavior follow the same ratios.

simo5: Think of it, 7 makes no sense either, given here we are in the case where max_session is < 8, so I am just going to use the old logic. I think this whole heuristic will have to be reviewed in another patch

src/session.c
```diff
@@ -493,17 +505,17 @@ static CK_RV session_pool_init(P11PROV_CTX *ctx, CK_TOKEN_INFO *token,
         && token->ulMaxSessionCount != CK_UNAVAILABLE_INFORMATION) {
         pool->max_sessions = token->ulMaxSessionCount;
         /* keep a max of 10% of the sessions */
-        pool->limit_sessions = pool->max_sessions / 10;
-        if (pool->limit_sessions == 0) {
-            pool->limit_sessions = 3;
+        pool->max_cached_sessions = pool->max_sessions / 10;
+        if (pool->max_cached_sessions == 0) {
+            pool->max_cached_sessions = 3;
```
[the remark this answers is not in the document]
simo5: As I mentioned before I think we should handle this later, the code is unchanged except for names fro what was there before

src/session.c
```diff
@@ -843,15 +779,20 @@ static CK_RV check_slot(struct p11prov_slot *provslot, P11PROV_URI *uri,
...
-                                          CK_FLAGS flags)
+static CK_RV fetch_session(P11PROV_SESSION_POOL *pool, CK_FLAGS flags,
+                           P11PROV_SESSION **_session)
 {
     P11PROV_SESSION *session = NULL;
     int ret;
 
+    ret = MUTEX_LOCK(pool);
+    if (ret != CKR_OK) {
+        goto done;
```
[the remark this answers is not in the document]
simo5: good catch!

Jakuje (approved):

src/session.c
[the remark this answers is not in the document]
simo5: I opened #169 to track this

PR #165, opened 2023-01-05
Optimize EC_KEY data parsing and storage

simo5:
Use vendor defined attributes to store additional parameters pre-formatted to OpenSSL liking, this way we can remove a lot of ASN1 reparsing done in multiple places on demand.

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -8,12 +8,15 @@
 #include <openssl/obj_mac.h>
 #include "platform/endian.h"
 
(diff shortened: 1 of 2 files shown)


simo5:
@Jakuje I thought of a way to optimize the curve_name code by using custom vendor CK attributes, and it ended up allowing me to simplify also the code that fetches EC public key data from certificates by eliminating all the ASN1 shenanigans from there.

Let me know what you think.

Jakuje (approved):

simo5:
Thanks a lot.

PR #166, opened 2023-01-06
Key generation callbacks

### LINKED ISSUE #161: No feedback is provided when generating keys
**Describe the feature**
Provide feedback to user while generating keys by using OpenSSL clabback system when possible.

**Expected behavior**
When the token supports it provide user feedback on key generation.

**Additional context**
PKCS11 support the use of CK_CALLBACK_FUNCTION to allow feedback when the token knows long operations are expected.
This could be used to pipe this feedback all the way to OpenSSL callling application by setting appropriate provider callbacks.

simo5:
This PR aims to address #161
I do not say fix because there is no requirement for a token to call callbacks.

In fact neither softokn nor softhsm call them during key generation so we can't regularly test this code.

However I manually tested that everything works well from pkcs11-provider up my hardcoding a few calls to token_session_callback() and running openssl genpkey and have verified that dots are printed by openssl (we return only zeros in the params and they correspond to the '.' char).

Resolves #161

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -421,6 +421,9 @@ struct p11prov_session {
     CK_FLAGS flags;
 
     pthread_mutex_t lock;
+
+    p11prov_session_callback_t cb;
+    void *cbarg;
 };
 
 struct p11prov_session_pool {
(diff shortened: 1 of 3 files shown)


simo5:
This PR depends on #162, and imports its commits, it will be rebased once that PR is pushed.
Marking as draft until #162 is pushed

src/session.c
```diff
@@ -439,19 +442,37 @@ struct p11prov_session_pool {
...
 
+static CK_RV token_session_callback(CK_SESSION_HANDLE hSession,
+                                    CK_NOTIFICATION event,
+                                    CK_VOID_PTR pApplication)
+{
+    P11PROV_SESSION *session = (P11PROV_SESSION *)pApplication;
+
+    if (session->session != hSession) {
+        /* something not right, let's ignore this callback */
+        return CKR_OK;
```
beldmit, line 453: Not sure CKR_OK is a correct behavior here

beldmit (approved):
LGTM with one suspicious decision

src/session.c
simo5: Returning an error would abort the PKCS#11 function that called the callback. I do not think we want to do that.

simo5:
Thanks for the review

PR #167, opened 2023-01-09
Improve authentication to tokens

### LINKED ISSUE #160: Handle tokens with CKF_PROTECTED_AUTHENTICATION_PATH
**Describe the feature**
Currently pkcs11 provider fails to login if a hardware token uses an onboard pinpad or some other out of band authentication method

**Expected behavior**
Authentication via out of band token authentication should work

**Additional context**
The token_login function unconditionally fails if a pin is not available w/o trying a C_Login operation. A C_login operation should be still tried if no pin is available but the token advertises 
CKF_PROTECTED_AUTHENTICATION_PATH

simo5:
Adds support for CKF_PROTECTED_AUTHENTICATION_PATH and prevents repeating authantication with the same failed pin.

Fixes #160

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -10,6 +10,8 @@ struct p11prov_slot {
     CK_SLOT_INFO slot;
     CK_TOKEN_INFO token;
 
+    char *bad_pin;
+
     P11PROV_SESSION_POOL *pool;
 
     CK_MECHANISM_TYPE *mechs;
@@ -249,6 +251,10 @@ void p11prov_free_slots(P11PROV_SLOTS_CTX *sctx)
     for (int i = 0; i < sctx->num; i++) {
         session_pool_free(sctx->slots[i]->pool);
         OPENSSL_free(sctx->slots[i]->mechs);
+        if (sctx->slots[i]->bad_pin) {
+            OPENSSL_clear_free(sctx->slots[i]->bad_pin,
+                               strlen(sctx->slots[i]->bad_pin));
+        }
         OPENSSL_cleanse(sctx->slots[i], sizeof(P11PROV_SLOT));
     }
     OPENSSL_free(sctx->slots);
@@ -689,7 +695,8 @@ CK_SLOT_ID p11prov_session_slotid(P11PROV_SESSION *session)
 
 /* returns a locked login_session if _session is not NULL */
 static CK_RV token_login(P11PROV_SESSION *session, P11PROV_URI *uri,
-                         OSSL_PASSPHRASE_CALLBACK *pw_cb, void *pw_cbarg)
+                         OSSL_PASSPHRASE_CALLBACK *pw_cb, void *pw_cbarg,
+                         struct p11prov_slot *slot)
 {
     char cb_pin[MAX_PIN_LENGTH + 1] = { 0 };
     size_t cb_pin_len = 0;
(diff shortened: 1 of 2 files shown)


simo5:
Marking as draft until the session patches are pushed in #162

Jakuje:
There is a typo in the commit message:
```
not rely on the application to provide a pid (like a physical pin pad, a
```
should say PIN instead of pid.

src/session.c
```diff
@@ -689,44 +691,54 @@ CK_SLOT_ID p11prov_session_slotid(P11PROV_SESSION *session)
...
+        if (uri) {
+            pin = (CK_UTF8CHAR_PTR)p11prov_uri_get_pin(uri);
+            if (slot->bad_pin && strcmp((char *)pin, slot->bad_pin) == 0) {
+                pin = NULL;
+            }
+        }
+        if (!pin) {
+            pin = p11prov_ctx_pin(session->provctx);
+            if (slot->bad_pin && strcmp((char *)pin, slot->bad_pin) == 0) {
+                pin = NULL;
```
Jakuje, line 713: I would probably at least log some debug message in these cases as this might not be something the user is expecting when they spend time debugging why the pin they provided is not being used.

src/session.c
```diff
@@ -689,44 +691,54 @@ CK_SLOT_ID p11prov_session_slotid(P11PROV_SESSION *session)
...
+            }
         }
-        if (cb_pin_len == 0) {
+        if (pin) {
+            pinlen = strlen((const char *)pin);
+        } else if (pw_cb) {
+            const char *info = "PKCS#11 Token";
```
Jakuje, line 719: I understand that this is just moved from existing code, but I think the prompt should have a bit more descriptive label. We are doing the login to the slot so slot label or something will be useful to distinguish which pin the user should provide.

src/session.c
```diff
@@ -735,6 +747,17 @@ static CK_RV token_login(P11PROV_SESSION *session, P11PROV_URI *uri,
                         pinlen);
     if (ret == CKR_USER_ALREADY_LOGGED_IN) {
         ret = CKR_OK;
+    } else {
+        if (pin && ret == CKR_PIN_INCORRECT) {
+            /* mark this pin as bad or we may end up locking the token */
+            OPENSSL_free(slot->bad_pin);
```
Jakuje, line 753: I think the pin should be considered as a sensitive data (regardless if it is good or bad). So it should be probably wiped from the memory before freeing here as well as on cleanup (do not see it freed on the session cleanup either).
One more consideration might be to store sha digest of the pin only here to avoid keeping the pin in memory for long time if not necessary.

src/session.c
simo5: Good idea.

src/session.c
simo5: Have you tried an operation that requires prompting?
Just remove the pin from openssl.cnf and try one of the command that requires a pin and can get it from the user.

simo5: Btw I think the PIN authentication is broken, or I forget which operations work with it (not all).

src/session.c
simo5: This is done only for pins already stored in memory, so I think hashing is a bit excessive?

src/session.c
simo5: Btw this is the prompt OpenSSL shows when it works (follow #172 for more fun):
> Enter pass phrase for PKCS#11 Token:

And now that I re-read your comment I understand you want to see something like:
"PKCS#11 Token (slot 1 - NSS Generic Crypto Services):" using the slot description field ?

simo5: See #173 for this

Jakuje (approved):

PR #173, opened 2023-01-11
Fix prompting for PIN

### LINKED ISSUE #172: Passing PIN on the command line stoped working
**Describe the bug**
OpenSSL has two separate phases in dealing with storemgmt.
It expects to be able to open the store w/o requiring credentials, and then passed any callback (that cna provide passwords) at time of key loading.

When I changed the code to be able to list public objects w/o requiring a login I broke the fallback code that was going to retry reading the store in the load phase when a password is available.

**To Reproduce**
Remove the PIN from the condifuration file and execute anopenssl command that requires the PIN (anything that uses a private key).
No prompting will happen and the command fails to execute with an error about not being able to open the pkcs11 uri passed in.

**Expected behavior**
Prompting happens and the operation completes.

N/A

N/A

**Additional context**
Potentially always deferring the open operation could solve this.

simo5:
While reviewing #167 it was found out that a regression had happened and prompting for a PIN was not working anymore.

Resolve the issue and add tests to prevent future regressions.

Fixes #172

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
           if [ -f /etc/fedora-release ]; then
             dnf -y install git ${{ matrix.compiler }} automake libtool \
               pkgconf-pkg-config autoconf-archive openssl-devel openssl \
-              diffutils
+              diffutils expect
             if [ "${{ matrix.token }}" = "softokn" ]; then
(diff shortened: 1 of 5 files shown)


Jakuje (approved):

simo5:
Thanks for the review @Jakuje

PR #175, opened 2023-01-12
Eliminate use of asprintf

simo5:
Asprintf is convenient but causes mixing of OPENSSL and standard memory management in the code base which can quickly lead to bad problems, undetected until an application changes the openssl allocators.

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -1,7 +1,6 @@
 /* Copyright (C) 2022 Simo Sorce <simo@redhat.com>
    SPDX-License-Identifier: Apache-2.0 */
 
-#define _GNU_SOURCE
 #include "provider.h"
 #include <string.h>
 
@@ -126,6 +125,7 @@ static void trim_padded_field(CK_UTF8CHAR *field, ssize_t n)
 }
 
 #define trim(x) trim_padded_field(x, sizeof(x))
+static const char slot_desc_fmt[] = "PKCS#11 Token (Slot %lu - %s)";
 
 CK_RV p11prov_init_slots(P11PROV_CTX *ctx, P11PROV_SLOTS_CTX **slots)
 {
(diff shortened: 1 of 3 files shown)


src/session.c
```diff
@@ -126,6 +125,7 @@ static void trim_padded_field(CK_UTF8CHAR *field, ssize_t n)
 }
 
 #define trim(x) trim_padded_field(x, sizeof(x))
+const char *slot_desc_fmt = "PKCS#11 Token (Slot %lu - %s)";
```
ueno, line 128:
```suggestion
static const char slot_desc_fmt[] = "PKCS#11 Token (Slot %lu - %s)";
```

Otherwise `sizeof(slot_desc_fmt)` below will always return a pointer size.

src/util.c
```diff
@@ -623,3 +623,41 @@ bool p11prov_x509_names_are_equal(CK_ATTRIBUTE *a, CK_ATTRIBUTE *b)
...
 }
+
+char *p11prov_alloc_sprintf(int size_hint, const char *format, ...)
+{
+    char *buf = NULL;
+    va_list args;
+    int repeat = 1;
+    int ret;
+
+    va_start(args, format);
```
ueno, line 634: Might need a `va_copy` so the following loop work?

By the way, here is a BSD licensed implementation of vasprintf:
https://github.com/p11-glue/p11-kit/blob/d39043f7c6e44247b5b1a237888e80b2a4d9c2b2/common/compat.c#L618

src/session.c
simo5: nice catch
amazing how broken things seemingly work :)

src/util.c
simo5: another nice catch, I wish OpenSSL had a BIO function instead of trying to implement my own ...

simo5: In this case all arguments would be consumed though,
so I wonder if just putting the va_start()/va_end() around the vsnpritf call isn't a simpler option

simo5:
Hopefully both are fixed now.

ueno (approved):
Looks good to me!

simo5:
Thanks!

PR #180, opened 2023-01-24
Fix test as OpenSSL change output formatting

simo5:
For certs, at some point the master tree has changed the way Subjects are printed (remving spaces around the equal signs that separates attribute names from values in DNs).

Fix tests to cope with both older and newer formatting.


beldmit (approved):
LGTM

PR #182, opened 2023-01-30
Ecc siganture fallback for non-raw signature mechanisms

### LINKED ISSUE #168: EC+digest signatures should fall back to raw EC signatures if the specific mechanism is not supported
**Describe the bug**
The SoftHSM supports only `CKM_ECDSA` mechanism, which is raw EC signature. The pkcs11-provider fails when we try to create hashed signature using any digest+signature fails with the following logs:
```
[interface.gen.c:291] p11prov_CopyObject(): Calling C_CopyObject
[objects.c:168] cache_key(): Key 17:25 cached as 27:27
[signature.c:1630] p11prov_ecdsa_set_ctx_params(): ecdsa set ctx params (ctx=0x2164b50, params=(nil))
[signature.c:1454] p11prov_ecdsa_digest_sign_update(): ecdsa digest sign update (ctx=0x2164b50, data=0x451f00, datalen=65)
[interface.gen.c:153] p11prov_GetMechanismInfo(): Calling C_GetMechanismInfo
[interface.gen.c:157] p11prov_GetMechanismInfo(): Error: 0x00000070; Error returned by C_GetMechanismInfo
C_GetMechanismInfo for CKM_ECDSA_SHA512(4166) failed 112
```
The provider should fall back to the `CKM_ECDSA` if the `CKM_ECDSA_SHA*` mechanism is not available by doing the digest inside OpenSSL. This might be problematic with the certifications, but it should improve the imteroperability with mode pkcs11 providers or smart cards.

**To Reproduce**
1. Go to https://gitlab.com/libssh/libssh-mirror/-/merge_requests/334
2. See the latest failed pipeline results

**Expected behavior**
The signatures using ECDSA keys work with CKM_ECDSA mechanism only.

 - OS: Fedora
 - Version 36

 - Device: SoftHSM
 - PKCS11 Driver version: 2.6.1
 - Application: libssh
 - Version: master

**Additional context**
This should also allow us enabling the ecdsa signature tests with softhsm pipeline. The following change:
```
diff --git a/tests/Makefile.am b/tests/Makefile.am
index c3c58aa..87fae4d 100644
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -46,7 +46,7 @@ dist_check_SCRIPTS = \
 test_LIST = \
 	basic-softokn basic-softhsm-proxy \
 	certs-softokn certs-softhsm-proxy \
-	eccsha2-softokn \
+	eccsha2-softokn eccsha2-softhsm-proxy \
 	oaepsha2-softokn \
 	hkdf-softokn \
 	rsapss-softokn \
```
Now gives the same error here:
```
Error signing raw input data                                                     
Public Key operation error                                                       
804B32944F7F0000:error:40800070:pkcs11:p11prov_GetMechanismInfo:An invalid mechanism was specified to the cryptographic operation:interface.gen.c:157:Error returned by C_GetMechanismInfo
804B32944F7F0000:error:40800001:pkcs11:p11prov_sig_operate_init:reason(1):signature.c:605:Failed to open session on slot 17
```

simo5:
Add the ability to fallback to raw signature and separate digest calculation when tokens do not support Sign/Verify with digest calculation functions, but still support plain raw signatures. 

Fixes #168


tests/tecc
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash -e
+# Copyright (C) 2022 Simo Sorce <simo@redhat.com>
```
Jakuje, line 2: nit: 2023 ;)

tests/tecdh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+# Copyright (C) 2022 Simo Sorce <simo@redhat.com>
```
Jakuje, line 2: here too 2023

tests/tecc
```diff
@@ -0,0 +1,57 @@
...
+              -digest sha256
+              -in ${RAND64FILE}
+              -rawin
+              -out ${TMPPDIR}/sha256-ecdgstsig.bin'
+ossl '
+pkeyutl -verify -inkey "${ECBASEURI}" -pubin
+                -digest sha256
+                -in ${RAND64FILE}
+                -rawin
+                -sigfile ${TMPPDIR}/sha256-ecdgstsig.bin'
```
Jakuje, line 49: would it make sense to give a shot to some other digests too, just to get more test coverage?

tests/Makefile.am
```diff
@@ -41,12 +41,13 @@ tmp.softhsm:
...
 	helpers.sh setup-softhsm.sh setup-softokn.sh softhsm-proxy.sh \
-	test-wrapper tbasic tcerts teccsha2 thkdf toaepsha2 trsapss
+	test-wrapper tbasic tcerts tecc tecdh thkdf toaepsha2 trsapss
 
 test_LIST = \
 	basic-softokn basic-softhsm-proxy \
 	certs-softokn certs-softhsm-proxy \
-	eccsha2-softokn \
+	ecc-softokn ecc-softhsm-proxy \
+	ecdh-softokn \
```
Jakuje, line 50: the derive does not work with the softhsm (again?)

Jakuje (commented):
looks much neater than I was expecting from what you were describing yesterday. Just nits. I pushed a change to try this branch on the libssh side to see if it will work:
https://gitlab.com/libssh/libssh-mirror/-/merge_requests/334
Edit: looks like it works as expected on the libssh side:
https://gitlab.com/jjelen/libssh-mirror/-/jobs/3687714156

tests/tecc
simo5: I didn't change contents, just move around stuff :)

src/signature.c
```diff
@@ -601,28 +606,60 @@ static int p11prov_sig_operate_init(P11PROV_SIG_CTX *sigctx, bool digest_op,
...
+        }
+
+        P11PROV_debug("Enable fallback for mechanism %lx", sigctx->mechtype);
+
+        ret = p11prov_get_session(sigctx->provctx, &slotid, NULL, NULL,
+                                  sigctx->mechtype, NULL, NULL, reqlogin, false,
+                                  &session);
+        if (ret != CKR_OK) {
+            P11PROV_raise(sigctx->provctx, ret,
+                          "Failed to open fallback(%lx) session on slot %lu",
```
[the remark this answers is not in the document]
simo5: As silly as this may seem I actually went twice back and forth on this and actually decided for the no space version before pushing :-)

tests/tecdh
simo5: Same, no new code, just moving stuff around

tests/tecc
simo5: yes, I think that is a reasonable thing to do.

tests/Makefile.am
simo5: nope, that is why I split it off

Jakuje: would be worth investigating if it is something in the softhsm, openssl or here to fix. Can you open a new issue for that?

tests/tecc
Jakuje: ok, missed that it was just moved around from the eccsha2. Thanks for clarification

tests/tecc
simo5: added sha384 and sha512, I do not think I can reasonably add more until we have a better way to exclude specific tests.
NSS softokn does not support SHA3 for example.

simo5: I guess if I make additional Digest fallback to OpenSSL digest I might be able to also to SHA3 ... Should we do that as part of this PR ?
It is possible that a token that does not support DigestSign APIs does not support Digest APIs at all after all ...

Jakuje: Sounds like separate PR. SoftHSM will not support SHA3 probably either as far as I know. But its usage in the applications is not very wide so I think it should not be an issue.
In the next PR, I think this can be extended to allow fallback to OpenSSL implementation of digests or disabling this functionality through configuration file if this might be a concern to a certification.

src/signature.c
```diff
@@ -21,6 +21,8 @@ struct p11prov_sig_ctx {
     P11PROV_SESSION *session;
 
     CK_RSA_PKCS_PSS_PARAMS pss_params;
+
+    bool digest_fallback;
```
Jakuje, line 25: last think after couple of reads of this PR, I am not sure about the naming of this variable. Its kind of fallback, but digest fallback sounds confusing/misleading, when this is basically flag saying "digest is calculated separately from the signature/verification operation" so my proposal would be "separate_digest". Or at least have on this line a comment describing what this variable/flag is doing.

Jakuje (approved):

src/signature.c
simo5: I couldn't find a better name indeed, I used digest fallback, because the fallback is about separately calculating the digest, I could have called it mechanism_fallback, because it is about handling an unsupported mechanism.

simo5: See the fixup commit I added.

Jakuje: After fixing the typo, looks good.

simo5:
@Jakuje luckily I tried to do the digest fallback and after 2 attempts I realized that:

a) for digests fallbacks are not necessary at all because openssl providers list them individually and we already have code to expose them only if the token supports them (although we may still have an issue if we have multiple token and they differ in capabilities, I think this is for us to sort at a later time if it happens).

b) openssl already know how to fallback through providers if one does not support a digest.

I used these facts to add digest fallback capability by changing how the signature fallback works from trying to use directly p11prov_DigestInit and friends, to instead use EVP_Digest and asking to "prefer" the pkcs11 provider via "?provider=pkcs11".

I pushed this as an additional commit for now. But if you agree this should go all in this PR I will rebase and merge the initial implementation with the refactoring one in a single commit instead then split the sha3 test into the tests commit and add also sha3-384 and sha3-512.

Jakuje:
I am worried a bit with this direction that we will lose the ability to control which device will perform the digest operation in case there would be more pkcs11 devices. Going through the the OpenSSL might lead to using one device for the signature and completely other for the digest. If we are going to mix up things, I would rather see the OpenSSL internal implementation than some "random" device used and probably have some control over the behavior through some configuration.

simo5:
> 

Uhmm... that may require us to try to do things internally and then call OpenSSL with provider=default ... it is possible but will double the code.

Unless we assume that if we are doing fallback we do not care for digesting on our own device at all and we always unconditionally defer to openssl default provider?

This would come back to byte us if we want to use FIPS devices that support hashes but not the combined operation on the token ... as we'd ask openssl default provider explicitly even though we can keep the operation on a FIPS validated device ...

beldmit:
I'd say that the fetching and its parameters belong to application logic and shouldn't be done on the provider logic.
Is it possible to add some more properties to algorithm objects, e.g. provider=pkcs11,driver=mycooltoken?

dengert:
https://www.rfc-editor.org/rfc/rfc7512 defines the PKCS11 URI that includes "module-name" and "module-path"  so provider could scan for this to get the PKCS11 module to load.

simo5:
> https://www.rfc-editor.org/rfc/rfc7512 defines the PKCS11 URI that includes "module-name" and "module-path" so provider could scan for this to get the PKCS11 module to load.

There is no URI for selecting ciphers, only to reference objects ...

simo5:
Ok it seem that if we can use properties to prefer "same" slot (passing ?pkcs11.slot=number) we can use the fallback.

There is still a bad case where we have multiple slots in a token and they have different capabilities. I think that means we may have to build fallbacks for pretty much all mechanisms ... but I think I want to handle that later.

simo5:
@Jakuje this latest version is now able to prefer pkcs11 tokens but fall back to openssl if the digest is not ssupported at all.

If the pkcs11 provider is chosen we tell it what slotid to use (the same on which the key which we'll later us for signing is stored).

I think this should clear your reservations ?

src/signature.c
```diff
@@ -92,12 +100,27 @@ static void *p11prov_sig_dupctx(void *ctx)
...
+    }
+
     if (sigctx->session == NULL) {
-        goto done;
+        return newctx;
     }
 
     /* This is not really funny. OpenSSL by default assumes contexts with
-     * operations in flight can be easily duplicated, with all the
+      operations in flight can be easily duplicated, with all the
```
beldmit, line 123: Nit: `*` removed by accident

src/signature.c
```diff
@@ -739,6 +847,64 @@ static int p11prov_sig_digest_update(P11PROV_SIG_CTX *sigctx,
...
+    err = p11prov_sig_operate(subctx, sig, siglen, sigsize, digest, mdsize);
+    if (err != RET_OSSL_OK) {
+        ret = CKR_GENERAL_ERROR;
+        P11PROV_raise(sigctx->provctx, ret, "Failure in sigver fallback");
+        goto done;
+    }
+
+done:
+    p11prov_sig_freectx(subctx);
+    return CKR_OK;
```
Jakuje, line 905: this should be `return ret` (with setting the `ret = CKR_OK` before the label) otherwise we are returning CKR_OK even if either of the last two actions fail.

simo5: good catch, changing to `return ret;`, and having *all* error conditions `goto done;`

simo5:
I fixed the issues you both reported, I also added a little bit of refactoring with the return errors because I found an issue where I mixed ossl error and pkcs11 errors and that caused a failure. It took so much time to find out what I did that I decided to decrease the level of mixing removing it from internal functions I touched and pushing it to the boundary (the functions called directly by openssl).

I also rebases on top of latest main.

Jakuje (approved):

src/signature.c
```diff
@@ -92,8 +100,23 @@ static void *p11prov_sig_dupctx(void *ctx)
     newctx->pss_params = sigctx->pss_params;
     newctx->operation = sigctx->operation;
 
+    if (sigctx->mechanism_fallback) {
+        int err;
+        newctx->mechanism_fallback = EVP_MD_CTX_new();
+        if (!newctx->mechanism_fallback) {
```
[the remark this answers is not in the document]
simo5: I prefer the if (!var) format, but do not enforce it.

dengert:
Just a comment... @simo5 @Jakuje  you keep referring to the tokens support things. There is the PKCS11 module that plus the token. No smartcards look anything list PKCS11 In the OpenSC case the PKCS11 module links in libopensc that has PKCS15 interface as well an card drivers that deal with ISO7816-4  APDUs to talk to the token.

simo5:
@dengert as mentioned, the caching was all of the problem. The token had *no* keys, but because of caching it looked like it did.
Once I removed caching from the config it all became clear. I since then generated a new RSA key in 9c and all the operations work as expected.

Btw in term of talking about "pkcs11 token", I know that tokens have pkcs#11 as an interface only and underneath there can be anything else, but that is just an abstraction the provider doesn't care about. From the PoV of the provider, pkcs#11 is the interface and tokens are pkcs#11 tokens.

PR #183, opened 2023-02-04
Simplify utulity to fetch attributes

### LINKED ISSUE #163: p11prov_fetch_attributes may return bad error code/data on failure
**Describe the bug**
If the attribute requested is sensitive or otherwise and is "required" the function may return short w/o processing all data and with the wrong error code (CKR_HOST_MEMORY)

**To Reproduce**
N/A

**Expected behavior**
All available attributes should be processed and returned even on error
P11PROV_debug statements (or even P11PROV_raise for required ones) are emitted so that there is a log of which attributes were required/requested and were not returned

N/A

N/A

**Additional context**
N/A

simo5:
REduce the size and reuse the CK_ATTRIBUTE structure within fetch_attrs to further simplify copying data between attribute storages when needed.

Also fix some checks in the derive key functions that I found wrong while refactoring.

Fixes #163

### DIFF
--- a/src/exchange.c
+++ b/src/exchange.c
@@ -212,7 +212,7 @@ static int p11prov_ecdh_derive(void *ctx, unsigned char *secret,
     CK_KEY_TYPE key_type = CKK_GENERIC_SECRET;
     CK_BBOOL val_true = CK_TRUE;
     CK_BBOOL val_false = CK_FALSE;
-    CK_ULONG key_size = 0;
+    CK_ULONG key_size = outlen;
     CK_ATTRIBUTE key_template[5] = {
         { CKA_CLASS, &key_class, sizeof(key_class) },
         { CKA_KEY_TYPE, &key_type, sizeof(key_type) },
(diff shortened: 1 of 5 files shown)


src/exchange.c
```diff
@@ -266,9 +260,14 @@ static int p11prov_ecdh_derive(void *ctx, unsigned char *secret,
 
     /* complete key template */
     if (ecdhctx->kdf_outlen) {
-        key_size = ecdhctx->kdf_outlen;
+        if (ecdhctx->kdf_outlen < outlen) {
+            key_size = ecdhctx->kdf_outlen;
+        }
```
Jakuje, line 265: with this change, the `key_size` is kept on 0 in case of `ecdhctx->kdf_outlen >= outlen` instead of an error that was raised previously if I see right, which does not look correct.

simo5: Uggh good catch.
I am not completely focused today, but I believe I should have fixed this now

Jakuje (approved):

PR #199, opened 2023-02-16
Fix broken debug output to stderr

manison:
This fixes crash when debug log file is not specified in the `PKCS11_PROVIDER_DEBUG` environment variable (e.g. `PKCS11_PROVIDER_DEBUG=level:99`). The bug has been introduced in commit b39e45e9.

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -54,6 +54,9 @@ void p11prov_debug_init(void)
         if (dbg_level < 1) {
             dbg_level = 1;
         }
+        if (stddebug == NULL) {
+            stddebug = stderr;
+        }
     }
 
 done:


simo5:
@manison hey good catch!

The DCO check complains your github account and your sign-off do not match, but looks like just that github does not know your identity or the DCO bot can't get it, so I'll just push through, it is a trivial change anyway, if you confirm the sign-off is correct.

manison:
@simo5 I don't know how to convince the DCO bot that the sign-off is correct. Based on the failure details the bot would probably be happy when using the anonymous noreply sign-off. I can change that if you want. Otherwise I declare the sign-off to be correct :)

PR #200, opened 2023-02-16
Error early if RSA PSS is not supported

### LINKED ISSUE #194: Implement software fallback from PSS to raw RSA signatures or return errors early
**Describe the feature**
OpenSSL does separate keymgmt for RSA vs RSA-PSS "keys" (which is not even a thing in pkcs#11), but implements a single provider for RSA signatures, and assumes both PKCS!.5 and PSS signatures are always doable ...

**Expected behavior**
Given OpenSSL does not understand that a provider may be able to perform PKCS signatures, but not PSS signatures, we need to implement software fallback if a token support RSA raw signatures.

**Additional context**
It may be ok to return an error as soon as a PSS padding mode is set via rsasig_set_ctx_params instead of performing ful emulation.

simo5:
I decided the simplest hing is to return early if the token does not support RSA PSS rather then trying to emulate it via raw RSA, because usually the pkcs11 driver emulates PSS on its own if raw RSA is available at all.
If it is not it is generally because the token is supporting only PKCS in Hardware and can't do PSS at all.

Fixes #194

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -332,6 +332,35 @@ int p11prov_slot_get_mechanisms(P11PROV_SLOT *slot, CK_MECHANISM_TYPE **mechs)
     P11PROV_raise(provctx, ret, "%s %s mutex (errno=%d)", _errstr, obj, err); \
     P11PROV_debug("Called from [%s:%d]%s()", file, line, func)
 
+int p11prov_check_mechanism(P11PROV_CTX *ctx, CK_SLOT_ID id,
+                            CK_MECHANISM_TYPE mechtype)
+{
+    P11PROV_SLOTS_CTX *sctx;
+    CK_RV ret;
+
+    ret = p11prov_take_slots(ctx, &sctx);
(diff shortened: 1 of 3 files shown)


Jakuje (approved):
I think this is a reasonable approach and assumption. After TLS 1.3 went out, the PSS started to be mandatory and now there are 4 places where to do the padding. Either in the card itself, in the pkcs11 module, in the pkcs11 provider or in the application itself. If the pkcs11 provider is out of the list, it can still be worked around in adjacent layers

Disabling SW PSS padding in pkcs11 provider should probably make it much less complicated and error-prone and until we will have a real-world user for this, we can probably live without that.

simo5:
Thanks for the review, my reasoning was exactly similar to yours, if need be we can always revise this later, for now let's defer to the pkcs11-driver if the card does not do it itself, I know opensc-pkcs11 can do PSS if raw RSA ia available, and we can do the fallback for the digest part if needed.

PR #202, opened 2023-02-18
Delay pkcs11 module loading and initialization

### LINKED ISSUE #187: C_Initialize called twice causes deadlock in futex_wait
**Describe the bug**
When an application that calls `C_Initialize` and also use OpenSSL where `openssl.cnf`  defines the `pkcs11-provider` the provider is initialized and calls C_Initialize a second time.

The `pkcs11-tool` loads  module `opensc-pkcs11.so` which uses `libopensc.so.8` which uses `libcrypto.so.3` 
which does 
The provider then load module `ossl-modules/pkcs11.so` that loads module `opensc-pkcs11.so`

If the deadlock did not occur this could lead to a loop. 

**To Reproduce**

`gdb --args ./pkcs11-tool -O` 

**Expected behavior**
`pkcs11-provider` does not initialize unless specifically requested, or is able to use existing PKCS11 initialized by application.  

 - OS: Xubuntu 22.04 as (VirtualBox guest with Windows 11 host)
 - Version 22.04
 - OpenSSL-3.0.8
 - OpenSC  form github master
 - pkcs11-provider from github main yesterday.

 - Device: Yubikey 5 NFC (PIV) 
 - PKCS11 Driver version: From OpenSC 
 - Application pkcs11-tool
 - Version OpenSC master 

**Additional context**

[pissue-opensc-debug.txt](https://github.com/latchset/pkcs11-provider/files/10698208/pissue-opensc-debug.txt)
Note that a few lines may be missing at end. 

[pissue-opensc-gdb.txt](https://github.com/latchset/pkcs11-provider/files/10698256/pissue-opensc-gdb.txt)

[pissue-openssl.cnf-diff.txt](https://github.com/latchset/pkcs11-provider/files/10698328/pissue-openssl.cnf-diff.txt)

Possible circumvention: only use openssl.cnf with provider when use with OpenSSL untilities.

simo5:
This patchset delays pkcs11 modules initialization until an application actually causes openssl to try to use pkcs11-provider for some operation.

This requires to make keymgmt static, because keymgm and store interfaces can be called directly by openssl before any other fnction causes module initialization.

All direct entrypoints now must call pkcs11_ctx_status() which causes module intialization and returns an error if it fails.

Fixes #187

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -253,7 +253,7 @@ CK_RV p11prov_module_new(P11PROV_CTX *ctx, const char *path,
 CK_RV p11prov_module_init(P11PROV_MODULE *mctx)
 {
     P11PROV_SLOTS_CTX *slots;
-    CK_C_INITIALIZE_ARGS args;
+    CK_C_INITIALIZE_ARGS args = { 0 };
     CK_INFO ck_info = { 0 };
     CK_RV ret;
 
--- a/src/interface.c
+++ b/src/interface.c
@@ -127,7 +127,9 @@ static void populate_interface(P11PROV_INTERFACE *intf, CK_INTERFACE *ck_intf)
     ASSIGN_FN(GenerateKeyPair);
     ASSIGN_FN(DeriveKey);
     ASSIGN_FN(GenerateRandom);
-    ASSIGN_FN_3_0(GetInterface);
+    if (intf->version.major == 3) {
+        ASSIGN_FN_3_0(GetInterface);
+    }
 }
 
 CK_RV p11prov_interface_init(void *dlhandle, P11PROV_INTERFACE **interface,
--- a/src/interface.c
+++ b/src/interface.c
@@ -9,6 +9,16 @@
  * at runtime by emulating or returning the proper error if a 3.0
  * only function is used on a 2.40 token. */
 
+struct p11prov_module_ctx {
+    P11PROV_CTX *provctx;
+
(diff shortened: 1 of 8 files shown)


simo5:
@dengert PTAL, I think you'll like this direction.

Jakuje (approved):
Direction looks good. The pkcs11 side also looks sensible. I do not follow all the openssl-side changes but they look reasonable too.

simo5:
I just tested that this indeed fixes pkcs11-tool to not end up into a loop and mutex lock.
So I'll merge.

@dengert let me know if there is more we can do, but I think this should resolve most of the issues with double pkcs11 initialization seen so far.

dengert:
With pkcs11-provider in  `ssl/openssl.cnf` all of Mouse's tests run as expected. OpenSC was built with  https://github.com/OpenSC/OpenSC/pull/2712 and some additional changes to pkcs11.tool to use OpenSSL libctx set to "default" which I started to work on.

I am going to rebuild OPenSC with https://github.com/OpenSC/OpenSC/pull/2712 to see if all works without it.

dengert:
Works without https://github.com/OpenSC/OpenSC/pull/2712 and my pkcs11-tool mode too!

simo5:
I know this work w/o OPenSC 2712, but I think it is extremely important to merge 2712 in OpenSC, will continue discussion there

PR #210, opened 2023-03-01
Add support for EDDSA

### LINKED ISSUE #145: Add support for EdDSA keys
The RSA and ECDSA are the most common keys these days in the PKCS#11 world, but the EdDSA is becoming also more and more popular. There are HSMs supporting these keys for some time ([yubiHSM](https://developers.yubico.com/YubiHSM2/Product_Overview/)) as well as some smart cards/tokens based on OpenPGP ([NitroKey](https://shop.nitrokey.com/shop/product/nksa-nitrokey-start-6) with the gnuk applet). It will also become more important when the [FIPS 186-5](https://csrc.nist.gov/publications/detail/fips/186/5/draft) will get approved.

simo5:
This adds initial support for handling EDDSA keys and signatures.

I haven't tested EDDSA key generation.

And I had to disable the signature tests (although checked manually that the code actually works against the softhsm implementation), because without the proxy softhsm segfaults in openssl destructors and with the proxy the proxy does not pass through CKM_EDDSA.

Fixes #145


Jakuje:
The p11-kit has a PR for support for PKCS#11 3.0 ready for some time, but not yet merged: p11-glue/p11-kit#374

src/keymgmt.c
```diff
@@ -1184,6 +1198,282 @@ const OSSL_DISPATCH p11prov_ec_keymgmt_functions[] = {
...
+        ret = OSSL_PARAM_set_int(p, secbits);
+        if (ret != RET_OSSL_OK) {
+            return ret;
+        }
+    }
+    p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_MAX_SIZE);
+    if (p) {
+        int sigsize;
+        if (group_size == 456) {
+            sigsize = 114;
```
Jakuje, line 1391: Would it make sense to reuse the constants `ED448_SIG_SIZE` and `ED25519_SIG_SIZE` here too?

simo5: Oh yeah,
as I was developing I sprinkled these macros a bit everywhere, and then missed backtracking and checking all uses for consistency.
I am going to revisit the commits, pull all the macros in a single header and use them consistently.

src/signature.c
```diff
@@ -2009,3 +2049,274 @@ const OSSL_DISPATCH p11prov_ecdsa_signature_functions[] = {
...
+{
+    P11PROV_CTX *ctx = (P11PROV_CTX *)provctx;
+    P11PROV_SIG_CTX *sigctx;
+
+    sigctx = p11prov_sig_newctx(ctx, CKM_EDDSA, properties);
+    if (sigctx == NULL) {
+        return NULL;
+    }
+
+    return sigctx;
```
Jakuje, line 2073: nit: Both of the branches do the same so this could be simplified to a single line:
```suggestion

    return p11prov_sig_newctx(ctx, CKM_EDDSA, properties);
```
unless we would want to do some more stuff here, keep the same structure as with other mechanisms or do some logging on error (OOM warning or something).

src/signature.c
```diff
@@ -2009,3 +2049,274 @@ const OSSL_DISPATCH p11prov_ecdsa_signature_functions[] = {
...
+        default:
+            ret = RET_OSSL_ERR;
+            break;
+        }
+        if (ret != RET_OSSL_OK) {
+            return ret;
+        }
+    }
+
+    return RET_OSSL_OK;
```
Jakuje, line 2201: nit: This also looks like over-complicated/obfuscated way of writing the following (if we initialize the `ret = RET_OSSL_OK` in the beginning:

```suggestion
    }

    return ret;
```

src/signature.c
```diff
@@ -2009,3 +2049,274 @@ const OSSL_DISPATCH p11prov_ecdsa_signature_functions[] = {
...
+}
+
+#ifndef OSSL_SIGNATURE_PARAM_INSTANCE
+#define OSSL_SIGNATURE_PARAM_INSTANCE "instance"
+#endif
+#ifndef OSSL_SIGNATURE_PARAM_CONTEXT_STRING
+#define OSSL_SIGNATURE_PARAM_CONTEXT_STRING "context-string"
+#endif
```
Jakuje, line 2209: Are these missing in some older version of openssl than 3.0.7? If so, if we bump the required version we can drop these. If there is different reason to include them, comment saying why and how long they are needed would be worth placing here.

src/signature.c
```diff
@@ -2009,3 +2049,274 @@ const OSSL_DISPATCH p11prov_ecdsa_signature_functions[] = {
...
+        ret = OSSL_PARAM_get_utf8_string_ptr(p, &instance);
+        if (ret != RET_OSSL_OK) {
+            return ret;
+        }
+        P11PROV_debug("Set OSSL_SIGNATURE_PARAM_INSTANCE to %s", instance);
+
+        size = p11prov_obj_get_key_bit_size(sigctx->key);
+        if (size != ED25519_BIT_SIZE && size != ED448_BIT_SIZE) {
+            P11PROV_raise(sigctx->provctx, CKR_KEY_TYPE_INCONSISTENT,
+                          "Invalid EdDSA key size %lu\n", size);
```
Jakuje, line 2237: the log/error messages do not need the trailing newline as it is added by the logging functions already

tests/setup-softhsm.sh
```diff
@@ -282,6 +305,10 @@ export ECPEERBASEURI="${ECPEERBASEURI}"
 export ECPEERPUBURI="${ECPEERPUBURI}"
 export ECPEERPRIURI="${ECPEERPRIURI}"
 export ECPEERCRTURI="${ECPEERCRTURI}"
+export EDBASEURI="${EDBASEURI}"
```
Jakuje, line 308: missing `export EDBASEURIWITHPIN="${EDBASEURIWITHPIN}"` on the line above

src/signature.c
simo5: I actually thought about this, and ultimately chose not to do that

src/signature.c
simo5: Initially the function was more complicated, then it reduced to what is now, no need to keep it complicated. Fixing.

src/signature.c
simo5: But given you also found this annoying, I will change it

src/signature.c
simo5: They are missing in 3.0.7 (they are upstream only)

simo5: We've done this for other stuff, I am fine with keeping these until they are not necessary anymore, they are harmless.

src/signature.c
simo5: good catch

tests/setup-softhsm.sh
simo5: none of the EC stuff seem to export it
and we do not use it, so I will just drop it for now

Jakuje (approved):
lgtm

Issue #133, opened 2022-11-30
Find way to test fallback to pull public key out of certificate


simo5:
The code that extract public key info can fallback to pull that information out of a certificate if there is no corresponding public key object in the token.
Unfortunately it is not easy to add a certificate w/o pkcs12-util also covertly extracting and adding a public key object to softokn.

dengert:
Can you delete the pubkey after install?  Or will softokn fail without a
pubkey?

simo5:
I haven't found a tool that allows you to delete just the public key.
I can probably write a small C program to do that, but that was beyond writing a quick test.

simo5:
And then there is softhsm, I was hoping to do create the right objects in the setup script.

Jakuje:
With softhsm you should be able to delete the public key through the pkcs11 interface with `pkcs11-tool --delete-object`. I have POC (which is failing) on top of #144 here:

https://github.com/latchset/pkcs11-provider/compare/main...Jakuje:pkcs11-provider:softhsm-pubkey
https://github.com/latchset/pkcs11-provider/commit/25f1c0acc8d48b745288e033d95aad9bdbdb7b3d

there are couple of problems:
 * all the tests use the same key store so removing the public key affects all the following tests. It can be moved to the setup though
 * the NSS requires the parameters to C_Initialize so it does not work with the `pkcs11-tool` -- for that some other steps might be needed
 * reading the public keys fails now in the session test:
```
$ cat tests/session-softhsm-proxy.log
Executing /home/jjelen/devel/pkcs11-provider/tests/softhsm-proxy.sh ./tsession

## Start the p11-kit server and check if it works
Using slot 0 with a present token (0x11)
Failed to load public key
killing p11-kit server
FAIL session-softhsm-proxy (exit status: 1)
```

Leaving this as WIP for now

Jakuje:
@simo5 I rebased and updated my branch from previous comment to add tests for reading a public keys from certificates, but neither of the attempts worked for me. The base URI, Priv, Pub and Cert URI just keep failing with the bogus errors like described in #212. The object looks like created successfully inside of the provider, but it looks like OpenSSL does not like it in the end somewhere around this place (this run was with base URI):
```
0x00007ffff7c0232c in OSSL_STORE_load (ctx=ctx@entry=0x5555556713c0) at crypto/store/store_lib.c:428
428	            if (!ctx->fetched_loader->p_load(ctx->loader_ctx,
Value returned is $10 = 1
[...]

PR #213, opened 2023-03-03
Draft: Test reading pubkey from certificate

Jakuje:
and few assorted fixes that can land separately.

Related to #133


src/objects.c
```diff
@@ -576,6 +576,8 @@ CK_RV p11prov_obj_from_handle(P11PROV_CTX *ctx, P11PROV_SESSION *session,
     CK_BBOOL token_supports_allowed_mechs = CK_TRUE;
     CK_RV ret;
 
+    P11PROV_debug("called");
+
```
simo5, line 580: not sure we really need this

src/objects.c
```diff
@@ -1179,6 +1182,8 @@ static CK_RV get_all_attrs(P11PROV_OBJ *obj, CK_ATTRIBUTE *attrs, int num)
     CK_ATTRIBUTE *res[num];
     CK_RV rv;
 
+    P11PROV_debug("called (obj=%p, atrs=%p, num=%d)", obj, attrs, num);
+
```
simo5, line 1186: This is redundant as get_public_attrs never modifies the data, only selects where to get it from.

src/objects.c
```diff
@@ -1205,6 +1210,8 @@ static CK_RV get_public_attrs(P11PROV_OBJ *obj, CK_ATTRIBUTE *attrs, int num)
     P11PROV_OBJ *tmp = NULL;
     CK_RV rv;
 
+    P11PROV_debug("called (obj=%p, atrs=%p, num=%d)", obj, attrs, num);
```
simo5, line 1213: called -> Get Public Attributes: ?

src/session.c
```diff
@@ -509,6 +509,8 @@ static CK_RV token_session_open(P11PROV_SESSION *session, CK_FLAGS flags)
     uint64_t startime = 0;
     CK_RV ret;
 
+    P11PROV_debug("called (session=%p, flags=%x)", session, flags);
+
```
simo5, line 513: Not sure how useful this one is, but if you think we should keep it, we need to note thatthe session is the session object pointer, not the session handle

src/session.c
```diff
@@ -682,6 +684,8 @@ static CK_RV session_check(P11PROV_SESSION *session, CK_FLAGS flags,
     CK_SESSION_INFO session_info;
     int ret;
 
+    P11PROV_debug("called (session=%p, flags=%x)", session, flags);
```
simo5, line 687: same here

src/session.c
```diff
@@ -759,6 +763,9 @@ static CK_RV token_login(P11PROV_SESSION *session, P11PROV_URI *uri,
     CK_ULONG pinlen = 0;
     CK_RV ret;
 
+    P11PROV_debug("called (session=%p, uri=%p, slot=%p, pw_cb=%p)", session,
```
simo5, line 766: probably need renaming if it is a keeper

src/session.c
```diff
@@ -828,12 +835,16 @@ static CK_RV token_login(P11PROV_SESSION *session, P11PROV_URI *uri,
 
 done:
     OPENSSL_cleanse(cb_pin, cb_pin_len);
+    P11PROV_debug("returning 0x%x", ret);
```
simo5, line 838: don't think we need this, we should have debug in the caller if ret is an error.

src/session.c
```diff
@@ -874,6 +885,8 @@ static CK_RV fetch_session(P11PROV_SESSION_POOL *pool, CK_FLAGS flags,
     P11PROV_SESSION *session = NULL;
     int ret;
 
+    P11PROV_debug("called (pool=%p, flags=%x)", pool, flags);
```
simo5, line 888: don't think we need this

tests/tpubkey
```diff
@@ -0,0 +1,72 @@
...
+source ${TESTSSRCDIR}/helpers.sh
+
+TSTOUT="${TMPPDIR}/testout"
+
+title PARA "Export RSA Public key to a file"
+ossl 'pkey -in $BASEURI -pubin -pubout -out ${TSTOUT}.pub'
+title LINE "Export Public key to a file (pub-uri)"
+ossl 'pkey -in $PUBURI -pubin -pubout -out ${TSTOUT}.pub'
+title LINE "Export Public key to a file (with pin)"
+ossl 'pkey -in $BASEURIWITHPIN -pubin -pubout -out ${TSTOUT}.pub'
```
simo5, line 14: I wonder if we should move this to the tbasic section where we remove the pin from the configuration in order to be useful ..

simo5 (changes requested):
I think most if this is quite good, however there are a bunch of additional debug statement that seem excessive.
If you think these are useful to debug harder to see issues, I think you should debug at level 2 or even 3. See how debugging of mechanisms is done when we load slots to figure out the higher level debugging.

tests/tpubkey
Jakuje: good point. Will move

src/objects.c
Jakuje: without these logs, I was unable to track what is going on in the provider as there were just very rough. For this case, I could see that the PKCS#11 functions got called, but I did not see where from, if they came from the right context and so on. I do not mind downgrading the log level though.

simo5:
I think I understand what is going on with the certificate issue.
You are telling openssl to expect a public key in, so that's what openssl expects from the store.

I think we can support that, but we'd need to do that directly in the store based on the expect flag.

I am thinking for now we should perhaps test with something like:
openssl x509 -pubkey -noout -in CERTURI -out pemfile ?

simo5:
As for exercising the fallback I think the operation to use is one where you specify just the private key to openssl, but then the code internally needs to access also the public key.
Perhaps ECDH ?

simo5:
@Jakuje so I think the -pubin is the actual issue here,
this manual test worked:
/tests/softhsm-proxy.sh openssl pkey -in $PRI2URI -pubout -out ${TSTOUT}-priv-cert.pub

tests/tpubkey
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash -e
+# Copyright (C) 2022 Jakub Jelen <jjelen@redhat.com>
+# SPDX-License-Identifier: Apache-2.0
+
+source ${TESTSSRCDIR}/helpers.sh
+
+TSTOUT="${TMPPDIR}/testout"
```
simo5, line 7: Btw this is not ok, it makes copy/pasting command to run on the command line use an undefined variable which means the manual test I ran crated a file named "-priv-cert.pub" in the current directory.

simo5: Just use ${TMPDIR}, non need for a separate directory for outputs...

tests/tpubkey
```diff
@@ -0,0 +1,72 @@
...
+    echo
+    exit 1
+fi
+
+# The softokn does not support removing only public key from the key pair
+if [ ! -z "$BASE2URI" ]; then
+    title PARA "Check we can get public keys from certificate objects"
+
+    title LINE "Export Public key to a file (priv-uri)"
+    ossl 'pkey -in $PRI2URI -pubin -pubout -out ${TSTOUT}-priv-cert.pub'
```
simo5, line 61: remove -pubin from this one

tests/tpubkey
```diff
@@ -0,0 +1,72 @@
...
+    title LINE "Export Public key to a file (crt-uri)"
+    ossl 'pkey -in $CRT2URI -pubin -pubout -out ${TSTOUT}-cert.pub'
```
simo5, line 63: This is not currently possible I think, as a cert is not a key, and openssl pkey handles only keys.

tests/tpubkey
simo5: Unfortunately this does not use the certificate fallback, because RSA Private keys include data for the public key, so there is no fallback.

It may work right using EC keys as private EC key PKCS#11 objects generally do not include public key material in their attributes

tests/tpubkey
Jakuje: This was just moved from `tests/tbasic`. But fine to remove that.

tests/tpubkey
Jakuje: Yeah. And I think this is the use case we are interested in, which would require some conversion. The certificate itself contains the public key, so the provider should be able to pull this information and serve it to the calling application if the separate public key is missing int he token.

tests/tpubkey
Jakuje: Right, the privuri is probably not useful as from the EC private key we can not get the public key. But again, we could get it from the certificate, if it exists and serve it to the calling application.

Jakuje:
With the current state, it looks like it works ok. I will probably remove the "disputed" logging so we can move on with this test to not block other stuff and I will open separate issue/pr for the logging.

But summary is that indeed, it works if I try to read the public key from the RSA private key URI or base URI without any conversion. I think the point of #133 is pulling the public key from the certificate transparently and probably also for EC keys, which now does not work as discussed above and as commented out in the test now.

src/objects.c
Jakuje: dropped

src/session.c
Jakuje: dropped for now

tests/tpubkey
simo5: Ah, then we should fix tests/tbasic as well

tests/tpubkey
simo5: It is exaclty this that makes EC useful, you pass a private key, then try to extract the public key, the fallback code will try to find an associated public key, fail, then try an associated certificate, succeed, and try to extract the public key from the certificate.

tests/tpubkey
simo5: No, it is the other one we are interested in (from private), as we are testing the fallback code to extract public from private.

simo5:
The EC case will test exactly what we want, the RSA case does not (I know I ran it in GDB) because the token always includes the public bits in the private key object as well (unlike for EC).

tests/tpubkey
Jakuje: This PR moves this part of the code from the `tbasic` to `tpubkey`

tests/tpubkey
Jakuje: So let me extend this to the EC.

simo5:
@Jakuje this looks good for merge.
You just need to rebase on master to pick up the fix for the democa test bug on mac and resolve the conflict

Jakuje:
> @Jakuje this looks good for merge. You just need to rebase on master to pick up the fix for the democa test bug on mac and resolve the conflict

Sorry for a delay. Rebased.

Jakuje:
Currently, there are two failures, but neither of them look like something that would be reproducible and I do not see a button to rerun them.

simo5:
@Jakuje I restarted them, sound like softhsm has flkey tests, I wonder if we need a sleep after starting the proxy, to make sure the process is up before the following commands execute...

simo5:
ah it is not a proxy delay, sounds more like killing the proxy sometimes triggers error due to set -e?
```
killing p11-kit server
FAIL pubkey-softhsm-proxy (exit status: 1)
```

simo5:
The macos build failed again, I am re-runnig a third time, but this seems like something that actually needs to be fixed before we can merge.

simo5:
> now, it worked.

Yes, but I am afraid this will introduce a constant flakiness issue ...

simo5:
I fixed the conflict and force pushed.
If it passes tests I will merge, otherwise I'll see if I can find the problem with the flakey tests

simo5:
Uhmm this time it failed with:
```
Export Public key to a file (base-uri)
openssl pkey -in $ECBASE2URI -pubout -out ${TMPPDIR}/ec-base-cert.pub
0056361501000000:error:40800000:pkcs11:find_associated_obj:pkcs11:objects.c:837:Error in C_FindObjects (count=0)
0056361501000000:error:40800012:pkcs11:p11prov_GetAttributeValue:Invalid attribute type specified in a template:./interface.gen.c:410:Error returned by C_GetAttributeValue
0056361501000000:error:40800012:pkcs11:p11prov_GetAttributeValue:Invalid attribute type specified in a template:./interface.gen.c:410:Error returned by C_GetAttributeValue
0056361501000000:error:0800008D:elliptic curve routines:group_new_from_name:invalid curve:crypto/ec/ec_lib.c:1492:
```

which seems a real error ?

Jakuje:
Please, try to rerun the job. There are also couple of lines like
```
p11-kit: 'bound != NULL' not true at fixed0_C_CloseSession
p11-kit: 'bound != NULL' not true at fixed0_C_Finalize
(p11-kit:9419) message: child 9425 died with sigsegv
```
coming from p11-kit on mac. I think I saw them before, but only on mac and I am not particularly sure what to do with these. @ueno or @neverpanic do you have some idea?

simo5:
@Jakuje those lines are seen for every command, so it seems like that in this case there was an actual error.
The main issue though, is that this test will make any future PR always fail several times. If we can't fix it on mac we should disable the test there.

ueno:
Does it happen if p11-kit is compiled with libffi support (i.e., no limit of the number of closures)?

Jakuje:
> Does it happen if p11-kit is compiled with libffi support (i.e., no limit of the number of closures)?

Not sure. The installation is from brew so it should be this one: https://github.com/Homebrew/homebrew-core/blob/master/Formula/p11-kit.rb so probably the libffi is present.

neverpanic:
Last time a test only failed on macOS it was a buffer overflow. The p11-kit segfault does indeed happen for every test, and I don't know why and haven't investigated it. Homebrew uses libffi from the OS on the system we test on, but I also saw the same thing with p11-kit from MacPorts, which links against its own copy of libffi.

Can we test this with AddressSanitizer?

simo5:
@neverpanic I think having an asan test wouldn't be bad, do you have a guide on how to enable it on mac ?

neverpanic:
> @neverpanic I think having an asan test wouldn't be bad, do you have a guide on how to enable it on mac ?

Any guide for clang that works on Linux basically works in the same way on macOS. The canonical source is https://clang.llvm.org/docs/AddressSanitizer.html.

To save on those precious macOS CPU minutes, I'd add that ASAN test on Linux, though. Last time it was also a general problem, Linux just ended up 0-initializing more memory and didn't crash.

ueno:
> The installation is from brew so it should be this one: https://github.com/Homebrew/homebrew-core/blob/master/Formula/p11-kit.rb so probably the libffi is present.

OK, I forgot that the current p11-kit code always uses the fixed closures and falls back to libffi only when the fixed ones are exhausted. Given the reproducibility I suspect this might be related to locking (in particular the initialization of `p11_virtual_mutex`, which makes use of `PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP`).

simo5:
Thanks I think this is now in a reasonable state and can be merged, any follow can come in another PR as needed.

PR #216, opened 2023-03-08
Add test to exercise a full TLS connection

### LINKED ISSUE #215: A test for using PKCS11 provider in TLS
**Describe the feature**
We need a test of openssl s_client/s_server communication using PKCS#11 token as a key storage and pkcs11 provider as a handshake digest implementation source. The digest context duplication must not be implemented for the digest. OpenSSL master branch required

**Expected behavior**
Handshake and data exchange passes.

**Additional context**
pkcs11 devices may support digest but not support digest context duplication. This behavior should be tested

simo5:
This uses certificates and keys from the PKCS#11 token

Fixes #215


simo5:
@jakuje initially I used the RSA cert/key  but softhsm was failing with:
[signature.c:449] p11prov_sig_pss_restrictions(): Error: 0x0000001B; mechanism not allowed with this key

Is there something special to be done to use an RSA key with "PSS" in softhsm ?

I changed the test to use the EC Cert/Key to avoid this issue for now.

beldmit (approved):
LGTM

Jakuje:
> @Jakuje initially I used the RSA cert/key but softhsm was failing with: [signature.c:449] p11prov_sig_pss_restrictions(): Error: 0x0000001B; mechanism not allowed with this key

AFAIK SoftHSM knows only some basic RSA-PSS parameters. That is the reason why we do not run the `rsapss` with SoftHSM: https://github.com/latchset/pkcs11-provider/blob/main/tests/Makefile.am#L59

> Is there something special to be done to use an RSA key with "PSS" in softhsm ?

AFAIK the simplest thing would be the fallback to raw rsa in the provider when the PSS parameters are not accepted. But I am not sure if it is possible this late in the process when we got `CKR_ACTION_PROHIBITED`.

simo5:
@Jakuje I need to check if ti is too late, it is definitely annoying to have to emulate PSS though, and we do not set restrictions afaik, so I am not even sure why this is returned. Is it possible missing support by the proxy process?

simo5:
Looks like all tests are now passing again.

PR #218, opened 2023-03-11
Add support for setting/removing error stack marks

### LINKED ISSUE #212: The `store_fetch()` logic is (still) flawed
**Describe the bug**
Debugging an issue with getting public keys from certificates, I am hitting weird issues with SoftHSM, which I believe are caused by a flawed logic in the `store_fetch()` function.

The `store_fetch()` iterates through all slots, tries to get a session in that slot and then read objects. This works ok for "good" slots, but SoftHSM intentionally provides a "bad" slot, which has `CKF_TOKEN_PRESENT`, but the token does not have `CKF_TOKEN_INITIALIZED` flag, which makes the `check_slot()` function in `p11prov_get_session()` immediately fail with `CKR_TOKEN_NOT_PRESENT`, which is then propagated to the `store_fetch()`, leading to the error `Failed to get session to load keys` and immediate return from the function without checking any further slots. This has the following implications:

 * if the "bad" slot is listed before the "good" slot, the provider will never try to read from the "good" slot
 * it might lead to random failures depending on the slot order
 * The error "Failed to get session to load keys" not being fatal for the store might be confusing when debugging other issues.

**To Reproduce**
1. Run tests with softhsm
2. Observe pkcs11 provider debug logs
3. See the `Failed to get session to load keys` error

**Expected behavior**
The above `error` should probably be only warning or debug message. It should be error only if the only slot that is available fails like this or some similar heuristic. There should be probably better debug messages around the store handling.

 - OS: Fedora
 - Version 36

 - Device: SoftHSM

simo5:
This allows to suppress errors stored on the stack in places where we can retry operations, so we reduce the amount of noise reported to users in case of failure.

Fixes #212

### DIFF
--- a/src/provider.c
+++ b/src/provider.c
@@ -359,6 +359,10 @@ static OSSL_FUNC_core_get_params_fn *core_get_params = NULL;
 static OSSL_FUNC_core_new_error_fn *core_new_error = NULL;
 static OSSL_FUNC_core_set_error_debug_fn *core_set_error_debug = NULL;
 static OSSL_FUNC_core_vset_error_fn *core_vset_error = NULL;
+static OSSL_FUNC_core_set_error_mark_fn *core_set_error_mark = NULL;
+static OSSL_FUNC_core_clear_last_error_mark_fn *core_clear_last_error_mark =
+    NULL;
+static OSSL_FUNC_core_pop_error_to_mark_fn *core_pop_error_to_mark = NULL;
 
 static void p11prov_get_core_dispatch_funcs(const OSSL_DISPATCH *in)
 {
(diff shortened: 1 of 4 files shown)


Jakuje:
should we drop the else branch as mentioned by @dengert in https://github.com/latchset/pkcs11-provider/issues/212#issuecomment-1464146425

Jakuje (approved):

simo5:
> should we drop the else branch as mentioned by @dengert in [#212 (comment)](https://github.com/latchset/pkcs11-provider/issues/212#issuecomment-1464146425)

I was thiking this may be fixed in a different PR, but on second thought, I will just add another commit there, given I have to fix some stuff with the style anyway.

PR #220, opened 2023-03-29
Add options for caching

simo5:
Allow to cache pins so that login session can be automatically re-established if needed.


src/session.c
```diff
@@ -395,24 +395,22 @@ static CK_RV token_login(P11PROV_SESSION *session, P11PROV_URI *uri,
...
-                pin = NULL;
-            }
+        }
+        if (!pin && cached_pin) {
+            pin = (CK_UTF8CHAR_PTR)cached_pin;
+        }
+        if (pin && bad_pin && strcmp((char *)pin, bad_pin) == 0) {
+            P11PROV_raise(session->provctx, CKR_PIN_INVALID,
+                          "Blocking stored PIN that failed a previous login"
+                          " to avoid blocking the token");
```
Jakuje, line 412: maybe update this log message that it can now affects also cached pin.

src/session.c
```diff
@@ -436,6 +434,16 @@ static CK_RV token_login(P11PROV_SESSION *session, P11PROV_URI *uri,
...
+
+            if (p11prov_ctx_cache_pins(session->provctx)) {
+                CK_RV trv;
+                trv = p11prov_slot_set_cached_pin(slot, (const char *)pin);
+                /* not much we can do on failure */
+                if (trv != CKR_OK) {
+                    P11PROV_raise(session->provctx, trv, "Failed to cache pin");
+                }
+            }
+
```
Jakuje, line 446: My initial though was that we should probably cache the pin only after the login is successful. But the current implementation look like not causing any harm. Maybe just to avoid some noise, when the pin would be cached, but login will fail and on another attempt it will be reported as a bad pin.

tests/tbasic
```diff
@@ -151,4 +151,12 @@ fi
 OPENSSL_CONF=${ORIG_OPENSSL_CONF}
 rm -f ${OPENSSL_CONF}.nopin
 
+title PARA "Test PIN caching"
+ORIG_OPENSSL_CONF=${OPENSSL_CONF}
+sed "s/^pkcs11-module-token-pin.*$/pkcs11-module-cache-pins = cache/" ${OPENSSL_CONF} > ${OPENSSL_CONF}.pincaching
+OPENSSL_CONF=${OPENSSL_CONF}.pincaching
+${TESTBLDDIR}/pincache
```
Jakuje, line 158: is the absolute path needed here? Other binaries are executed with relative path as `./pincache` as we should be in the `TESTBLDDIR`.

src/provider.c
```diff
@@ -531,6 +533,18 @@ int p11prov_ctx_login_behavior(P11PROV_CTX *ctx)
...
 
+bool p11prov_ctx_cache_pins(P11PROV_CTX *ctx)
+{
+    P11PROV_debug("cache_pins = %s", ctx->cache_pins ? "true" : "false");
+    return ctx->cache_pins;
+}
+
+int p11prov_ctx_cache_keys(P11PROV_CTX *ctx)
+{
+    P11PROV_debug("cache_keys = %d", ctx->cache_keys);
```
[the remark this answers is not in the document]
simo5: There is only one value, maybe later if it grows :)

src/objects.c
```diff
@@ -325,6 +325,13 @@ static void cache_key(P11PROV_OBJ *obj)
     CK_SESSION_HANDLE sess;
     CK_BBOOL can_cache = CK_TRUE;
     CK_RV ret;
+    int cache_keys;
+
+    /* check whether w should cache keys in session at all */
```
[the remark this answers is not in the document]
simo5: Reworded differently

src/provider.c
```diff
@@ -1218,74 +1248,84 @@ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
...
+                          p11prov_cfg_names[P11PROV_CFG_LOGIN_BEHAVIOR],
+                          cfg[P11PROV_CFG_LOGIN_BEHAVIOR]);
             p11prov_ctx_free(ctx);
             return RET_OSSL_ERR;
         }
     }
 
-    if (load != NULL && strcmp(load, "early") == 0) {
+    if (cfg[P11PROV_CFG_CACHE_PINS] != NULL
+        && strcmp(cfg[P11PROV_CFG_CACHE_PINS], "cache") == 0) {
```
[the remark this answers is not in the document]
simo5: I do not want to use a boolean in the config because I think we may see more than 2 options. the code is boolean because that is an internal detail and we can change it later at will.

src/session.c
simo5: I am not sure why, if the pin fails it is never cached in the first place.

src/session.c
simo5: We should cache only if successful, it is what I intended.
Will fix the code.

tests/tbasic
simo5: Yes it is needed, I initially used ./pincache and it failed in make distcheck

src/session.c
simo5: While looking at this I also realized I did not free the cached_pin in the slot free function, fixed that as well.

Jakuje (approved):

PR #227, opened 2023-04-13
Support OSSL_PKEY_PARAM_DEFAULT/MANDATORY_DIGEST

simo5:
This is needed to inform OpenSSL that EDDSA must not use a digest, and also allow openssl to inquire what the default digest is, if any.


Jakuje (approved):
lgtm

PR #229, opened 2023-04-14
Fix SoftHSM issues wth PSS keys

### LINKED ISSUE #217: TLS 1.3 exchange with RSA-PSS fails to work with softhsm
**Describe the bug**
I used the RSA cert/key but softhsm was failing with:
[signature.c:449] p11prov_sig_pss_restrictions(): Error: 0x0000001B; mechanism not allowed with this key

**To Reproduce**
1. Change the tests/ttls s_server test to use the RSA certs instead of EC
2. Run make check 
3. softhsm test fails

**Expected behavior**
softhsm shouldn;t fail

 - OS: Fedora
 - Version 37

 - Device: softhsm
 - PKCS11 Driver version: softhsm-2.6.1-5.fc37.5.x86_64

simo5:
Looks like SoftHSM sometimes returns an empty CKA_ALLOWED_MECHANISMS attributes, which clearly makes no sense.

For right now we work around this, if no mechanisms are really allowed a following operation will simply fail anyway.

Fixes #217

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -432,7 +432,7 @@ static CK_RV p11prov_sig_pss_restrictions(P11PROV_SIG_CTX *sigctx,
 
     if (allowed_mechs) {
         CK_ATTRIBUTE_TYPE *mechs = (CK_ATTRIBUTE_TYPE *)allowed_mechs->pValue;
-        int num_mechs = allowed_mechs->ulValueLen;
+        int num_mechs = allowed_mechs->ulValueLen / sizeof(CK_MECHANISM_TYPE);
         bool allowed = false;
 
         if (num_mechs == 0) {
--- a/src/signature.c
+++ b/src/signature.c
@@ -435,6 +435,14 @@ static CK_RV p11prov_sig_pss_restrictions(P11PROV_SIG_CTX *sigctx,
         int num_mechs = allowed_mechs->ulValueLen;
         bool allowed = false;
 
+        if (num_mechs == 0) {
+            /* It makes no sense to return 0 allowed mechanisms for a key,
+             * this just means the token is bogus, let's ignore the check
+             * and try the operation and see what happens */
+            P11PROV_debug("Buggy CKA_ALLOWED_MECHANISMS implementation");
+            return CKR_OK;
+        }
+
         for (int i = 0; i < num_mechs; i++) {
             if (mechs[i] == mechanism->mechanism) {
                 allowed = true;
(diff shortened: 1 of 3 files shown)


src/signature.c
```diff
@@ -435,6 +435,14 @@ static CK_RV p11prov_sig_pss_restrictions(P11PROV_SIG_CTX *sigctx,
         int num_mechs = allowed_mechs->ulValueLen;
         bool allowed = false;
 
+        if (num_mechs == 0) {
+            /* It makes no sense to return 0 allowed mechanisms for a key,
+             * this just means the token is bogus, let's ignore the check
+             * and try the operation and see what happens */
```
Jakuje, line 441: I would say that the `CKA_ALLOWED_MECHANISMS` is underspecified. I think when I implemented this in SoftHSM, returning empty array was closest thing to expressing "anything is allowed"/"there are no limitations".

Jakuje (approved):

src/signature.c
simo5: the specification says:
```
A list of mechanisms allowed to be used with this key. The number of mechanisms in the array is the ulValueLen component of the attribute divided by the size of CK_MECHANISM_TYPE.
```

If you  return an attribute at all you are indicating this list, if you return an empty list you are saying nothing is allowed ... I think this is a bug in softhsm.

simo5: But I am glade for this bug, as I just saw that my code to then go through the list is broken :-D

simo5: Just to be clear, the correct behavior to mean "anythign is allowed" is to not return this attribute at all.
Alternatively return all the known mechanisms related to this key.

Jakuje: I think it is still not completely clear from the specs. Right, there is s paragraph:

> 1.     If the specified attribute (i.e., the attribute specified by the type field) for the object cannot be revealed because the object is sensitive or unextractable, then the ulValueLen field in that triple is modified to hold the value CK_UNAVAILABLE_INFORMATION.

but it talks only about "technical" issues to read the attribute. But thinking about it further, you are probably right that getting `CK_UNAVAILABLE_INFORMATION` would be better.

Jakuje: oh, The allowed mechanisms was not the part I was implementing, I think. Its just the list provided by the one who generated the key or imported the key through pkcs11 so if there were no provided in the templates, it returns just the empty array ... so in the end, its probably the creator responsiblity, but yeah, softhsm should probably handle this corner case better.

simo5: Most key generation code does not want to restrict algorithms so they set nothing, and in that case softhsm should return CK_UNAVAILABLE_INFORMATION.
I think returning 0 length array is never expected and improper behavior.

PR #230, opened 2023-04-18
Improve URI management

simo5:
Overhaul how URI are parsed.
Add missing attributes specified in [RFC 7512](https://www.rfc-editor.org/rfc/rfc7512)
Print URIs instead of ID/Label in text encoders

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -145,15 +145,40 @@ void p11prov_fetch_attrs_free(struct fetch_attrs *attrs, int num)
     }
 }
 
+#define ATTR_library_description "library-description"
+#define ATTR_library_manufacturer "library-manufacturer"
+#define ATTR_library_version "library-version"
+#define ATTR_token "token"
+#define ATTR_manufacturer "manufacturer"
+#define ATTR_model "model"
+#define ATTR_serial "serial"
+#define ATTR_slot_description "slot-description"
+#define ATTR_slot_id "slot-id"
+#define ATTR_slot_manufacturer "slot-manufacturer"
+#define ATTR_id "id"
+#define ATTR_object "object"
+#define ATTR_type "type"
+
 struct p11prov_uri {
-    char *model;
-    char *manufacturer;
+    /* CK_INFO attributes */
+    char *library_description;
+    char *library_manufacturer;
+    CK_VERSION library_version;
+    /* CK_TOKEN_INFO attributes */
     char *token;
+    char *manufacturer;
+    char *model;
     char *serial;
+    /* CK_SLOT_INFO attributes */
+    char *slot_description;
+    CK_SLOT_ID slot_id;
+    char *slot_manufacturer;
+    /* object attributes */
     CK_ATTRIBUTE id;
-    CK_ATTRIBUTE label;
+    CK_ATTRIBUTE object;
+    CK_OBJECT_CLASS type;
+    /* pin */
     char *pin;
-    CK_OBJECT_CLASS class;
 };
 
 static int hex_to_byte(const char *in, unsigned char *byte)
@@ -177,10 +202,10 @@ static int hex_to_byte(const char *in, unsigned char *byte)
     return 0;
 }
 
-static int parse_attr(const char *str, size_t len, unsigned char **output,
+static int parse_attr(const char *str, size_t len, uint8_t **output,
                       size_t *outlen)
 {
-    unsigned char *out;
+    uint8_t *out;
     size_t index = 0;
     int ret;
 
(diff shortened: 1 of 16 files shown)


simo5:
@mouse07410 you may want to try this PR and execute:
```
openssl storeutl -keys -text pkcs11:
```

This way you will see exactly what URI the pkcs11-provider reconstruct from looking at keys.
You can use model/manufacturer/token/serial pkcs11 attributes to print only the keys of a specific token.

simo5:
Note that although I added support for recognizing these additional pkcs11 URI attributes:
- library-description
- library-manufacturer
- library-version
- slot-description
- slot-id
- slot-manufacturer

At the moment they are ignored and can't be used to filter down keys.

I am on the fence on whether I should allow all them to be used.
I see nothing wrong in allowing it at least for the slot ones (except slot id ..).
It is a bit silly to do it for the library ones, but if it turns out to be easy I'll do that as well.
Neither slot or library attributes are use to print out URIs, they are uncommon and rarely useful attributes anyway.
Only Token and Object attributes are utilized to print object URIs.

src/util.c
```diff
@@ -287,32 +320,261 @@ static int get_pin_file(const char *str, size_t len, char **output,
...
+        ret = EINVAL;
+        goto done;
+    }
+    sep = strchr(str, '.');
+    if (!sep) {
+        ret = EINVAL;
+        goto done;
+    }
+
+    /*major */
```
Jakuje, line 450: nit
```suggestion
    /* major */
```

Jakuje (approved):

src/util.c
simo5: Thanks, fixed!

Jakuje (approved):
lgtm.
Maybe notes regarding test coverage:
 * the uri parsing functions would be grate candidate for some extensive unit tests
 * can we have some automated tests with the `openssl storeutl` to check the reported uri as a whole is sensible?

simo5:
Yeah, I will hold on merging until I have a test specific to URIs.
Just need to decide what should be tested, if you have ideas I would like to know.

Jakuje:
I would go ahead with some corner cases, percent encoding, non-printable bytes when using the `storeutil` and grepping for the known parts of the uris in the outputs. No more specific thoughts.

dengert:
> Just need to decide what should be tested, if you have ideas I would like to know.

@Jakuje  could @simo5  use the `regress/unittests/pkcs11/tests.c` you wrote in 2017 for OpenSSH as discussed in OpenSC:
https://github.com/OpenSC/OpenSC/issues/2741#issuecomment-1506407965 and I was trying to port your patch to OpenSSH current It has some 56 tests.  

 https://github.com/dengert/openssh-portable/blob/pkcs11-URI/regress/unittests/pkcs11/tests.c

Jakuje:
Sure. Thats a good idea. I would not say it is complete coverage or it is perfect, but it can be certainly used as a start or at least for ideas for some weird uris I could think of back in that time.

simo5:
We do not really have unit tests at this stage yet, mostly because the vast majority of the code requires a configured token to work at this time.

So I'll add some testing just to make sure the convesion back and forth works.

dengert:
The OpenSSH and `regress/unittests/pkcs11/tests.c`  does not test against a token.  But there are a lot of tests of making URIs. 
The tests are run as part of actions/workflow  https://github.com/openssh/openssh-portable/blob/master/.github/workflows/c-cpp.yml

OpenSC sets up a virtual PIV token https://github.com/OpenSC/OpenSC/blob/master/.github/test-piv.sh  
and virtual OpenPGP  token. https://github.com/OpenSC/OpenSC/blob/master/.github/test-openpgp.sh 
Could also use SoftHSM 

@Jakuje  is familiar with the OpenSC actions.  I only look at them if something fails.

simo5:
@Jakuje I did what I could using just openssl storeutl as a proxy to probe URIs.

Perhaps in future we'll be able to add more complete unit tests, but it looks fine for nopw, and this test will catch simple regressions.

simo5:
I think I am happy with this the way it is for now

mouse07410:
> @mouse07410 you may want to try this PR and execute:
> 
> ```
> openssl storeutl -keys -text pkcs11:
> ```
> 
> This way you will see exactly what URI the pkcs11-provider reconstruct from looking at keys. You can use model/manufacturer/token/serial pkcs11 attributes to print only the keys of a specific token.

First - **thank you**, it works and shows keys on any one token. Here's an example of one key (out of several printed out) for YubiHSM2:
```text
10: Pkey
PKCS11 RSA Private Key (3072 bits)
[Can't export and print private key data]
PKCS11 RSA Public Key (3072 bits)
Modulus:
    00:c7:04:b3:a2:b2:eb:15:45:de:34:74:23:0c:fb:
    52:6a:75:fa:18:ad:aa:25:72:5a:0a:91:3e:45:3b:
    .  .  .
    53:8a:58:9a:e4:3e:22:c0:60:b5:23:e7:1a:de:12:
    98:39:00:6a:a3:0c:79:f6:15:51
Exponent: xxxxxxxx (0xxxxxxxxx)
Key ID:
    04:02
Label: RSA-OAEP
```

So far, everything it printed out was correct and matched my expectations for what it should find on the given token in the given slot/key-id.

But it only confirms that at least this part works correctly, and does not offer any hint why some of the operations fail.
Do you suggest integrating it into a test-script somehow?

PR #233, opened 2023-04-27
Implement support to expose PKCS#11 Random as openssl rand provider

### LINKED ISSUE #184: Support PKCS#11 RNGs
Most of PKCS#11 tokens support RNGs. We need to use these RNGs as a part of openssl RNG support (probably just seeding and/or separate RNG)

simo5:
Fixes #184

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -39,4 +39,4 @@ jobs:
     - name: Check the Style
       run: make check-style || (make check-style-show; exit -1)
     - name: Check spelling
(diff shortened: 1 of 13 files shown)


mouse07410:
I like it!

beldmit (approved):
LGTM.

I'm not sure about reseeding and its role in the process.

simo5:
This is just a start, we'll probably do some more touches if we sart using this facility more, but for now I just wanrted to make sure we *can* use it.

PR #236, opened 2023-05-02
Add tunable to influence session caching

### LINKED ISSUE #169: Refactor max session [cache] calculations
In src/sessions.c we try to come up with some numbers to use to limit the amount of sessions we use at the same time and the number we can keep open as a cache.

This code really needs to be rethought, as these numbers are pulled out of thin air and the calculations are rough and inconsistent.

simo5:
This reworks a bit session caching to be a little more resonable and gives the user a way to influence it if needed.

Fixes #169

### DIFF
--- a/src/provider.c
+++ b/src/provider.c
@@ -30,6 +30,7 @@ struct p11prov_ctx {
     int login_behavior;
     bool cache_pins;
     int cache_keys;
+    int cache_sessions;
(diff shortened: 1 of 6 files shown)


Jakuje (approved):
lgtm

simo5:
Thank you for the review!

PR #243, opened 2023-05-23
Run tests under valgrind, address-sanitizer and address several memory-related issues

Jakuje:


src/signature.c
```diff
@@ -1212,10 +1212,9 @@ static int p11prov_rsasig_digest_sign_final(void *ctx, unsigned char *sig,
 {
     P11PROV_SIG_CTX *sigctx = (P11PROV_SIG_CTX *)ctx;
 
-    P11PROV_debug(
-        "rsa digest sign final (ctx=%p, sig=%p, siglen=%zu, "
-        "sigsize=%zu)",
-        ctx, sig, *siglen, sigsize);
+    P11PROV_debug("rsa digest sign final (ctx=%p, sig=%p, siglen=%p, "
```
simo5, line 1215: Maybe we should just initialize the output to 0 rather then changing what we print ?

simo5 (commented):
Looks good, great addition!

simo5:
The output of the failing tls-softokn test is amusing but not useful :-D

src/signature.c
Jakuje: Given that this is output variable, I am not sure what would be the point in initializing it to 0 and then printing it. Yeah, I agree that printing the pointer is not much more helpful, but I am ok with either direction (as long as we will not dereference null pointer or print uninitialized variables).

src/objects.c
```diff
@@ -1581,6 +1581,16 @@ int p11prov_obj_export_public_ec_key(P11PROV_OBJ *obj, OSSL_CALLBACK *cb_fn,
 
     n = 0;
     if (key_type == CKK_EC) {
+        /* Need to add a null terminator here when converting the byte array to
```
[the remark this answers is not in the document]
Jakuje: I am not sure. The functions under `get_public_attrs()` do not have any idea what is the type of the gathered object and what the caller plans to do with it (pass it to openssl in different format). I see there is already some byte added in `p11prov_fetch_attributes()` to the allocated size so I guess I will have to investigate this a bit more.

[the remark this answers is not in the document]
simo5: Well actually CKA_P11PROV_CURVE_NAME is a synthetic attribute we generate in pre_process_ec_key_data(), and we already allocate 1 more byte, and assign it via OPENSSL_strdup() which generates a zero terminated buffer, so .. why do we need this again ?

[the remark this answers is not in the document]
simo5: Is this happening when we get CKA_P11PROV_CURVE_NAME from a certificate in get_all_from_cert() ?
Sounds like we should fix it there if that is the case...

src/signature.c
simo5: Printing 0 after assigning it still tells the reader what values we have in there, better than printing a pointer. Doesn't really matter that we assigned that value.

Jakuje: ok. Will do.

src/objects.c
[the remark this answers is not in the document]
Jakuje: Sounds like this is the case and it comes from the certificate. I changed the allocation there and it looks ok now.

Jakuje:
Summary now:
 * Works, but on Fedora, the memory leak from softhsm is not suppressed for some reason (will have to investigate probably later).
 * There are also couple of places where some memory handling was intentionally or unintentionally left leaking (see comments in the code). We might need to keep them and suppress them or fix them (or figure out how to minimize the possible issues)
 * There is some housekeeping needed to generate useful traces under address sanitizer, which might need some more comments -- let me know if there is something unclear

tests/setup-softhsm.sh
```diff
@@ -335,6 +335,7 @@ export OPENSSL_CONF="${OPENSSL_CONF}"
 export SOFTHSM2_CONF=${BASEDIR}/${TMPPDIR}/softhsm.conf
 export TESTSSRCDIR="${TESTSSRCDIR}"
 export TESTBLDDIR="${TESTBLDDIR}"
+export PORT_INCREMENT=10
```
simo5, line 338: Why do we need this?
Are you planning to run tests in parallel?

src/interface.c
```diff
@@ -289,7 +289,13 @@ CK_RV p11prov_module_init(P11PROV_MODULE *mctx)
     P11PROV_debug("PKCS#11: Initializing the module: %s", mctx->path);
 
     dlerror();
+
+#if P11PROV_ADDRESS_SANITIZER
```
simo5, line 293: Probably better to do something like this in provider.h:
```
#if P11PROV_ADDRESS_SANITIZER
#define P11PROV_DLOPEN_FLAGS RTLD_NOW | RTLD_LOCAL
#else
#define P11PROV_DLOPEN_FLAGS RTLD_NOW | RTLD_LOCAL | RTLD_DEEPBIND
#endif
```

Then this in the code:
`mctx->dlhandle = dlopen(mctx->path, P11PROV_DLOPEN_FLAGS0);`

src/util.h
```diff
@@ -4,6 +4,15 @@
...
 
+#if __SANITIZE_ADDRESS__
+#define P11PROV_ADDRESS_SANITIZER 1
+#endif
+#if defined(__has_feature)
+#if __has_feature(address_sanitizer)
+#define P11PROV_ADDRESS_SANITIZER 1
+#endif
+#endif
+
```
simo5, line 15: This sounds like it should go in provider.h ?

src/objects.c
```diff
@@ -523,7 +523,8 @@ CK_ULONG p11prov_obj_get_key_size(P11PROV_OBJ *obj)
 void p11prov_obj_to_reference(P11PROV_OBJ *obj, void **reference,
                               size_t *reference_sz)
 {
-    *reference = p11prov_obj_ref_no_cache(obj);
+    /* XXX this would leak memory */
+    *reference = obj;  //p11prov_obj_ref_no_cache(obj);
     *reference_sz = sizeof(P11PROV_OBJ);
```
simo5, line 528: Does not look right to me, what will prevent freeing the object while openssl still holds a pointer to it?

simo5:
Some minor nits and some questions, but this is definitely good.

tests/setup-softhsm.sh
Jakuje: It started as part of debugging the issues as the tests kept the processes running as everything was significantly slow (especially under valgrind). It can be probably dropped though. Let me give it a try.

src/interface.c
Jakuje: Good point. Initially I thought we might need some more overrides, but it looks lkie there are none needed now.

The third option is to use `LD_PRELOAD` to drop this flag in the function wrapper as we already override the dlclose(). This is something I considered, but found too intrusive, but not touching the code. What do you think?

src/objects.c
Jakuje: I think the handles are still referenced in the internal context object list (as you use it for refresh and similar stuff).

Does the openssl unref the reference somehow when it stops using this? I did not find a code that would take this reference and unref it. This keeps the objects in the memory after we free the provider context.

The other option might be somehow forcing the free on cleanup, regardless the amount of handed references.

src/util.h
Jakuje: moved there for now

tests/setup-softhsm.sh
Jakuje: Removed. Locally works ok. Lets see the CI.

src/interface.c
simo5: we already have a dlclose() wrapper as you said, adding another one is ok, and keeps the code cleaner.
but either way is fine by me.
Once flags defines are moved to a header the code is also clean enough.

src/objects.c
simo5: I think the right thing would be to unref when we know openssl is done with the object.

If you are passing params to a function, just clean up the params after the function is done?

Jakuje: Here is the example of the leak from the basic test:
```
Direct leak of 112 byte(s) in 1 object(s) allocated from:                                               
    #0 0x7f44dc0ba6af in __interceptor_malloc (/usr/lib64/libasan.so.8.0.0+0xba6af)                     
    #1 0x7f44dbbaf14c in CRYPTO_zalloc (/lib64/libcrypto.so.3+0x1af14c)                                 
    #2 0x7f44dac69050 in p11prov_obj_new /home/jjelen/devel/pkcs11-provider/src/objects.c:250           
    #3 0x7f44dac6e7c0 in p11prov_obj_from_handle /home/jjelen/devel/pkcs11-provider/src/objects.c:800   
    #4 0x7f44dac70a89 in p11prov_obj_find /home/jjelen/devel/pkcs11-provider/src/objects.c:978          
    #5 0x7f44dac9f11b in store_fetch /home/jjelen/devel/pkcs11-provider/src/store.c:135                 
    #6 0x7f44dac9fc1c in p11prov_store_load /home/jjelen/devel/pkcs11-provider/src/store.c:249          
    #7 0x7f44dbc0232b in OSSL_STORE_load (/lib64/libcrypto.so.3+0x20232b)                               
    #8 0x55fcedc074db in load_key_certs_crls_suppress apps/lib/apps.c:972                               
    #9 0x55fcedc087e6 in load_key_certs_crls apps/lib/apps.c:1099                                       
    #10 0x55fcedc087e6 in load_key apps/lib/apps.c:560                                                  
    #11 0x55fcedbe3a33 in init_ctx apps/pkeyutl.c:549                                                   
    #12 0x55fcedbe3a33 in pkeyutl_main apps/pkeyutl.c:305                                               
    #13 0x55fcedbd9500 in do_cmd apps/openssl.c:418                                                     
    #14 0x55fcedbb9711 in main apps/openssl.c:298                                                       
    #15 0x7f44db84a50f in __libc_start_call_main (/lib64/libc.so.6+0x2750f)                             
    #16 0x7f44db84a5c8 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x275c8)                      
    #17 0x55fcedbb99e4  (/usr/bin/openssl+0x489e4)                                                      
```
I went ahead to check how the references are managed. We are getting this reference through the store API, which allocates the internal object structure and then saves them to the store context. The `p11prov_store_load` then gives away the references, which might or might not be exported to the `EVP_PKEY`s through the `p11prov_common_load()` function, which references the key again.

I think that by storing the key objects in the store context marks the lifetime of the references handed over to the application. Once the store context is closed, the references should no longer be valid (unless it was exported to EVP_PKEY -- in that case, it is indeed kept alive with the reference counter). This is correctly implemented in `p11prov_store_ctx_free()` which decrements the references to the keys, freeing the unused and keeping the exported (which are still references by key management objects). Before this change, this does not work and keys are left in memory infinitely.

Or do I miss something here?

src/interface.c
Jakuje: Fixed.

src/objects.c
simo5: Your argument is convincing.
Would you mind changing the name of the function to p11prov_obj_to_store_reference() and remove the commented parts.
Add a comment that say we intentionally do not take a ref because it is already held by the store.
Then I think we are good.
I want to change the name to make it clear that is is meant to be used only with the store and not as a generic function (the rest already use p11prov_obj_ref() anyway.

simo5 (approved):
this is just great!
LGTM!

(only nit the reference function comment)

src/objects.c
Jakuje: Renamed function added comment, reordered the last commit so the CI tests are last.

Jakuje:
All issues addressed, the CI is still green after couple of reruns so I believe we are good to go!

simo5:
Absolutely fantastic!

beldmit:
Nice work!

Issue #205, opened 2023-02-20
Add support for unnamed EC


manison:
**Describe the feature**
Currently the provider only supports standard named elliptic curve objects. This feature would allow loading EC keys with nonstandard EC parameters.

**Expected behavior**
The provider is loading EC keys with nonstandard EC parameters.

**Additional context**
Currently we use libp11 engine and OpenSSL 1.1 to work with our custom EC. We would like to migrate to OpenSSL 3 and this provider but this missing feature is one of the things that prevent us to migrate. I would try to prepare PR. Would this feature be accepted?

manison:
I would like to. Now I have some work-in-progress that needs heavy polishing and I'm struggling with integrating into test suite.

PR #245, opened 2023-05-30
Initial support for explicit EC

manison:
This is initial PoC for #205 for review purposes.


simo5:
I know I created curve name and pub key attributes for convenience, but I think it is just too much to create that many attributes that are used only in the export function.

I would rather revert and remove those convenience attributes and explicitly decode CKA_EC_PARAMS where needed rather than extending their use this way.

manison:
I have rewritten the export function to not use the pseudo attributes. Now it decodes the CKA_EC_PARAMS as needed. Can you take a look whether it is the right direction?

src/objects.c
```diff
@@ -1555,53 +1555,263 @@ const char *p11prov_obj_get_ec_group_name(P11PROV_OBJ *obj)
...
+        P11PROV_raise(obj->ctx, CKR_GENERAL_ERROR, "Invalid curve");
+        return RET_OSSL_ERR;
+    }
+
+    if (!ossl_param_construct_bn(&params[(*nparam)++], export_ctx,
+                                 OSSL_PKEY_PARAM_EC_P, p)
+        || !ossl_param_construct_bn(&params[(*nparam)++], export_ctx,
+                                    OSSL_PKEY_PARAM_EC_A, a)
+        || !ossl_param_construct_bn(&params[(*nparam)++], export_ctx,
+                                    OSSL_PKEY_PARAM_EC_B, b)) {
```
simo5, line 1668: please P11PROV_raise here as well

simo5: or better do it in ossl_param_construct_bn() and mention for what key the operation failed.

src/objects.c
```diff
@@ -1555,53 +1555,263 @@ const char *p11prov_obj_get_ec_group_name(P11PROV_OBJ *obj)
     return (const char *)attr->pValue;
 }
 
+struct obj_export_ec_ctx {
+    BN_CTX *bnctx;
+    void *allocs[6];
```
simo5, line 1560: Not a huge fan of very opaque structures like this, do I understand correctly that this is done just to keep a reference to the allocated temporary values until the export is complete?

Why not simply go through params matching and freeing the buffer for specific keys so that this whole export_ctx becomes unnecessary ?

src/objects.c
```diff
@@ -1555,53 +1555,263 @@ const char *p11prov_obj_get_ec_group_name(P11PROV_OBJ *obj)
...
+        rv = get_public_attrs(obj, attrs, 1);
+        if (rv != CKR_OK) {
+            P11PROV_raise(obj->ctx, rv, "Failed to get EC key curve nid");
+            return RET_OSSL_ERR;
+        }
+        curve_nid = *(int *)attrs[0].pValue;
+        OPENSSL_free(attrs[0].pValue);
+        attrs[0].type = CKA_P11PROV_PUB_KEY;
+        attrs[1].type =
+            curve_nid == NID_undef ? CKA_EC_PARAMS : CKA_P11PROV_CURVE_NAME;
```
simo5, line 1763: please use a proper if else construct, readability > compactness

simo5 (commented):
Just nits, but this version is much better than the previous

simo5:
Sorry for late review, I was on vacation.

manison:
Thanks for the review. No need to be sorry, hope you enjoyed your vacation :) I'll walk through your comments and prepare updated revision of the PR.

src/encoder.c
```diff
@@ -466,22 +466,28 @@ const OSSL_DISPATCH p11prov_rsa_encoder_spki_pem_functions[] = {
 /* ECDSA */
 
 struct ecdsa_key_point {
-    ASN1_OBJECT *curve;
+    void *curve;
```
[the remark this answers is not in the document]
manison: I employed a union, but I'm not convinced it helped readability too much. What do you think?

src/objects.c
manison: Done.

src/objects.c
manison: Done as you suggested.

src/objects.c
manison: Done.

manison:
I pushed another update, can you please take a look? Thanks.

src/encoder.c
```diff
@@ -466,17 +466,25 @@ const OSSL_DISPATCH p11prov_rsa_encoder_spki_pem_functions[] = {
 /* ECDSA */
 
 struct ecdsa_key_point {
-    ASN1_OBJECT *curve;
+    union {
+        ASN1_OBJECT *object;
+        ASN1_STRING *sequence;
+        void *p;
```
simo5, line 472: why do you need p?

you can always just set k->curve.object = NULL when you need to clear this.

src/encoder.c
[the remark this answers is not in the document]
simo5: looks ok to me

simo5:
Looks mostly good, but I do a full re-review in the next few days.

simo5 (changes requested):
LGTM
Only style fixes and the p attribute in the union should be fixed, otherwise I am good with this.

manison:
Done. Should I prepare the final PR? I suppose I will squash everything into a single commit (both code changes and tests). Are you ok with that?

simo5:
> Done. Should I prepare the final PR? I suppose I will squash everything into a single commit (both code changes and tests). Are you ok with that?

I think we are good.
Either one commit or one for code and one for tests, up to you.

manison:
The PR is ready, I ended up making it two commits.

The SEGV from LSAN might be bug in LSAN itself (see e.g. https://github.com/skupperproject/skupper-router/issues/206, https://github.com/google/sanitizers/issues/1322, https://bugzilla.mozilla.org/show_bug.cgi?id=1635327 and lot more). I cannot reproduce locally and in my fork it fails rather randomly.

simo5:
I am rerunning the test, let's see if it is a flake or if more analysis is needed

simo5 (approved):

simo5:
Well, what do you know, it is flakey,
That is not great, but I see no reason to stop merging your contribution.

And thanks a lot for your patience, I really believe the code was improved 10 folds from first submission and really like the result!

manison:
> And thanks a lot for your patience, I really believe the code was improved 10 folds from first submission and really like the result!

I'm happy for your review and I like the result better too. Thanks for providing your feedback and merging my contribution.

PR #253, opened 2023-06-22
Fixes to spec file to include the right release, source URL, man page, and docs.

sahanaprasad07:
Fixes to spec file to include the right release, source URL, man page and docs.
These errors were discovered by fedora-review
https://bugzilla.redhat.com/show_bug.cgi?id=2211754

### DIFF
--- a/packaging/pkcs11-provider.spec
+++ b/packaging/pkcs11-provider.spec
@@ -1,10 +1,10 @@
 Name:          pkcs11-provider
 Version:       0.1
-Release:       1%{dist}
+Release:       1%{?dist}
 Summary:       A PKCS#11 provider for OpenSSL 3.0+
 License:       Apache-2.0
 URL:           https://github.com/latchset/pkcs11-provider
-Source:        pkcs11-provider-%{version}.tar.gz
+Source0:       %{url}/releases/download/v%{version}/%{name}-%{version}.tar.xz
 
 BuildRequires: openssl-devel >= 3.0.5
 BuildRequires: gcc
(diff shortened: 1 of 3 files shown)


packaging/pkcs11-provider.spec
```diff
@@ -54,8 +55,8 @@ make check || if [ $? -ne 0 ]; then cat tests/*.log; exit 1; fi;
 
 %files
 %license COPYING
-%{_mandir}/man7/*
-%doc README
+%{_mandir}/man7/%{name}*
```
simo5, line 58: Why this change?
The man page will not match this
name = pkcs11-provider
manpage = provider-pkcs11.7

simo5 (changes requested):

packaging/pkcs11-provider.spec
sahanaprasad07: It  was a suggestion on the fedora-review bug that too unspecified wildcard was used.
If we don't want to align the manpage to name, then I guess we can let it be as is.
I have reverted the change.

simo5 (approved):

simo5:
Thanks!

packaging/pkcs11-provider.spec
```diff
@@ -24,6 +24,7 @@ BuildRequires: p11-kit-server
 BuildRequires: gnutls-utils
 BuildRequires: xz
 BuildRequires: expect
+BuildRequires: make
```
Jakuje, line 27: nit: I would move the `make` above the comment `# for tests`

PR #254, opened 2023-06-23
tests: Avoid bogus failures from ossl helper

Jakuje:
Fixes #252

PR #256, opened 2023-06-29
Run Shellcheck as part of CI and squash the shellcheck and yamlint warnings

Jakuje:
Follow-up to #254 where the shell did not do what it was supposed to be doing, I decided to take shellcheck into action and squash the warnings to keep the tests "clean". On the way I also reformatted they github yml files as the editor was going craze every time I had to touch them.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,10 +1,11 @@
+---
 name: Build
 
 on:
   push:
-    branches: [ "main" ]
+    branches: ["main"]
   pull_request:
-    branches: [ "main" ]
+    branches: ["main"]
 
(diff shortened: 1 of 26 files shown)


Jakuje:
> I like this but do not get why the new shellcheck test fails?

The GH action is still WIP. Give me couple of minutes :)

Jakuje:
the shellcheck looks good (hope I did not mess up some functionality with the changes). Homebrew looks broken though.

Jakuje:
> macos is actually failing the uri test now

aaand it works now!

simo5 (approved):
LGTM

PR #261, opened 2023-07-09
Fix mismatching version in spec file

simo5:


Jakuje (approved):
for the use in packit, I think the version is updated automatically somehow, but I do not know details.  If we want to keep it updated manually, it will be repetitive task after each release.

simo5:
> for the use in packit, I think the version is updated automatically somehow, but I do not know details. If we want to keep it updated manually, it will be repetitive task after each release.

No this is just a quick fix, later on I will fix this issue by moving spec to be a spec.in file and generate actual spec via configure.

Issue #266, opened 2023-07-26
Digest + Sign operation no longer working (in default config)


fabled:
The BIND dnssec-signzone no longer is able to sign using RSA keys. The RSA (also EC) Digest + Sign operation no longer works.

Bisected to commit a912c04fc36ee65f6a7d02b57cba9794bf66a982.

```
dnssec-signzone: info: DNSKEY rsasha256.example/RSASHA256/13390 (ZSK) is now active
dnssec-signzone: info: DNSKEY rsasha256.example/RSASHA256/27089 (KSK) is now active
dnssec-signzone: rsasha256.example/NSEC:
dnssec-signzone: 	signing with dnskey rsasha256.example/RSASHA256/13390
dnssec-signzone: warning: EVP_SignFinal (opensslrsa_link.c:269) failed (failure)
dnssec-signzone: info: error:408000E0:pkcs11::The token was not present in its slot when the function was invoked:store.c:114:Failed to get session to load keys
dnssec-signzone: info: error:408000E0:pkcs11::The token was not present in its slot when the function was invoked:store.c:114:Failed to get session to load keys
dnssec-signzone: info: error:408000E0:pkcs11::The token was not present in its slot when the function was invoked:store.c:114:Failed to get session to load keys
dnssec-signzone: info: error:408000E0:pkcs11::The token was not present in its slot when the function was invoked:store.c:114:Failed to get session to load keys
dnssec-signzone: info: error:03000096:digital envelope routines::operation not supported for this keytype:crypto/evp/signature.c:579:
dnssec-signzone: fatal: dnskey 'rsasha256.example/RSASHA256/13390' failed to sign data: failure
```

The main error that happens on every commit afterwards (other errors vary) is:
```
dnssec-signzone: info: error:03000096:digital envelope routines::operation not supported for this keytype:crypto/evp/signature.c:579:
```

As if the EVP structure is not setup properly for signatures.

simo5:
If it does not work with that option can you provide a debug log from pkcs11-provider?

fabled:
Yes, things work with `pkcs11-module-load-behavior = early`. The issue is happening against SoftHSMv2 using OpenSSL backend.

Find attached debug log with default load behavior and failed operation. (Though I have locally modified check out to fix #265
[debug.log](https://github.com/latchset/pkcs11-provider/files/12174032/debug.log)
).

fabled:
[...]

Issue #265, opened 2023-07-26
Getting RSA public components no longer working


fabled:
Getting RSA public components no longer works. Bisected to commit 00c82644b544f9ac2379dd591577eeeb43a12fb9.

Seems that the removal of public object export in `p11prov_store_export_object` triggered this issue. Adding the removed code back to it makes things work again.

Steps to reproduce:
 1. opened with `OSSL_STORE_open` with pkcs11 URL to match a public RSA key
 2. get `EVP_PKEY` using  `OSSL_STORE_INFO_get1_PUBKEY` for the public key
 3. try to get `e` or `n` with `EVP_PKEY_get_bn_param(pub, OSSL_PKEY_PARAM_RSA_E, &e)`

The above used to work before mentioned commit, but no longer works.

Is this intentional or considered a regression? If intentional, is there any workaround or is it now no longer possible to get the public components of public key?

simo5:
Unexpected side effect, but I need to figure out why I removed it in the end.

fabled:
For reference, I'm now running locally with:
``` diff
diff --git a/src/store.c b/src/store.c
index c76b314..21cfe03 100644
--- a/src/store.c
+++ b/src/store.c
@@ -440,9 +440,28 @@ static int p11prov_store_export_object(void *loaderctx, const void *reference,
                                        size_t reference_sz,
                                        OSSL_CALLBACK *cb_fn, void *cb_arg)
 {
+    P11PROV_OBJ *obj = NULL;
+
     P11PROV_debug("store (%p) export object %p, %zu", loaderctx, reference,
                   reference_sz);
 
+    /* the contents of the reference is the address to our object */
+    obj = p11prov_obj_from_reference(reference, reference_sz);
+    if (!obj) {
+        return RET_OSSL_ERR;
+    }
+
+    /* we can only export public bits, so that's all we do */
+    if (p11prov_obj_get_class(obj) != CKO_PUBLIC_KEY) {
+        return RET_OSSL_ERR;
+    }
+
+    switch (p11prov_obj_get_key_type(obj)) {
+    case CKK_RSA:
+        return p11prov_obj_export_public_rsa_key(obj, cb_fn, cb_arg);
+    case CKK_EC:
+        return p11prov_obj_export_public_ec_key(obj, cb_fn, cb_arg);
+    }
     return RET_OSSL_ERR;
 }
 
```

I can do PR out of this, if it looks like a good idea and there's no other unwanted side effects.

[...]

PR #267, opened 2023-07-26
Change how key export works, and add key import functionality as well as key match

simo5:
This are various fixes to small things I had in my tree and to recent issues reported in #265 and #266 

Note that I changed how export work, including an attempt to re-enable direct export from store and then again disable it.

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -3,6 +3,7 @@
 
 #include "provider.h"
 #include <string.h>
+#include <openssl/bn.h>
 #include <openssl/ec.h>
 #include <openssl/x509.h>
 #include <openssl/obj_mac.h>
@@ -11,6 +12,8 @@
 #define CKA_P11PROV_CURVE_NAME CKA_P11PROV_BASE + 1
 #define CKA_P11PROV_CURVE_NID CKA_P11PROV_BASE + 2
 #define CKA_P11PROV_PUB_KEY CKA_P11PROV_BASE + 3
+#define CKA_P11PROV_PUB_KEY_X CKA_P11PROV_BASE + 4
+#define CKA_P11PROV_PUB_KEY_Y CKA_P11PROV_BASE + 5
 
 struct p11prov_key {
     CK_KEY_TYPE type;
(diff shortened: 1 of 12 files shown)


fabled:
Tested and things don't fully work with the latest commit disabling store export. The things that fails without is that when I get matching public key from store against same openssl public key using the `EVP_PKEY_eq` fails.

simo5:
Are these all RSA keys?
I am aware that EC keys are not working yet.

If not can you describe a simple reproducer (or contribute a test that is a simple reproducer) ?

simo5:
I think EVP_PKEY_eq fails because I did not implement the match() function.

It works when store export is available, because when that happens the public key is exported and imported into the default provider instead, and the default provider implements the match function.

I will add implementation of a match function and EVP_PKEY_eq test to match.

simo5:
I also now know why ECDH fails.
It is because allowing to export from store exports the public key to the default provider, then later openssl tries to import the key back into the pkcs11-provider because it needs it to perform the ECDH using the private key part that is not exportable, but importing keys is not yet supported because I did not think of tying to import keys into a session, and importing to the token is problematic as many tokens do not support that or would cause littering he token with ephemeral keys ...

fabled:
> Are these all RSA keys? I am aware that EC keys are not working yet.

Yes, testing first with RSA keys. Then will move on to EC keys too.

> If not can you describe a simple reproducer (or contribute a test that is a simple reproducer) ?

Sure, I can add a small test case for this. Not sure yet when I get the time slot. Possibly this week, or hopefully by latest early next week. If you get to it first, that's also ok.

simo5:
Nice I will steal it into the PR then

simo5:
So I use this test but it does not reproduce your case.
With this code what happens is that OpenSSL imports the public key into the default provider, so the comparison is happening there ...
I can force keys to not be exportable at all, will see if that gives failures with current code.

simo5:
Forcing keys to stay on the token make them fail without a ec_match function in the provider and succeed when ec_match is available.

fabled:
My commit is in top of the last working commit of this PR. Rebasing it on top of the latest commit in the PR should produce the failure.

simo5:
> My commit is in top of the last working commit of this PR. Rebasing it on top of the latest commit in the PR should produce the failure.

I realized that I had tested it with the patch to remove export disabled.
But that turned out to be a good thing so I could test the matching function.

fabled:
Also, it seems that the public key export works different depending if commit 1ca4a1758d402e614667156d8bbd3c0cc6c458c0 is present or not. With the commit present, the equality check fails because it's unable to load the RSA public key from the `.pem` file.

The difference seems to be that with the commit (store export blocked), the `.pem` file contains `-----BEGIN RSA PUBLIC KEY-----                                                  
` block. And without the commit there's a `-----BEGIN PUBLIC KEY-----` block. Also the whole ASN1 structure is different. I suspect this is also a regression?

simo5:
It is a side effect of exporting to the default provider or not. I am working on various things to allow us to decide what to do.
One area is working on allowing openssl to import keys into the pkcs11-provider... but in the end muy intention is still to block store exporting as it has more side effects than not blocking it.

simo5:
I also filed https://github.com/openssl/openssl/issues/21596

simo5:
@fabled the current set of commits makes all tests pass, but that does not mean everything I did is enough for the way bind uses the API.

Can you please recheck RSA (and maybe even EC) with this patchset applied?

fabled:
> Can you please recheck RSA (and maybe even EC) with this patchset applied?

With checkout at commit e763a88ecdb9fcdbae2d52ffe981272e6ca15294 and `pkcs11-module-load-behavior = early` (still required), the first set of BIND RSA and EC tests PASS. Thank you!

I'll work next with the second set of tests. But I would expect those to pass also, as they should not introduce different paths to openssl, just different paths inside bind.

simo5:
@Jakuje can you help me figuring out why address sanitizer blows up on Debian on a single test with Softhsm ?

Jakuje:
hmm. Sounds like an issue in debian side ... we use Sid, which is rolling release and it might have some bugs. Does the Bookworm/`debian:latest` already have OpenSSL 3.x so we could use more stable release?

I just rerun the test locally in container and it worked ok for me.  Can you try to rerun it in github or try locally?

simo5:
@Jakuje rerunning does not help.
I guess I will ignore this test for now, and either disable asan on debian in a separate PR or find out if there is a more stable debian version with OpenSSL 3 we can use.

Jakuje:
Problem is that I can not reproduce the issue locally in container. Sorry ...

> if there is a more stable debian version with OpenSSL 3 we can use.

Bookworm has 3.0: https://packages.debian.org/bookworm/openssl and should be more stable -- I would give it a try in separate PR. Will need a modification to the debuginfo repo though.

src/objects.c
```diff
@@ -1799,3 +1813,748 @@ int p11prov_obj_export_public_ec_key(P11PROV_OBJ *obj, OSSL_CALLBACK *cb_fn,
...
+    val = ec_params->pValue;
+    group = d2i_ECPKParameters(NULL, &val, ec_params->ulValueLen);
+    if (!group) {
+        ret = RET_OSSL_ERR;
+        goto done;
+    }
+
+    x = BN_CTX_get(bnctx);
+    y = BN_CTX_get(bnctx);
+    if (!y) {
```
Jakuje, line 1922: missing null check for `x`.

simo5: From openssl man page directly:
`Once BN_CTX_get() has failed, the subsequent calls will return NULL as well, so it is sufficient to check the return value of the last BN_CTX_get() call.`

tests/tbasic
```diff
@@ -152,6 +151,25 @@ fi
 OPENSSL_CONF=${ORIG_OPENSSL_CONF}
 rm -f "${OPENSSL_CONF}.nopin"
```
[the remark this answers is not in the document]
simo5: Fixed and rebased the relevant commit.

src/objects.c
```diff
@@ -1799,3 +1813,748 @@ int p11prov_obj_export_public_ec_key(P11PROV_OBJ *obj, OSSL_CALLBACK *cb_fn,
...
+        P11PROV_raise(key->ctx, CKR_KEY_INDIGESTIBLE,
+                      "Failure to encode EC point to DER");
+        return CKR_KEY_INDIGESTIBLE;
+    }
+    key->attrs[key->numattrs].type = CKA_EC_POINT;
+    key->attrs[key->numattrs].pValue = der;
+    key->attrs[key->numattrs].ulValueLen = len;
+    key->numattrs++;
+
+    /* TODO: EC_PARAMS */
```
Jakuje, line 2438: is this todo still valid? If so, I think it needs some more description as it is not clear to me what needs to be done and why.

src/objects.c
```diff
@@ -1799,3 +1813,748 @@ int p11prov_obj_export_public_ec_key(P11PROV_OBJ *obj, OSSL_CALLBACK *cb_fn,
...
+    }
+
+    if (allocattrs < findctx.numattrs) {
+        allocattrs = findctx.numattrs;
+    }
+
+    /* A common case with openssl is the request to import a key we already
+     * actually have on the token. This happens because OpenSSL is greedy
+     * and tries to export keys to its default provider before it even knows
+     * wat kind of operation it needs to do. Sometimes the operation ends up
```
Jakuje, line 2508:
```suggestion
     * what kind of operation it needs to do. Sometimes the operation ends up
```

src/objects.c
```diff
@@ -1799,3 +1813,748 @@ int p11prov_obj_export_public_ec_key(P11PROV_OBJ *obj, OSSL_CALLBACK *cb_fn,
...
+        rv = return_dup_key(key, findctx.found);
+        goto done;
+    }
+
+    /*
+     * FIXME:
+     * For things like ECDH we can get away with a mock objects that just holds
+     * data for now, but is not backed by an actual handle and key in the token.
+     * Once this is not sufficient, we'll probably need to change functions to
+     * pass in a valid session when requesting a handle from an keyect, so that
```
Jakuje, line 2528: `keyect`?

tests/tbasic
```diff
@@ -150,7 +149,25 @@ if [ $FAIL -ne 0 ]; then
...
+title PARA "Test EVP_PKEY_eq on public EC key via import"
+$CHECKER ./tcmpkeys "$ECPUBURI" "${TMPPDIR}"/ec.pub.uripin.pem
+
+title PARA "Test EVP_PKEY_eq with key exporting disabled"
+ORIG_OPENSSL_CONF=${OPENSSL_CONF}
+OPENSSL_CONF=${OPENSSL_CONF}.noexport
+title PARA "Test RSA key"
+$CHECKER ./tcmpkeys "$PUBURI" "$PUBURI"
+title PARA "Test EC key"
+$CHECKER ./tcmpkeys "$ECPUBURI" "$ECPUBURI"
```
Jakuje, line 169: We should test also comparison of the EdDSA keys if possible

tests/tgenkey.c
```diff
@@ -67,6 +68,49 @@ static void check_keys(OSSL_STORE_CTX *store)
         exit(EXIT_FAILURE);
     }
 
+    /* check we can get pub params from key */
```
Jakuje, line 71: Same here, no coverage for the eddsa

src/objects.c
```diff
@@ -1799,3 +1813,748 @@ int p11prov_obj_export_public_ec_key(P11PROV_OBJ *obj, OSSL_CALLBACK *cb_fn,
...
+        return rv;
+        ;
+    }
+    dst->numattrs = 0;
+    for (int i = 0; i < src->numattrs; i++) {
+        rv = p11prov_copy_attr(&dst->attrs[i], &src->attrs[i]);
+        if (rv != CKR_OK) {
+            rv = CKR_HOST_MEMORY;
+            P11PROV_raise(dst->ctx, rv, "Failed attr copy");
+            return rv;
```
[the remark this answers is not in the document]
simo5: no actually it is not needed, this is stored on the key object which will be freed if the whole thing fails, that is why I am careful to update dst->numattrs after each step, so the p11prov_obj_free() will deal with this correctly.

src/objects.c
simo5: lol I did a s/obj/key/g at some point ... good catch

tests/tbasic
simo5: maybe open an issue for that?

tests/tgenkey.c
simo5: yeah I think we should have a separate PR for that, this one grew enough

Jakuje (approved):
no other technical issue noticed so giving my ack

simo5:
Thanks a lot @Jakuje
Given the only issues is this weird address sanitezer misbehavior, I am going to merge this now, and create new PRs for followups

tests/tgenkey.c
simo5: Great, thank you

Issue #270, opened 2023-07-31
Do not login the session if already authenticated outside the provider


manison:
Actually I don't know whether this is a bug or a feature request :) However the libp11 pkcs11 engine once behaved this way and we (ab)used the behavior in our application.

**Describe the feature**
When the provider opens a session for the token for the first time (I believe it's called the login session in the terms of this project) it might be already authenticated if there is another client in the process that uses the PKCS#11 module directly and bypasses the pkcs11-provider. In that case the `C_Login` call that the provider makes is unnecessary.

**Expected behavior**
When the pkcs11-provider opens a new session, it first checks its login state and skips `C_Login` if the session is already authenticated.

simo5:
If a login session is already active, we just get an "error" that says the token is already logged in. Why do you need to disable this attempt?

manison:
Because even this attempt requires user interaction - in our case a user have to enter the PIN using pinpad, so we try to limit the number of interactions to minimum, ideally the user enters the PIN once at the start of the application.

simo5:
It was created for cases where you do not want to log in ... the problem is that I am not sure there is a way to find out if the token is logged in without trying to do a login operation.

simo5:
Hadn't thought about that, I will look into it.

simo5:
I consider this issue completed since #273 was merged.

Issue #271, opened 2023-08-03
Concurrent signing fails intermittently (regression)


fabled:
**Describe the bug**
Signing concurrently from multiple threads using same token has intermittent failures.

**To Reproduce**
 1. Get a private key from token
 2. Start N threads
 3. In each thread sign data using key from step 1
 4. Depending on timing of things, a signing operation can fail intermittently with:
```
dnssec-signzone: info: error:4080000A:pkcs11::Internal locking failure:session.c:621:Failed to fetch a login_session
dnssec-signzone: info: error:4080000A:pkcs11::Internal locking failure:signature.c:837:Failed to open session on slot 131283951
```

**Expected behavior**
No errors during signing.

**Additional context**
It seems that since commit https://github.com/latchset/pkcs11-provider/commit/6403f22383f8c00860eb422a75c17a6968fb3a13 fetch_session will immediately fail if two threads are concurrently doing the first operation on a slot that would need login.

It seems that setting `reqlogin` in https://github.com/latchset/pkcs11-provider/blob/main/src/signature.c#L807 triggers `p11prov_get_session` to call `slot_login` which has this race.

Additionally this will cause all threads to serialize to have access for the login session even if already logged in. Since all session share login state, it might make sense to first get a session from the pool, and check the logged in state from it. Only if not-logged in, fall back to the slow path of acquiring the login session.

simo5:
Sounds like this goes hand in hand with #270

fabled:
> Sounds like this goes hand in hand with #270

In my case, pkcs11-provider still needs to do login on first use. This is more about how concurrent actions are synchronized. But I suppose there is some overlap with these two tickets.

simo5:
I f we have a check for session login existing I think we could relax the serialization on login session ... need to think about it

simo5:
@fabled, is there something left here todo or does #273 cover this fully?

PR #273, opened 2023-08-04
Smarten up handling of login session

simo5:
Although we want to ensure a login session is available for certain operations, we also do not want to unnecessarily fail them when mutliple operations are running concurrently.
As login status is token global, all we need to ensure is that a login session exist. This can be reasonably easily determined by session status as well.
So where we do not actually need to use the login_session directly we can simply check the status and return positively.

The only place that really wants to use a login_session directly is the caching layer which is tolerant to failure already.

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -12,6 +12,7 @@ struct p11prov_session {
 
     CK_SLOT_ID slotid;
     CK_SESSION_HANDLE session;
+    CK_STATE state;
     CK_FLAGS flags;
 
     pthread_mutex_t lock;
@@ -61,6 +62,7 @@ static CK_RV token_session_callback(CK_SESSION_HANDLE hSession,
 #define SLEEP 50000
 static CK_RV token_session_open(P11PROV_SESSION *session, CK_FLAGS flags)
 {
+    CK_SESSION_INFO session_info;
     uint64_t startime = 0;
     CK_RV ret;
 
@@ -78,11 +80,20 @@ static CK_RV token_session_open(P11PROV_SESSION *session, CK_FLAGS flags)
     if (ret != CKR_OK) {
         session->session = CK_INVALID_HANDLE;
         session->flags = DEFLT_SESSION_FLAGS;
+        session->state = CK_UNAVAILABLE_INFORMATION;
         return ret;
     }
 
     session->flags = flags;
-    return CKR_OK;
+
+    /* get current state */
+    ret = p11prov_GetSessionInfo(session->provctx, session->session,
+                                 &session_info);
+    if (ret == CKR_OK) {
+        session->flags = session_info.flags;
+        session->state = session_info.state;
+    }
+    return ret;
 }
 
 static void token_session_close(P11PROV_SESSION *session)
@@ -93,6 +104,7 @@ static void token_session_close(P11PROV_SESSION *session)
         /* regardless of the result the session is gone */
         session->session = CK_INVALID_HANDLE;
         session->flags = DEFLT_SESSION_FLAGS;
+        session->state = CK_UNAVAILABLE_INFORMATION;
     }
 }
 
@@ -183,6 +195,7 @@ void p11prov_session_pool_fork_reset(P11PROV_SESSION_POOL *pool)
 
             session->session = CK_INVALID_HANDLE;
             session->flags = DEFLT_SESSION_FLAGS;
+            session->state = CK_UNAVAILABLE_INFORMATION;
             session->in_use = false;
             session->cb = NULL;
             session->cbarg = NULL;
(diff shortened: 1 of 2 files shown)


simo5:
@manison can you check if this addresses #270 ?
@fabled can you check if this addresses #271 ?

src/session.c
```diff
@@ -526,7 +550,11 @@ static CK_RV fetch_session(P11PROV_SESSION_POOL *pool, CK_FLAGS flags,
         ret = MUTEX_LOCK(pool->login_session);
         if (ret == CKR_OK) {
             if (pool->login_session->in_use) {
-                ret = CKR_CANT_LOCK;
+                if (is_login_state(pool->login_session->state)) {
```
fabled, line 553: This has a race condition: if the other thread is still executing login, but has not completed it, this will immediately return `CKR_CANT_LOCK`. The time size of the window depends on how slow/fast the token is to complete a login, and how fast the parallel login attempt happens. This also likely happens only if two threads initiate a key search operation requiring login at exact same time.

src/session.c
```diff
@@ -611,21 +639,24 @@ static CK_RV slot_login(P11PROV_SLOT *slot, P11PROV_URI *uri,
     P11PROV_SESSION_POOL *pool = p11prov_slot_get_session_pool(slot);
     P11PROV_SESSION *session = NULL;
     CK_FLAGS flags = DEFLT_SESSION_FLAGS;
-    CK_STATE state = CKS_RO_PUBLIC_SESSION;
     int num_open_sessions = 0;
     CK_RV ret;
 
     /* try to get a login_session */
     ret = fetch_session(pool, flags, true, &session);
+    if (ret == CKR_USER_ALREADY_LOGGED_IN && _session == NULL) {
```
fabled, line 647: This is now subtly behaving differently depending whether there's concurrent operations happening or not. The thread that gets the login session, will end up doing `session_check` other things, where as the contending thread losing access will not.

Perhaps this is intentional design as it will then make sure `session_check` is called periodically / often enough, but ensure race losing threads work too.

fabled (commented):
Significantly improves things! Thank you! I think it also fixes fully the use case I have, because the first operation done is a single threaded call to find the token object. Also crypto operations should just work now, as the login session should be established when the private key is searched. Minor race still exists if multiple threads concurrently initiate an operation that needs a login.

src/session.c
simo5: Yeah, that is why I still can return a CANT_LOCK error ...

src/session.c
simo5: So here is the rub of the problem.
The login operation can take arbitrarily long because you may be in a situation where a callback to the user is request to insert the PIN manually.

So the question is, what to do ?

Should we block all threads until this happens?
Perhaps we should ... I think I was worried about possible deadlocks in the past.

Maybe have a large timeout via backoff loop? So that eventually (say in 5 minutes) threads will unblock (and fail) ?

fabled: So there's two cases:
- PIN is part of the URL or config, and the race is purely the time of what it takes to execute a login; usually less than few seconds
- PIN is not present, and user interaction is required; getting input can take arbitrary time

I think as along as the PIN provided case works, things should be good. The timeout should likely be a sane default (possibly configurable) on how much it takes time to provide PIN. I would probably think a minute should be good enough.

And I'd also make it so that the timeout count starts from the first operation initiated, and if concurrent threads join, they timeout when the first operation timeouts.

src/session.c
fabled: A small nit regarding using `CKR_CANT_LOCK`: it is against the specification's purpose for this error:
>  #define CKR_CANT_LOCK
> This value can only be returned by '''C_Initialize'''. It means that the type of locking requested by the application for thread-safety is not available in this library, and so the application cannot make use of this library in the specified fashion. 

A better option would perhaps be:
> #define CKR_OPERATION_ACTIVE
> There is already an active operation (or combination of active operations) which prevents Cryptoki from activating the specified operation.

Or just specify our own from the `CKR_VENDOR_DEFINED` range.

manison:
With a small addition this also seems to fix #270:

When a session is open it might have been already authenticated outside the provider. The following patch checks for this situation.

```diff
         ret = token_session_open(session, flags);
         if (ret == CKR_OK) {
             num_open_sessions++;
+            if (is_login_state(session->state)) {
+                /* we seem to have a valid session authenticated outside this provider */
+                goto done;
+            }
         } else {
             goto done;
         }
```

simo5:
> With a small addition this also seems to fix #270:
> 
> When a session is open it might have been already authenticated outside the provider. The following patch checks for this situation.
> 
> ```diff
>          ret = token_session_open(session, flags);
>          if (ret == CKR_OK) {
>              num_open_sessions++;
> +            if (is_login_state(session->state)) {
> +                /* we seem to have a valid session authenticated outside this provider */
> +                goto done;
> +            }
>          } else {
>              goto done;
>          }
> ```

Actually what we need is to just move the existing is_login_session() check to just before we call token_login(), but good catch.

simo5:
@fabled can you recheck this PR?
I have slightly adjusted where login session is checked as per reply to @mansion's comment.

simo5:
Given this PR is about smarter logins, I added a minor tweak to the store function that determines if login is necessary based on the URI filter.
If we are requesting public keys only we do not login by default.
Seem to be working fine on my end.

PR #275, opened 2023-08-09
Extend key comparison tests

manison:
This PR adds a few additional key comparison tests. Mainly it demonstrates that implementation of matching keys in the pkcs11-provider is currently broken for EC when matching private key to public key. This functionality is required by some higher OpenSSL functions like `CMS_add1_signer` where signer's certificate is being matched to signer's private key.

This PR also includes explicit EC in the tests.


manison:
I updated the PR to synthetize `CKA_EC_PARAMS` for imported keys. I'm not sure about various `MAX_xxx` constants that define maximum number of cached attributes, this will eventually need thorough review. Also I put the code directly in `prep_ec_find` instead of the `fix_ec_key_import` as was suggested by the _TODO_ comment. This is due to the fact that there might be no information to synthetize the EC params from (curve nid/name may be undefined for explicit EC). Also in `prep_ec_find` the `EC_GROUP` is already conveniently available.

src/objects.c
```diff
@@ -2156,18 +2212,21 @@ int p11prov_obj_key_cmp(P11PROV_OBJ *key1, P11PROV_OBJ *key2, CK_KEY_TYPE type,
...
-        /* if nothing fails it is a match */
-        return RET_OSSL_OK;
+        break;
 
     default:
         return RET_OSSL_ERR;
     }
+
+    if (cmp_type & OBJ_CMP_KEY_PUBLIC) {
+        ret = match_public_keys(key1, key2);
```
simo5, line 2222: you can `return match_public_keys(key1, key2);` here.

simo5 (approved):
LGTM

simo5:
If you can fix the CI failures and sort out the commits this is good to be merged.

manison:
Polished the commits and suppressed false positives from shellcheck.

PR #277, opened 2023-08-21
Allow setting OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY for EC keys

manison:
This PR implements `set_params` method for EC keys that allows assigning `OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY` parameter. This parameter is required for `EVP_PKEY_set1_encoded_public_key` to function properly. This function is in turn used by `CMS_decrypt_set1_pkey_and_peer` so without this decrypting a CMS does not work at all.

This PR also adds a test to check CMS encrypt/decrypt using an EC key. Since decrypting a CMS uses `CKM_ECDH1_DERIVE` mechanism with shared data the test is being run on nss-softokn only. Although SoftHSM2 also supports ECDH1 derivation mechanism it currently (https://github.com/opendnssec/SoftHSMv2/pull/599) only does when no shared data is specified.

### DIFF
--- a/src/keymgmt.c
+++ b/src/keymgmt.c
@@ -1003,6 +1003,8 @@ DISPATCH_KEYMGMT_FN(ec, export_types);
 DISPATCH_KEYMGMT_FN(ec, query_operation_name);
 DISPATCH_KEYMGMT_FN(ec, get_params);
 DISPATCH_KEYMGMT_FN(ec, gettable_params);
+DISPATCH_KEYMGMT_FN(ec, set_params);
+DISPATCH_KEYMGMT_FN(ec, settable_params);
 
 static void *p11prov_ec_new(void *provctx)
 {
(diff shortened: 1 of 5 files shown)


manison:
It's during `CMS_Decrypt`, it should be easily visible from the _tcms_ test I added in this PR.

This is the backtrace:

```
pkcs11.dll!p11prov_obj_set_ec_encoded_public_key(p11prov_obj * key, const void * pubkey, unsigned __int64 pubkey_len)
pkcs11.dll!p11prov_ec_set_params(void * keydata, const ossl_param_st * params)
[Inline Frame] libcrypto3.dll!EVP_PKEY_set_params(evp_pkey_st *)
[Inline Frame] libcrypto3.dll!EVP_PKEY_set_octet_string_param(evp_pkey_st *)
libcrypto3.dll!EVP_PKEY_set1_encoded_public_key(evp_pkey_st * pkey, const unsigned char * pub, unsigned __int64 publen)
libcrypto3.dll!ecdh_cms_set_peerkey(evp_pkey_ctx_st * pubkey, X509_algor_st *)
libcrypto3.dll!ecdh_cms_decrypt(CMS_RecipientInfo_st * ri)
libcrypto3.dll!CMS_RecipientInfo_kari_decrypt(CMS_ContentInfo_st * cms, CMS_RecipientInfo_st * ri, CMS_RecipientEncryptedKey_st * rek)
[Inline Frame] libcrypto3.dll!cms_kari_set1_pkey_and_peer(CMS_ContentInfo_st *)
libcrypto3.dll!CMS_decrypt_set1_pkey_and_peer(CMS_ContentInfo_st * cms, evp_pkey_st * pk, x509_st * cert, x509_st * peer)
[Inline Frame] libcrypto3.dll!CMS_decrypt_set1_pkey(CMS_ContentInfo_st *)
libcrypto3.dll!CMS_decrypt(CMS_ContentInfo_st * cms, evp_pkey_st * pk, x509_st * cert, bio_st * dcont, bio_st * out, unsigned int flags)
```

The [`ecdh_cms_set_peerkey`](https://github.com/openssl/openssl/blob/7c729851d169f30d9e0c0ad6e7c1cf6cefb37935/crypto/cms/cms_ec.c#L73) calls `EVP_PKEY_copy_parameters` so my guess is that they duplicate a key and then replace its public part?

simo5:
Ok, I see how this is messed up, but it is what it is.
What I need here before we can merge is a check that will refuse to set parameters if this is a key from the token. Parameter setting should be available only for generated keys.
We may need to add a boolean to the key structure to record this fact.

manison:
> What I need here before we can merge is a check that will refuse to set parameters if this is a key from the token. Parameter setting should be available only for generated keys.

Agreed.

> We may need to add a boolean to the key structure to record this fact.

Wouldn't it be enough to check the object handle and/or slot id?

simo5:
> Wouldn't it be enough to check the object handle and/or slot id?

I guess that will work, yeah.

simo5:
Looks good

simo5:
Sadly we have some CI woes lately, but this looks good, so I'll merge and fix CI later.

PR #283, opened 2023-09-05
Fix key generation template size and debug log message

fabled:
- Fix EC private key generation template size
- Fix p11prov_common_gen_init debug log entry


simo5:
Looks good to me, but DCO check needs to be fixed.

simo5 (changes requested):
Fix DCO check please

simo5 (approved):
LGTM

PR #286, opened 2023-09-06
Reinit module only once after fork

### LINKED ISSUE #281: Reinitialization after fork should be thread safe
**Describe the bug**
1. Application initializes openssl and pkcs11-provider.
2. Application forks to background (daemonizes). The slots are set to `P11PROV_NEEDS_REINIT` state.
3. Application starts multiple threads to do crypto.
4. `p11prov_ctx_status` gets called in parallel in multiple threads and ends up executing https://github.com/latchset/pkcs11-provider/blob/main/src/provider.c#L488-L496 in multiple threads.
5. All threads serialize in `p11prov_module_reinit` at  https://github.com/latchset/pkcs11-provider/blob/main/src/interface.c#L377
6. Thread 1 - will reinitialize the slot. And start doing crypto.
7. Thread 2 - will reinitialize again the slot and calling C_Finalize while thread 1 doing crypto causes crash. pkcs11-provider call C_Finalize while other thread is performing crypto is violation of the pkcs11 spec.

**Expected behavior**
No crash. The reinit code should be thread safe.

Main issue is that https://github.com/latchset/pkcs11-provider/blob/main/src/provider.c#L488-L496 is not thread safe.

simo5:
This prevent multiple threads from stomping on each other and reinit the module multiple times.

Fixes #281

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -23,6 +23,7 @@ struct p11prov_module_ctx {
 
     pthread_mutex_t lock;
     bool initialized;
+    bool reinit;
 };
 
 /* This structure is effectively equivalent to CK_FUNCTION_LIST_3_0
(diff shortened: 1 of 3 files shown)


simo5:
@fabled can you check if this fixes #281 properly ?

fabled (approved):
:+1: The test case I had with bind `named` now seems to work. Thank you!

PR #288, opened 2023-09-06
Plug a small race in token login operations

### LINKED ISSUE #287: Small race condition in token login code
This has a race condition: if the other thread is still executing login, but has not completed it, this will immediately return `CKR_CANT_LOCK`. The time size of the window depends on how slow/fast the token is to complete a login, and how fast the parallel login attempt happens. This also likely happens only if two threads initiate a key search operation requiring login at exact same time.

_Originally posted by @fabled in https://github.com/latchset/pkcs11-provider/pull/273#discussion_r1284580669_

simo5:
Some code really need to obatin a session from a login. Some other code just need to attempt but can fail. If a login session is in use cycle again until we have a login session for sure if the calling code really requires a logged in token.

Fixes #287

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -364,7 +364,7 @@ static void cache_key(P11PROV_OBJ *obj)
     }
 
     ret = p11prov_take_login_session(obj->ctx, obj->slotid, &session);
-    if (ret != CKR_OK) {
+    if (ret != CKR_OK || session == NULL) {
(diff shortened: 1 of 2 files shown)


simo5:
@fabled can you check if this is enough to addresses your concern about races for obtaining a login session ?

fabled (approved):
LGTM. Should cover the race.

Issue #282, opened 2023-09-04
Newly generated EC PKEY does not support getting X/Y coordinates 


fabled:
**Describe the bug**
1. Generate new EC key on token
2. Using the PKEY directly from key generation, try to extract X/Y coordinate. This fails.

**To Reproduce**
Extended test suite to reproduce this error:
https://github.com/fabled/pkcs11-provider/commit/42a459cdc550e7b4f11e07b7a88ac01d7ce9a585

**Expected behavior**
Able to extract public X/Y coordinate from the newly generated key.

**Extra context**
If the same key is later retrieved from store, the X/Y coordinate extraction works just fine.

dengert:
Depends on the token and the pkcs11 module. PKCS11 C_GenerateKeyPair returns CK_OBJECT_HANDLE_PTR phPublicKey and CK_OBJECT_HANDLE_PTR phPrivateKey. "The key objects created by a successful call to C_GenerateKeyPair will have their CKA_LOCAL attributes set to CK_TRUE. In addition, the key objects created will both have values for CKA_UNIQUE_ID generated and assigned."  The PKCS11 module should be able to access this  LOCAL public key. 

The public key may or may not be directly accessible on the token.  PIV for example returns the public key when the key is generated. The only public keys on the PIV card are in certificates. Thus public key must be stored long enough to be copied somewhere. A CSR can be created which contains the public key and is signed by the private key. A CA will then create the certificate which is the written to the  to card.  A PIV card driver uses the certificates to find type, size and parameters of the private keys on the card.

fabled:
Exactly. To clarify. This is `pkcs11-provider` issue. The `EVP_PKEY`returned from `EVP_PKEY_generate` fails to allow getting these coordinates. This PKEY is a keypair which should have both public and private key present. But currently it is not.

To quote https://www.openssl.org/docs/man3.0/man3/EVP_PKEY_generate.html in NOTES:
> In OpenSSL an EVP_PKEY structure containing a private key also contains the public key components and parameters (if any). An OpenSSL private key is equivalent to what some libraries call a "key pair". A private key can be used in functions which require the use of a public key or parameters.

However, when this newly generated public key is searched via the store API and that `EVP_PKEY` of the public key is used, the components are gettable.

This issue is about `pkcs11-provider` returning a `EVP_PKEY` object from generate, which does not work as expected.

This of course assumes that the PKCS11 module supports these. See the referenced commit to improve the test case to see code example of these.

[...]

PR #292, opened 2023-09-20
implement more tests with EdDSA keys (export and comparison)

### LINKED ISSUE #272: Test coverage for key comparison and getting public key data from EdDSA
The PR #267 added key comparison test `tcmpkeys` which runs for RSA and EC keys, but there is no test coverage for EdDSA keys. The implementation should be complete though.

It also added reading of public attributes from the generated keys in `tgenkey`, but also without EdDSA coverage, which needs to be added somehow separately (as the EdDSA is available only in SoftHSM for now).

BTW when is the NSS getting EdDSA support? There were some patches, werent they?

Jakuje:
Fixes #272.

Now, it fails though. If I read it right, the OpenSSL returns `OSSL_STORE_INFO_PARAMS` type from the STORE API when I read the public PEM file exported from the provider, while I do not see anything missing in there so I will just drop it here and hope to get back to investigate the issue later.

### DIFF
--- a/src/decoder.c
+++ b/src/decoder.c
@@ -169,7 +169,7 @@ static int p11prov_pem_decoder_p11prov_der_decode(
         return RET_OSSL_CARRY_ON_DECODING;
     }
 
-    P11PROV_debug("PEM_read_pio (fpos:%u)", BIO_tell(bin));
+    P11PROV_debug("PEM_read_bio (fpos:%u)", BIO_tell(bin));
 
     if (PEM_read_bio(bin, &pem_label, &pem_header, &der_data, &der_len) > 0
         && strcmp(pem_label, P11PROV_PEM_LABEL) == 0) {
--- a/src/encoder.c
+++ b/src/encoder.c
@@ -958,3 +958,75 @@ p11prov_common_encoder_priv_key_info_pem_does_selection(void *inctx,
     }
     return RET_OSSL_ERR;
 }
+
+DISPATCH_TEXT_ENCODER_FN(ec_edwards, encode);
+
+static int p11prov_ec_edwards_encoder_encode_text(
+    void *inctx, OSSL_CORE_BIO *cbio, const void *inkey,
+    const OSSL_PARAM key_abstract[], int selection,
+    OSSL_PASSPHRASE_CALLBACK *cb, void *cbarg)
+{
+    struct p11prov_encoder_ctx *ctx = (struct p11prov_encoder_ctx *)inctx;
+    P11PROV_OBJ *key = (P11PROV_OBJ *)inkey;
+    CK_KEY_TYPE type;
+    CK_ULONG keysize;
+    const char *type_name = "ED25519";
+    char *uri = NULL;
+    BIO *out;
+    int ret;
+
+    P11PROV_debug("EdDSA Text Encoder");
+
+    type = p11prov_obj_get_key_type(key);
+    if (type != CKK_EC_EDWARDS) {
+        P11PROV_raise(ctx->provctx, CKR_GENERAL_ERROR, "Invalid Key Type");
+        return RET_OSSL_ERR;
+    }
+
+    out = BIO_new_from_core_bio(p11prov_ctx_get_libctx(ctx->provctx), cbio);
+    if (!out) {
+        P11PROV_raise(ctx->provctx, CKR_GENERAL_ERROR, "Failed to init BIO");
+        return RET_OSSL_ERR;
+    }
+
+    keysize = p11prov_obj_get_key_bit_size(key);
+    if (keysize == 448) {
+        type_name = "ED448";
+    }
+    if (selection & OSSL_KEYMGMT_SELECT_PRIVATE_KEY) {
+        CK_OBJECT_CLASS class = p11prov_obj_get_class(key);
+        if (class != CKO_PRIVATE_KEY) {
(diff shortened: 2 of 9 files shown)


simo5:
Perhaps this needs #282 fixed first?

sahanaprasad07:
I know what to do for #282 . It will be fixed in the next few days.

Jakuje:
> Perhaps this needs #282 fixed first?

yeah, sounds like the EdDSA keys have the same issue when used the code from your [genkey_test_more](https://github.com/latchset/pkcs11-provider/compare/main...simo5:pkcs11-provider:genkey_test_more?expand=1) branch.

But the Ed keys will likely have some other issue too.

> I know what to do for 282 . It will be fixed in the next few days.

Thanks!

Jakuje:
The #282 partially resolves the issue, but it looks like we still need to fix the encoders to work correctly with Edwards curves.

Jakuje:
Resolved conflict now but the basic test still fails to import the exported Ed25519 key (also the newly generated ed keys by openssl only). The encoders implemented in the providers are now only for the RSA and ECDSA, but implementing one in the provided did not help (but I am not aware of any gotchas that would be needed for the EdDSA keys so the default ones should export them correctly).

Filled openssl/openssl#22246

simo5:
@Jakuje I have not looked at this in deep but why do you need a separate function (p11prov_obj_get_ed_pub_key) for Edwards curves?
So far we handled classic EC curves and Edward curves with the same code, so what's missing in p11prov_obj_get_ec_public_x_y() that we'd have to do?

Jakuje:
> @Jakuje I have not looked at this in deep but why do you need a separate function (p11prov_obj_get_ed_pub_key) for Edwards curves? So far we handled classic EC curves and Edward curves with the same code, so what's missing in p11prov_obj_get_ec_public_x_y() that we'd have to do?

I though that overloading the ED (which has just one value on outside) with X and Y separate arguments would be more confusing than not. The Ed keys are much simpler than the whole bunch of parameters and separate X/Y values as in the generic ECDSA.

simo5:
So looking at OpenSSL what we need to return is indeed just:
OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY
and
OSSL_PKEY_PARAM_PUB_KEY

simo5:
But I do not see OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY handled in your patch, only OSSL_PKEY_PARAM_PUB_KEY ...

simo5:
Ah I see, I confused myself.
OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY is only used for ECX in OpenSSL which is ECDH keys ...

simo5:
OTOH neither ed25519_get_params() nor  ed448_get_params() seem to return the public key at all in OpenSSL ... they seem to return only ED25519/448_SECURITY_BITS.

simo5:
The Key Export function in contrast does export OSSL_PKEY_PARAM_PUB_KEY, OSSL_PKEY_PARAM_PRIV_KEY, and only those params.

Jakuje:
Rebased the PR + implemented the ed encoder for text as suggested offline, but I do not think implementing other encoders for various formats will solve the issue that the openssl is unable to read PEM file with the key when the pkcs11 provider is loaded (described in openssl/openssl#22246).

simo5:
The direction looks good, but there is still a softhsm test failure?

Jakuje:
Its still basically what is described in openssl/openssl#22246 -- when pkcs11 provider is loaded, openssl can not read PEM files with EDDSA keys. I tried to dig through that with gdb, but I assume it just skips some macros doing stuff with decoders so I did not get far enough.

I tried to rebuild rawhide openssl with traces to get some more information from the runs, and while updating to `openssl-3.2.1-4.fc41.x86_64` version, it looks like it works just ok. It fails with the current `openssl-3.1.1-4.fc39` in F39 so I believe it is some issue in OpenSSL fixed somewhere between these two versions. Fedora 40 should fix this as far as I know. I pushed a commit to build against fedora:rawhide (normal build, not the dist checks) and it works.

So what now? We can claim the OpenSSL 3.1 version broken for this functionality and focus on 3.2 branch skipping these tests on older openssl and document this (this might have some consequences for RHEL9) or we can try to hunt down the issue in OpenSSL and get it backported. @beldmit any guess what could be the change fixing the eddsa keys handling when there are provider handlers?

beldmit:
No idea from the top of my head

beldmit:
Probably this https://github.com/openssl/openssl/pull/19705 
Or, more possible, https://github.com/openssl/openssl/pull/21219

simo5:
We can definitely assert that certain stuff works only with a later openssl version, unless we have an easy workaround to apply to make it work with 3.0 and 3.1

It would be nice to have a fix in those distributions that ship with a previous version though.

Jakuje:
Lets wait for resolution of openssl/openssl#24038. Regardless of it, after Fedora 40 GA (in following weeks) we should be able to merge this. Debian and Mac OS likely picked up 3.2 ...

Jakuje:
Rebased. I think the typos still make sense to merge as well as ed text export and minor missing tests for completeness.

simo5 (dismissed):
LGTM

simo5:
The additional patch fixes the issue covscan found out, merging.

PR #293, opened 2023-09-29
Embed the private key with public key attributes.

### LINKED ISSUE #282: Newly generated EC PKEY does not support getting X/Y coordinates

sahanaprasad07:
After generating a key pair via p11prov_GenerateKeyPair()  in p11prov_common_gen(), embed the private key with public key attributes through this new function p11prov_merge_pub_attrs_into_priv().
This would be useful to reconstruct the public key from the private key.
Further, attributes like  CKA_P11PROV_PUB_KEY_X and CKA_P11PROV_PUB_KEY_Y can be extracted directly from EC private keys.

The following attributes are copied-
For RSA: CKA_MODULUS, CKA_PUBLIC_EXPONENT
For EC: CKA_EC_POINT, CKA_EC_PARAMS, CKA_P11PROV_PUB_KEY

Fixes #282


src/keymgmt.c
```diff
@@ -526,14 +527,26 @@ static void *p11prov_common_gen(struct key_generator *ctx,
...
+
+    ret = p11prov_merge_pub_attrs_into_priv(pub_key, priv_key);
     if (ret != CKR_OK) {
         p11prov_return_session(session);
         return NULL;
     }
 
     p11prov_return_session(session);
-    return key;
+    return priv_key;
```
simo5, line 549: You need to free pub_key here or it leaks.

simo5 (changes requested):
One issue, but otherwise LGTM

simo5 (approved):

simo5:
LGTM and tests pass!

PR #298, opened 2023-10-23
Implement support for CKA_ALWAYS_AUTHENTICATE

### LINKED ISSUE #42: Implement support for keys with CKA_ALWAYS_AUTHENTICATE=CK_TRUE

simo5:
Fixes #42

Unfortunately I do not have a way to add a test for this in CI just yet, so some manual testing will be required before merging.

- [ ] Manual test performed
- [ ] Docs updated


Jakuje:
I think in softhsm, this could be automated as when you create the key in softhsm you can provide the private key template and OpenSC's pkcs11-tool has `--always-auth` switch, which should create keys with this attribute. Not sure if it is enforced though.

Jakuje:
Sure, I can give it a try.

Jakuje (commented):
Added couple of commits implementing a test (unfortunately, there is no simple way to check the context specific login happens except for reading through the logs now) for this and fixing some issues noticed on the way in:
https://github.com/Jakuje/pkcs11-provider/commits/always_auth

Feel free to pull the changes if you consider them useful.

PR #304, opened 2023-11-06
Switch build system to Meson

ueno:
This adds a basic support for building and installing the provider
with Meson.  To test:

```console
meson setup build
meson compile -C build
meson test -C build --num-processes 1
```

Note that parallel execution of tests are currently not working, as
the same port number is used in multiple tests.

Here is some benchmark:
```console
$ time sh -c 'autoreconf -fi && ./configure && make && make check'
18.18s user 2.97s system 73% cpu 28.639 total

$ time sh -c 'meson setup build && meson compile -C build && meson test -C build --num-processes 1'
10.61s user 1.82s system 81% cpu 15.173 total
```


simo5:
ok so I think a switch to meson is something we can consider, but it would have to be a full switch, definitely not going to maintain the same stuff in parallel in two places

FYI ignore debian, macos test failures for now they got the broken NSS 3.94;
but they should only fail tests and nothing else of course

Anyway for a change like this we'd have to wait until debian and macos upgrade to 3.95 in a month because we definitely need to ensure all the platforms work end-to-end

ueno:
@simo5 I think this is almost ready, though I would need to take a closer look at the current CI failures (macOS build, address sanitizer with Debian, scan-build, and check style).

After the xz backdoor, it seems to be more relevant to switch to a simpler build system that doesn't allow extensions in a implicit/obfuscated way.

simo5:
Given you are removing the old build system, can you provide a amke file with targets that correspond to today high level targets that just turn around and run meson commands?

At least:
- make
- make check
- make check-stile
- make check-style-show
- make check-style-fix
- make generate-code
- make generate-docs

packaging/pkcs11-provider.spec
```diff
@@ -54,18 +51,17 @@ compatible to previous versions as well.
...
 
 %install
-%make_install
+%meson_install
 
 
 %check
 # do not run them in parrallel with %{?_smp_mflags}
-make check || if [ $? -ne 0 ]; then cat tests/*.log; exit 1; fi;
+%meson_test --num-processes 1
```
simo5, line 64: does this print logs on failure like the previous line did?

.github/workflows/scan-build.yml
```diff
@@ -13,24 +13,19 @@ jobs:
...
-          ./configure
+          meson setup _build
       - name: Scan Build
         run: |
-          scan-build --html-title="PKCS#11 Provider ($GITHUB_SHA)" \
-                    --keep-cc \
-                    --status-bugs \
-                    --keep-going \
-                    -o scan-build.reports make
+          SCANBUILD=$PWD/.github/scan-build.sh ninja -C _build scan-build
```
simo5, line 25: Why does this need to be spun out as an additional build script?

.github/workflows/scan-build.yml
```diff
@@ -13,24 +13,19 @@ jobs:
...
       - uses: actions/upload-artifact@v3
         if: failure()
         with:
           name: Scan Build logs
           path: |
-            scan-build.reports/
+            _build/meson-logs/scanbuild/
```
simo5, line 31: why change from a simple file to a directory?

.github/workflows/build.yml
```diff
@@ -54,35 +58,34 @@ jobs:
         uses: actions/checkout@v3
       - name: Setup
         run: |
-          autoreconf -fiv
-          CC=${{ matrix.compiler }} ./configure
+          CC=${{ matrix.compiler }} meson setup _build
```
simo5, line 61: I really dislike that meson forces you to create a build directory, however if it is necessary can we name it "builddir" as is customary (afaik) with meson projects?
Why prefix it with "_" anyway ?

simo5 (changes requested):
Mostly looks good, just some nitpicks

.github/workflows/style.yml
```diff
@@ -13,14 +13,11 @@ jobs:
     steps:
       - name: Install Dependencies
         run: |
-          dnf -y install $COMPILER automake libtool pkgconf-pkg-config \
-                autoconf-archive git openssl-devel clang-tools-extra \
+          dnf -y install $COMPILER meson pkgconf-pkg-config \
+                git openssl-devel clang-tools-extra \
```
simo5, line 17: I think that either automake or libtool may have been implicitly be pulling in a c compiler, but meson does not, so you'll have to explicitly now.

.github/workflows/scan-build.yml
ueno: This is the standard way of passing options to scan-build with Meson. See https://mesonbuild.com/howtox.html#use-clang-static-analyzer.

.github/workflows/build.yml
ueno: That is because some projects have `build` directly already. I can change it to `builddir`.

.github/workflows/scan-build.yml
ueno: Actually, it hasn't changed: the logs of scan-build are always stored in a directory, afaik.

.github/workflows/style.yml
ueno: But here, `$COMPILER` is specified; where does it come from?

.github/workflows/scan-build.yml
```diff
@@ -13,24 +13,19 @@ jobs:
     steps:
       - name: Install Dependencies
         run: |
-          dnf -y install $COMPILER automake libtool pkgconf-pkg-config \
-              autoconf-archive git openssl-devel clang-analyzer
+          dnf -y install $COMPILER meson pkgconf-pkg-config \
```
simo5, line 16: sounds like this $COMPILER variable is some spurious remnant, probably froma a copy of an old build.yml that was used to jump-start the scan-build action.
I think we can drop it, and/or, if needed, replace it with clang, but I guess clang-analyzer in the next line already pulls in all the required deps.

.github/workflows/style.yml
simo5: old remnant, but given we use clang to check style, perhaps we should just install clang instead of gcc as the compiler here.

.reuse/dep5
```diff
@@ -10,16 +10,20 @@ Files: **/Makefile.am
...
        .gitignore
        Makefile.am
+       meson.build
+       meson_options.txt
        *.md
        configure.ac
-       configure.ac
        src/Makefile.am
+       src/meson.build
        src/provider.exports
```
simo5, line 19: should we drop provider.exports here ?
Is it still used once we move to provider.map ?

Makefile.am
```diff
@@ -1,56 +0,0 @@
...
-		rm "$${gfile}.tmp"; \
-	done
-
-generate-docs:
-	for mdfile in docs/*.md; do \
-		echo "Processing $${mdfile}"; \
-		manfile=`echo $${mdfile} | sed s/\.md//`; \
-		pandoc --standalone --to man $${mdfile} -o $${manfile}; \
-	done
-
```
simo5, line 40: Where are the replacements for check-style, check-style-show, check-stuyle-fix, generate-code and generate-docs ?

simo5 (changes requested):
Current changes look good,
however there is loss of functionality in the removal of the base Makefile.am that needs to be resolved before this PR can be adopted.

.github/workflows/integration.yml
```diff
@@ -22,15 +22,13 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v4
       - name: Install Build Requirements
-        run: |
-          dnf -y install git autoconf autoconf-archive automake libtool \
-            openssl-devel
+        run: |\
```
simo5, line 25: this backslash is incorrect

.github/workflows/integration.yml
```diff
@@ -22,15 +22,13 @@ jobs:
...
+          dnf -y install git meson openssl-devel
       - name: Setup, Build and Install pkcs11-provider
         run: |
-          autoreconf -fiv
-          ./configure --libdir=/usr/lib64
-          make
-          make install
+          meson setup -Dlibdir=/usr/lib64 builddir
+          meson compile -C builddir
+          meson install -C builddir
```
simo5, line 31: Is there a way to set some variable so that -C builddir does not have to be repeated on each command ?

Jakuje:
> Uhm I wonder why the integration tests haven't run ...

Likely because of the backslash.

.reuse/dep5
ueno: Removed.

Makefile.am
ueno: I think we could add a Makefile for those "maintainer-mode" targets. Let me try in the next iteration.

.reuse/dep5
```diff
@@ -10,16 +10,20 @@ Files: **/Makefile.am
        .github/*
        .gitignore
        Makefile.am
+       meson.build
+       meson_options.txt
        *.md
        configure.ac
```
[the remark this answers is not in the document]
ueno: Removed.

Makefile.am
ueno: Done.

.github/workflows/address-sanitizer.yml
```diff
@@ -27,17 +27,17 @@ jobs:
       - name: Install Dependencies
         run: |
           if [ -f /etc/fedora-release ]; then
-            dnf -y install git clang gcc automake libtool pkgconf-pkg-config \
-              autoconf-archive openssl-devel openssl diffutils expect \
+            dnf -y install git clang gcc pkgconf-pkg-config meson \
+              openssl-devel openssl diffutils expect \
```
Jakuje, line 31: This is not a note for the meson change, but generic note to the way how the github actions are used here now.

When we were moving the CI of OpenSC from something previous to GH actions, I wanted to keep the bare minimum in the GH yml files and keep most of the logic in separate shell scripts, such as:

https://github.com/OpenSC/OpenSC/blob/master/.github/setup-linux.sh

This avoids when we change some dependencies to do it in half a dozen of yaml files.

This does not have to be changed as this PR, but I wanted to mention it here if you would find it useful, we can do it later.

meson.build
```diff
@@ -52,6 +52,12 @@ libssl = dependency('libssl', version: '>= 3.0.7')
 host_system = host_machine.system()
 if host_system == 'windows'
   shlext = '.dll'
+elif host_system == 'darwin'
+  # On macOS, loadable modules typically use .so as filename extension, but
+  # both NSS softokn and OpenSSL providers ignore this rule and use .dylib for
+  # loadable modules anyway. Note that p11-kit correctly uses .so, though, so
+  # be careful when using this variable, as it might not be correct on macOS.
+  shlext = '.dylib'
```
simo5, line 60: If the correct thing is to call these .so, why are we changing them to .dylib here?

.reuse/dep5
```diff
@@ -6,20 +6,15 @@ Source: https://github.com/latchset/pkcs11-provider/
 #
 # Build system, data files from tests, and misc cruft
 #
-Files: **/Makefile.am
-       .github/*
+Files: .github/*
        .gitignore
-       Makefile.am
+       GNUmakefile
```
simo5, line 11: Why GNUmakefile instead of the common Makefile?

simo5 (commented):
Just a couple nits/comments?

.reuse/dep5
ueno: There are a couple of minor motivations here. First, you wouldn't need to care about compatibility with other make implementations, as GNUmakefile is only recognized by GNU make. Secondly, with GNU make, you could write the current rules slightly simpler, using wildcards without shell functions (I included this change in https://github.com/latchset/pkcs11-provider/pull/304/commits/30c77f5acda9f11c143a4c37504ddb35809eb195 but can revert it if you don't like it).

meson.build
ueno: @neverpanic could you elaborate this?

.reuse/dep5
simo5: I think we are making it hard enough for other platforms already with the switch to meson, let's just use a standard Makefile please.

ueno: Reverted.

.github/workflows/integration.yml
ueno: Not that I know of, I'm afraid.

tests/test-wrapper
```diff
@@ -48,4 +48,4 @@ for option in "${TEST_PARAMS[@]}"; do
 done
 
 echo "Executing ${COMMAND}"
-${COMMAND}
+${COMMAND} 2>&1 | tee "${TESTBLDDIR}/${TEST_NAME}-${TOKEN_DRIVER}.log"
```
simo5, line 51: Hi Daiki,
do not worry about this, I already tested a correct solution (this one is broken, already been there) and will opena  PR for after we merge yours.

simo5 (changes requested):

src/signature.c
```diff
@@ -877,7 +877,9 @@ static CK_RV p11prov_sig_operate(P11PROV_SIG_CTX *sigctx, unsigned char *sig,
         if (sigctx->operation == CKF_VERIFY) {
             return CKR_ARGUMENTS_BAD;
         }
-        return p11prov_sig_get_sig_size(sigctx, siglen);
+        if (siglen) {
+            return p11prov_sig_get_sig_size(sigctx, siglen);
+        }
```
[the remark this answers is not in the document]
ueno: Done.

tests/test-wrapper
ueno: OK, I removed this commit.

simo5 (changes requested):
Ok I think we are really close.
The bind test failure is a regression in bind and we will waive it.

However I just realized that this PR will break the coverity scan.
Coverity relies on the command "make" to generate a build, it may even pass variables there.

Unfortunately it cannot be easily tested on a PR as the coverity scan can only run against an action already merged in the main tree.

But I can attempt a test in a local checkout to confirm that just providing a make :all target that call meson compile will work

simo5:
So I think coverity depends only on a PATH change, injecting a different gcc in a coverity/bin path or similar.

How will meson interact with that ?

When I run `meson compile -vC builddir` it gives me what looks like a pseudo-command line as it identifies the compiler as just 'cc', so iut is not clear to me if meson executes the compiler via an abslute path or will allow interposition by calling the compiler with it's relative name.

simo5:
I did find his post though:
https://community.synopsys.com/s/question/0D5Hr00006VdMFsKAN/in-ninja-build-environment-covbuild-cannot-emit-c-files

Which means you may need to tweak the .github/workflows/coverity-scan.yml steps before we can merge this PR, sorry for not realizing this earlier.

ueno:
> I did find his post though: https://community.synopsys.com/s/question/0D5Hr00006VdMFsKAN/in-ninja-build-environment-covbuild-cannot-emit-c-files
> 
> Which means you may need to tweak the .github/workflows/coverity-scan.yml steps before we can merge this PR, sorry for not realizing this earlier.

I added `env.CC: gcc` in the setup step, following this [example](https://github.com/OpenRC/openrc/blob/3d30b6fddaf92c612deea88d2ced7114ed1fcf9c/.github/workflows/coverity.yml#L19) linked from the coverity-scan-action [README](https://github.com/vapier/coverity-scan-action/blob/main/README.md#live-examples). Hopefully it should work.

simo5:
@dueno we sill need a Makefile all target to call meson compile as the covscan build just calls `make` with no arguments

ueno:
> @dueno we sill need a Makefile all target to call meson compile as the covscan build just calls `make` with no arguments

I can add such make rule, but doesn't coverity-scan-action take care of that already, given the openrc [example](https://github.com/OpenRC/openrc/blob/HEAD/.github/workflows/coverity.yml) mentioned above doesn't have Makefile at all?

simo5:
I am going to ignore the MacOS failure as it seems a homebrew screwup, as all of this was working until today.
Hopefully it will get fixed soon, and if not we'll handle it separately anyway.

simo5 (approved):

neverpanic:
The macOS CI failure is a Homebrew screw-up, see https://github.com/Homebrew/homebrew-core/issues/169728.

I can locally reproduce:
```
$ podman pull ghcr.io/homebrew/core/openssl/3:3.3.0@sha256:636c1743241379d29fb964fcccce1ea86f46fa66a85319c1ac59c9698e4ac809  # this is the pre-compiled OpenSSL for Homebrew
$ podman image save 1f09999b3994 | tar xO f35bb52aadbdae43149bf5eb3a3a4948c1149879d13d2041e0139cbd80a08535.tar | tar tv | grep libcrypto.dylib
lrwxrwxrwx  0 0      0           0 Apr  9 14:12 openssl@3/3.3.0/lib/libcrypto.dylib -> libcrypto.3.dylib
$ podman image save 1f09999b3994 | tar xO f35bb52aadbdae43149bf5eb3a3a4948c1149879d13d2041e0139cbd80a08535.tar | tar xO openssl@3/3.3.0/lib/pkgconfig/libcrypto.pc | head -1
libdir=@@HOMEBREW_CELLAR@@/openssl@3/3.3.0
```

^ there's a `/lib` missing at the end of that path in the pkg-config file.

PR #307, opened 2023-11-13
Fix: enable cross compilation by removing the AC_CHECK_FILE macro

oleorhagen:
From the man-page on the `AC_CHECK_FILE` macro, we can read that:

```
You might also need to check for the existence of files. Before using these macros, ask yourself whether a runtime test might not be a better solution. Be aware that, like most Autoconf macros, they test a feature of the host machine, and therefore, they die when cross-compiling.
```

Therefore, move to using a simple shell if/else as the replacement.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -128,11 +128,13 @@ PKG_CHECK_EXISTS(
         )]
 )
 
-AC_CHECK_FILE($SOFTOKENDIR/libsoftokn3$SHARED_EXT, [],
-    [AC_CHECK_FILE($SOFTOKENDIR/nss/libsoftokn3$SHARED_EXT,
-        [AC_SUBST([SOFTOKEN_SUBDIR], "nss/")],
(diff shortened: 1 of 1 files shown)


simo5:
Not sure why this is needed, if you fail the check simply nss testing is unavailable, which is what you want during a cross compile anyway, because you can't test cross compiled binaries directly on the build host.
What am I missing ?

oleorhagen:
> Not sure why this is needed, if you fail the check simply nss testing is unavailable, which is what you want during a cross compile anyway, because you can't test cross compiled binaries directly on the build host. What am I missing ?

The `AC_CHECK_FILE` macro will literally exit the with the error: 

` configure: error: cannot check for file existence when cross compiling`

when cross compiling :disappointed:

simo5:
I see, that is annoying.
I will merge this.

oleorhagen:
Thanks :smile_cat:

PR #309, opened 2023-11-16
Implement support for ALWAYS_AUTH and interactive prompting when the caller did not provide pin

### LINKED ISSUE #295: Consider a default callback call for PIn requests
When loading a key with ENGINE_load_private_key() the openssl-pkcs11 engine is able to prompt the user even if no explicit ui_method and callback are provide (ie they are NULL).

This is because internally the engine receives the ui_method directly and call UI_new_method(ui_method) which returns a default implementation from openssl if ui_method is null.

A clear and concise description of what the request is.

Through the store API we only get a OSSL_PASSPHRASE_CALLBACK and if that is null we have no way to call back the user.

Consider adding a cal to UI_new_method(NULL) to use the default OpenSSL prompter implementation if a custom implementation was not provided.

See https://github.com/systemd/systemd/pull/29539 for the discussion that prompted this.

Jakuje:
Based on #298. The current version passes all existing tests and the default UI is not gated with a configuration option (but this can be added if needed).

Fixes #295 and #42


src/signature.c
```diff
@@ -833,10 +833,12 @@ static CK_RV p11prov_sig_operate_init(P11PROV_SIG_CTX *sigctx, bool digest_op,
         slotid = p11prov_obj_get_slotid(sigctx->key);
 
         ret = mech_fallback_init(sigctx, slotid);
+        goto done;
```
simo5, line 836: Is this one correct?
mech_fallback_init() is a valid case, why would we skip login?
Maybe this branch needs to retry to grab a session instead (or cause mech_fallback_init() to return the session it allocates ?

src/session.c
```diff
@@ -431,8 +473,15 @@ static CK_RV token_login(P11PROV_SESSION *session, P11PROV_URI *uri,
             };
             ret = pw_cb(cb_pin, sizeof(cb_pin), &cb_pin_len, params, pw_cbarg);
             if (ret != RET_OSSL_OK) {
-                ret = CKR_GENERAL_ERROR;
-                goto done;
+                /* this error can mean anything from the user canceling the prompt to no UI method provided
+                 * Fall back to our prompt here */
```
simo5, line 477: if we did have a callback I do not think we should repeat the prompting,
we should prompt on our own if no callback is present, gated behind a config option that enables this behavior.

src/session.c
```diff
@@ -443,10 +492,19 @@ static CK_RV token_login(P11PROV_SESSION *session, P11PROV_URI *uri,
             pinlen = cb_pin_len;
 
             cache = p11prov_ctx_cache_pins(session->provctx);
-
         } else {
-            ret = CKR_CANCEL;
-            goto done;
+            /* We are asking the user off-band for the user consent -- from store
+             * we will always receive non-null (but unusable) callback */
+            ret = p11prov_session_get_pin(slot, (char *)cb_pin, &cb_pin_len);
```
simo5, line 498: let's gate this, see above

src/session.c
```diff
@@ -443,10 +492,19 @@ static CK_RV token_login(P11PROV_SESSION *session, P11PROV_URI *uri,
...
+            if (ret != RET_OSSL_OK) {
+                ret = CKR_GENERAL_ERROR;
+                goto done;
+            }
+
+            pin = (CK_UTF8CHAR_PTR)cb_pin;
+            pinlen = strlen(cb_pin);
+
+            cache = p11prov_ctx_cache_pins(session->provctx);
```
simo5, line 507: maybe move this after the if/else block so we do it only once ?

simo5 (commented):
in general it looks good to me, but there are some nits we can discuss.

src/session.c
Jakuje: Done.

src/session.c
Jakuje: As described in this comment, from store, we are always getting a callback, but if the calling application does not provide the UI, it just fails. I think we are getting the pointer to `ossl_pw_get_passphrase()` function from `crypto/passphrase.c` unconditionally. And based on what application provided, it works of fails with `No password method specified`:
https://github.com/openssl/openssl/blob/cf6342bc024868f5a55f2225f2e083415fb1329a/crypto/passphrase.c#L275

src/session.c
Jakuje: Ok, this wont work as it tries to cache the cached pin, which goes into user-after-free. Indenting this one more tab will make it more ugly, but works. We can reiterate later though.

src/session.c
```diff
@@ -387,10 +387,49 @@ CK_SLOT_ID p11prov_session_slotid(P11PROV_SESSION *session)
     return session->slotid;
 }
 
+static int p11prov_session_get_pin(struct p11prov_slot *slot, char *cb_pin,
```
simo5, line 390: Can you rename this one to p11prov_session_prompt_for_pin() ?

src/session.c
```diff
@@ -421,18 +463,39 @@ static CK_RV token_login(P11PROV_SESSION *session, P11PROV_URI *uri,
...
+                                                  &cb_pin_len);
+                    if (ret != RET_OSSL_OK) {
+                        /* give up */
+                        ret = CKR_GENERAL_ERROR;
+                        goto done;
+                    }
+                }
+            } else {
+                /* We are asking the user off-band for the user consent -- from
+                 * store we will always receive non-null (but unusable) callback
```
simo5, line 491: Not sure why this is unusable? When we receive a callback it is always usable, the problem is that often we just do not have one, and we can't store the callback we were provided for the store operation for later operations as the calling application may not expect it.

simo5 (approved):
Only two very minor nits, if you can quickly address them I'd like to merge asap.

src/session.c
Jakuje: Done

src/session.c
Jakuje: As discussed above, from the store we will always get a callback (at least in OpenSSL in Fedora 38), but the callback just fails with the error mentioned above.

Jakuje:
> Only two very minor nits, if you can quickly address them I'd like to merge asap.

So lets keep it for now without the configuration option discussed previously?

Jakuje:
Noting here, that the macos CI picked up gnutls 3.8.2 with the following regression in handling ed25519 keys: https://gitlab.com/gnutls/gnutls/-/issues/1515

Jakuje:
> Noting here, that the macos CI picked up gnutls 3.8.2 with the following regression in handling ed25519 keys: https://gitlab.com/gnutls/gnutls/-/issues/1515

And now we have the broken gnutls also in Fedora 39. The update to Fedora 38 is on the way too:

https://bodhi.fedoraproject.org/updates/FEDORA-2023-e075ac32be

simo5:
Merged, and opened a followup issue for the gating config option.

src/session.c
Jakuje: Its the conversion from the UI methods on the store API to the low-level callbacks on the provider level. What you describe would make sense and I would be assuming this, but this is OpenSSL. So when the store API is called without the UI method, the provider gets non-null callback, but that one fails because there is no assigned UI method. Sounds as stupid as it is (or I followed the gdb wrongly -- there is obviously no documentation. You can double-check that this wont work if you remove this code branch).

PR #314, opened 2023-11-28
tests: Fix expected output and fix Debian package name

Jakuje:
After Debian pulled fixed NSS version, we started getting the failure where the basic test expected public key data to be printed when we selected private key. This was probably the case in some openssl versions, but given that the test reads explicitly private key uri, I think checking for the private key information makes more sense.


simo5 (commented):
LGTM,
Fixes Debian issues

PR #316, opened 2023-11-30
Use a cached version of the URI to refresh objects

### LINKED ISSUE #313: Key handle invalid after fork - when PIN is supplied via URI
I am currently trying to enhance Apache httpd to allow it to use this PKCS#11 provider and load the keys via PKCS#11 URI from the STORE. 

In general this works fine, but so far only in httpd's debug mode (`httpd -X`) where it does not fork.
Once I run httpd in normal mode, it forks and this is where the problem starts.

Here is the debug log from after the fork:

```
[keymgmt.c:1287] p11prov_ec_export(): ec export 0x3095880
[provider.c:584] p11prov_ctx_allow_export(): allow_export = 0
[signature.c:1818] p11prov_ecdsa_digest_sign_init(): ecdsa digest sign init (ctx=0x3ff7000fe60, digest=SHA256, key=0x3095880, params=(nil))
[interface.c:395] p11prov_module_reinit(): PKCS#11: Re-initializing the module: libopencryptoki.so
[interface.gen.c:31] p11prov_Finalize(): Calling C_Finalize
[interface.gen.c:35] p11prov_Finalize(): Error: 0x00000190; Error returned by C_Finalize
[interface.gen.c:12] p11prov_Initialize(): Calling C_Initialize
[interface.gen.c:50] p11prov_GetInfo(): Calling C_GetInfo
[objects.c:400] p11prov_obj_ref_no_cache(): Ref Object: 0x3095880 (handle:0)
[provider.c:602] p11prov_ctx_cache_keys(): cache_keys = 1
[session.c:919] p11prov_take_login_session(): Get login session from slot 4
[interface.gen.c:214] p11prov_OpenSession(): Calling C_OpenSession
[session.c:73] token_session_open(): C_OpenSession ret:0 (session: 1)
[interface.gen.c:253] p11prov_GetSessionInfo(): Calling C_GetSessionInfo
[objects.c:368] cache_key(): Failed to get login session. Error 1
[signature.c:2028] p11prov_ecdsa_set_ctx_params(): ecdsa set ctx params (ctx=0x3ff7000fe60, params=(nil))
[objects.c:431] p11prov_obj_free(): Free Object: 0x3095880 (handle:0)
[objects.c:438] p11prov_obj_free(): object free: reference held
[keymgmt.c:1363] p11prov_ec_get_params(): ec get params 0x3095880
[keymgmt.c:1287] p11prov_ec_export(): ec export 0x3095880
[provider.c:584] p11prov_ctx_allow_export(): allow_export = 0
[signature.c:1818] p11prov_ecdsa_digest_sign_init(): ecdsa digest sign init (ctx=0x3ff700143c0, digest=SHA2-256, key=0x3095880, params=(nil))
[objects.c:400] p11prov_obj_ref_no_cache(): Ref Object: 0x3095880 (handle:0)
[provider.c:602] p11prov_ctx_cache_keys(): cache_keys = 1
[session.c:919] p11prov_take_login_session(): Get login session from slot 4
[interface.gen.c:214] p11prov_OpenSession(): Calling C_OpenSession
[session.c:73] token_session_open(): C_OpenSession ret:0 (session: 2)
[interface.gen.c:253] p11prov_GetSessionInfo(): Calling C_GetSessionInfo
[objects.c:368] cache_key(): Failed to get login session. Error 1
[signature.c:2028] p11prov_ecdsa_set_ctx_params(): ecdsa set ctx params (ctx=0x3ff700143c0, params=(nil))
[signature.c:1862] p11prov_ecdsa_digest_sign_final(): ecdsa digest sign final (ctx=0x3ff700143c0, sig=(nil), siglen=0, sigsize=0)
[signature.c:1841] p11prov_ecdsa_digest_sign_update(): ecdsa digest sign update (ctx=0x3ff700143c0, data=0x3ff7eafe766, datalen=146)
[signature.c:780] p11prov_sig_operate_init(): called (sigctx=0x3ff700143c0, digest_op=true)
[objects.c:1030] find_associated_obj(): Find associated object
[session.c:765] p11prov_get_session(): Get session on slot 4, reqlogin=true, rw=false
[session.c:774] p11prov_get_session(): single-shot request for slot 4
[session.c:494] check_slot(): Checking Slot id=4, uri=(nil), mechtype=ffffffffffffffff, rw=false)
[provider.c:590] p11prov_ctx_login_behavior(): login_behavior = 0
[session.c:266] session_new(): Creating new P11PROV_SESSION session on pool 0x30dddb0
[session.c:300] session_new(): Total sessions: 3
[interface.gen.c:214] p11prov_OpenSession(): Calling C_OpenSession
[session.c:73] token_session_open(): C_OpenSession ret:0 (session: 3)
[interface.gen.c:253] p11prov_GetSessionInfo(): Calling C_GetSessionInfo
[session.c:102] token_session_close(): Closing session 3
[interface.gen.c:233] p11prov_CloseSession(): Calling C_CloseSession
[signature.c:790] p11prov_sig_operate_init(): Error: 0x00000060; Provided key has invalid handle
[objects.c:431] p11prov_obj_free(): Free Object: 0x3095880 (handle:0)
[objects.c:438] p11prov_obj_free(): object free: reference held
```

I debugged it so far that I can see that `obj->raf` is true (as expected after fork), and thus it runs into `p11prov_obj_refresh()` where it calls `find_associated_obj()` and there it gets a new session via `p11prov_get_session()`.  `p11prov_get_session()` calls `slot_login()` which finally calls `token_login()`. In `token_login()` it tries to get the PIN, but does not find any PIN, because there is no cached pin, and no URI either. `find_associated_obj()` calls `p11prov_get_session()` with NULL as URI and also the passphrase callback is NULL. 

The log also shows failures from `cache_key()` where it calls `p11prov_take_login_session()` which also calls `slot_login()` with NULL for URI and passphrase callback. So no re-login possible with PIN from URI either. 

So the codes does not seem to be prepared for re-logging in a session when the PIN is provided via URI only.

Once I set the PIN via `pkcs11-module-token-pin` provider config option in the OpenSSL config file, it works.

simo5:
This allows us to retain better information about the object and critically things like a PIN passed in the URI to be able to login to the token after a fork when the PIN is not specified in the token configuration.

Fixes #313

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -40,6 +40,8 @@ struct p11prov_obj {
     CK_BBOOL cka_copyable;
     CK_BBOOL cka_token;
 
+    P11PROV_URI *refresh_uri;
+
     union {
         struct p11prov_key key;
         struct p11prov_crt crt;
@@ -1031,6 +1033,9 @@ CK_RV p11prov_obj_find(P11PROV_CTX *provctx, P11PROV_SESSION *session,
             /* unknown object or other recoverable error to ignore */
             continue;
         } else if (ret == CKR_OK) {
+            /* keep a copy of the URI for refreshes as it may contain
+             * things like a PIN necessary to log in */
+            obj->refresh_uri = p11prov_copy_uri(uri);
             ret = cb(cb_ctx, obj);
         }
         if (ret != CKR_OK) {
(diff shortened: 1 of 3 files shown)


simo5:
Sigh softhsm2 setup is now broken across the board as more distributions get GnuTLS 3.8.2 ...

src/objects.c
```diff
@@ -1084,13 +1089,86 @@ static P11PROV_OBJ *find_associated_obj(P11PROV_CTX *provctx, P11PROV_OBJ *obj,
...
+    ret = p11prov_FindObjects(obj->ctx, sess, &handle, 1, &objcount);
+
+    /* Finalizing is not fatal so ignore result */
+    p11prov_FindObjectsFinal(obj->ctx, sess);
+
+    if (ret != CKR_OK) {
+        P11PROV_debug("Failed to find refresh object %p", obj);
+        goto done;
+    }
+    if (objcount != 1) {
```
sahanaprasad07, line 1159: Maybe we should distinguish the cases  objcount = 0 and objcount > 1 here.
And if objcount > 1, should we pick the first one? (Although we expect a single entry?)

simo5: No, if we pick the wrong one we could cause cryptographic operations to happen using the wrong key, that would be CVE worthy material.

src/objects.c
```diff
@@ -1084,13 +1089,86 @@ static P11PROV_OBJ *find_associated_obj(P11PROV_CTX *provctx, P11PROV_OBJ *obj,
...
+        P11PROV_raise(obj->ctx, ret,
+                      "Too many objects found on refresh (count=%ld)",
+                      objcount);
```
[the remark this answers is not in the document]
simo5: ok will do

PR #319, opened 2023-12-01
tests: tls test without sleep

0pq76r:
I ran into some flakeyness when running the TLS tests with a HSM.

I changed the test to use expect such that the timing for sending stuff to the server and the client process depends on the program output.

### DIFF
--- a/tests/ttls
+++ b/tests/ttls
@@ -4,60 +4,59 @@
 
 source "${TESTSSRCDIR}/helpers.sh"
 
-SLEEP=0.5
-# with valgrind/asan, it might take a bit longer
-if [ -n "$CHECKER" ]; then
-    SLEEP=10
-fi
-
 title PARA "Test SSL_CTX creation"
 $CHECKER ./tlsctx
 
 title PARA "Test an actual TLS connection"
-rm -f "${TMPPDIR}/s_server_input"
+
 rm -f "${TMPPDIR}/s_server_output"
+rm -f "${TMPPDIR}/s_server_ready"
+mkfifo "${TMPPDIR}/s_server_ready"
 
-# Set up command fifo
-mkfifo "${TMPPDIR}/s_server_input"
-exec 3<>"${TMPPDIR}/s_server_input"
-
+SERVER_PID=-1
 # Make sure we terminate programs if test fails in the middle
 # shellcheck disable=SC2317  # Shellcheck for some reason does not follow trap
-kill_children_print() {
-    kill_children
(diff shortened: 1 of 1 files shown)


simo5:
Ignore the softhsm failures, there is a bug in GnuTLS 3.8.2 that causes it

0pq76r:
I don't have access to a Mac at the moment. I'll try to fix it but it may take a few days.

simo5 (approved):
LGTM!
Not sure why shellcheck is complaining, looks like a false positive
@Jakuje could you review and merge if it is good for you as well.

Jakuje:
> LGTM! Not sure why shellcheck is complaining, looks like a false positive @Jakuje could you review and merge if it is good for you as well.

There was a comment that was removed which disabled this check as I probably hit the same issue before:
```
# shellcheck disable=SC2317  # Shellcheck for some reason does not follow trap
```
Returning this back should silence the shellcheck.

tests/ttls
```diff
@@ -4,60 +4,57 @@
...
+             expect {
+               \"END SSL SESSION PARAMETERS\" {};
+               default {exit 1;};
+             }
+             send \" TLS SUCCESSFUL \n\"
+	     send \"Q\n\"
+             expect {
+               eof {exit 0;};
+               default {exit 1;};
+             }" > "${TMPPDIR}/s_server_output" &
```
Jakuje, line 44: only nit I would have against the extensive first indent and just two spaces indent in the expect blocks. I would just keep it flat with 4 spaces for each. The same for the expect below.

tests/ttls
```diff
@@ -4,60 +4,57 @@
...
 
-# Set up command fifo
-mkfifo "${TMPPDIR}/s_server_input"
-exec 3<>"${TMPPDIR}/s_server_input"
+rm -f "${TMPPDIR}/s_server_output"
+rm -f "${TMPPDIR}/s_server_ready"
+mkfifo "${TMPPDIR}/s_server_ready"
 
-# Make sure we terminate programs if test fails in the middle
-# shellcheck disable=SC2317  # Shellcheck for some reason does not follow trap
```
Jakuje, line 25: keep this comment before the trap function to avoid the shellcheck failure.

Jakuje (approved):
There are some memory leaks in the CI (as fedora already picked up good versions of gnutls package), but they are not related to your changes so after squashing the changes I think we are good to go.

simo5:
Thanks @0pq76r this looks good now, merging.

PR #324, opened 2023-12-05
Quirk: no-operation-state

0pq76r:
Configuration property to indicate whether the PKCS11 provider supports getting/setting the operation state. If it is not supported, no attempts will be made to duplicate the context.

I already mentioned it in the discussion (https://github.com/latchset/pkcs11-provider/discussions/320) that the current approach leaves traces in the SSL error stack.

Please let me know what you think of this change.

(Tests will follow)

### DIFF
--- a/docs/provider-pkcs11.7
+++ b/docs/provider-pkcs11.7
@@ -170,17 +170,25 @@ immediately at application startup)
 .PP
 Workarounds that may be needed to deal with some tokens and cannot be
 autodetcted yet are not appropriate defaults.
+.SS no-deinit
 .PP
-The only quirk currently implemented is \[lq]no-deinit\[rq] It prevent
-de-initing when OpenSSL winds down the provider.
+It prevents de-initing when OpenSSL winds down the provider.
 NOTE this option may leak memory and may cause some modules to misbehave
 if the application intentionally unloads and reloads them.
+.SS no-operation-state
+.PP
+OpenSSL by default assumes contexts with operations in flight can be
+easily duplicated.
+That is only possible if the tokens support getting and setting the
+operation state.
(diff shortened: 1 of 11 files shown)


docs/provider-pkcs11.7
```diff
@@ -170,17 +170,26 @@ immediately at application startup)
...
-The only quirk currently implemented is \[lq]no-deinit\[rq] It prevent
-de-initing when OpenSSL winds down the provider.
+It prevents de-initing when OpenSSL winds down the provider.
 NOTE this option may leak memory and may cause some modules to misbehave
 if the application intentionally unloads and reloads them.
+.SS no-operation-state
+.PP
+OpenSSL by default assumes contexts with operations in flight can be
+easily duplicated.
+That is only possible is the tokens support getting and setting the
```
simo5, line 182: is -> if

docs/provider-pkcs11.7
```diff
@@ -170,17 +170,26 @@ immediately at application startup)
...
+operation state.
+If the quirk is enabled the context duplication only performs a context
+move.
```
simo5, line 185: Just say "Context Duplication is not performed" because of: #209 which I plan to merge sometime soon once I verify my patch to deal with this properly is actually in openssl 3.2.0

simo5 (approved):
Only minor nits on the docs text, otherwise LGTM

simo5 (approved):
LGTM

simo5:
Thanks, this looks good.
The only minor thing is that we should probably apply this to all operations state requests not just signature ones.
I'll file a followup issue to look at doing that in a separate PR though.

PR #325, opened 2023-12-05
CKK_EC_EDWARDS: EC_PARAMS containing oID

0pq76r:
I have a token where the EC_PARAMS describe the curve using the oID  as specified in RFC 8410.

I imagine that the proposed solution is fairly inefficient compared to the memcmp that is currently done for the PrintableString representation. Would you recommend to introduce new constants for the oId  instead? 
Or do you know of an efficient solution that supports both/all valid representations?


Jakuje (approved):
looks good!

simo5:
Merged, thanks a lot!

PR #326, opened 2023-12-06
objects: Avoid memory leak

Jakuje:
This was introduced in 9855e8eaf40d8c047245bc1767ab4a0b7b2220ab


simo5:
Obviously correct and quite interesting that Coverity has not picked it up ...

Issue #222, opened 2023-03-30
Support key in PEM file (like TPM2 TSS KEY)


quality-leftovers:
**Describe the feature**
To integrate the provider with existing / COTS applications it would be nice to be able to provide the key in PEM format, likt it is possible in `tpm2-tss`.

**Expected behavior**
It should be possible to generate PEMs for keys and provide them to an application similar to "TPM2 TSS KEY"

**Additional context**
I've played around a bit with the pkcs11-provider and am trying to add the ability to load a key from a pem.
=> https://github.com/quality-leftovers/pkcs11-provider/commit/138a909c93020e3bca26b9fae0b4adc1c9aafcb7

So far I've just taken the encoder / decoder bits from the tpm2-tss project and integrated them (https://github.com/tpm2-software/tpm2-tss) and now I'm wondering what the best way to load a key from a decoder would be.

Haven't done a lot with OpenSSL so far, so it's a bit difficult for me :)

simo5:
What would the PEM file consist of?
A pkcs11 URI?

simo5:
I am not entirely sure about this, the code seem straightforward, but I question whether it makes sense to do really provide this level of obfuscation.
With TPM it may be needed because there is no standard URI scheme (afaik), so a PEM file may be a reasonable option.
But with pkcs11 we do have standard URIs and the OpenSSL 3 STORE API takes URIs just fine.

So what is the tradeoff here? Can you provide some pros/cons to help me evaluate if this is really desirable for the pkcs11 provider?

quality-leftovers:
It is basically all about enabling "Plug & Play" of the pkcs11-provider with existing software.

Existing software that uses OpenSSL often does not support OpenSSL 1 key engines or OpenSSL 3 provider URIs.
These software packages usually load key "PEMs" using `BIO_read_filename` and `PEM_read_bio_PrivateKey` and will not accept key URIs.

For example there is an old issue in haproxy (https://github.com/haproxy/haproxy/issues/71) where support to load an key from a pkcs11 OpenSSL key engine was requested, but it never did go anywhere (probably because haproxy relies on loading keys from a file named <certificate-file-name>.<key-suffix> and changing that would be a breaking changes in the haproxy configuration format - as far as I understand it).

**Pros**
- Adding key loading via PEM would open the door to use pkcs11 HSM with popular existing software solutions
- Offering functionality to lots of people is more important then having it done the "correct" way

**Cons**
- Additional code / complexity (less transparent)
[...]

PR #328, opened 2023-12-18
Decoder: pkcs11-uri in pem

0pq76r:
This is an attempt to implement the feature "Support key in PEM file" #222.

The PEM file consists out of two fields an ASN1_OBJECT called "type" containing an oId to identify the object and an ASN1_UTF8STRING called "uri". The URI is parsed and created using the existing `p11prov_parse_uri` and `p11prov_key_to_uri` functions.

There are still some points left to improve:
- Flag to disable this non-standardized feature
- Improve variable/file naming (keypair_ref, active_object, ...)
- Find a more appropriate oId (Currently, I'm using "2.5.4.83"  == {joint-iso-itu-t(2) ds(5) attributeType(4) uri(83)})
- Add tests


0pq76r:
> There are still some points left to improve:
> 
> ~* Flag to disable this non-standardized feature~

The encoder has to be enabled by setting  "pkcs11-module-encode-key-uri-to-pem = true".

> ~* Improve variable/file naming (keypair_ref, active_object, ...)~

Slightly improved but I'm terrible at naming stuff. Suggestions are welcome.

> * Find a more appropriate oId (Currently, I'm using "2.5.4.83"  == {joint-iso-itu-t(2) ds(5) attributeType(4) uri(83)})

This still needs to be addressed. Dropping the oId and just encoding an UTF8String could also be an option.

> ~* Add tests~

Tests showed that I need to loop over all tokens in the decoder. A similar loop is used in the store_fetch function, it could make sense to re-use that code.

I added libtasn1 as dependency for generating pem files in the tests. Originally I planed to use ossl asn1parse but it is broken in 3.2.0 (https://github.com/openssl/openssl/issues/22992)

simo5:
Do we really need this whole ASN.1 dependency when all we are doing is the simplest ASN.1 thing ever?
Pretty sure we can basically hardcode a generator if it is for testing only, given the only variable here is the length of the UTF8String and it's content.

0pq76r:
I removed the dependency, what else needs to be changed? Do you have any opinion about the oId? Should it be removed completely?

simo5:
Does it need to be an OID?
What if we have a regular string as the first member of the asn.1 structure and we put in a value like: "PKCS#11 Provider URI v1.0" ?

simo5:
If OpenSSL expects OIDs (I have not checked) I can provide one from the Red Hat pool, let me know and I can assign.

0pq76r:
It's not necessary to have an OID. I update the code to have a ASN1_VISIBLESTRING instead.
Though, I'm starting to wonder if the decoder is the right mechanism. My gut tells me that there should be a way to return the URI to the store and have it do the actual key lookup. But I have not figured out how that would work (or if it is possible at all).

src/decoder.c
```diff
@@ -0,0 +1,393 @@
...
+    obj = d2i_P11PROV_PK11_URI(NULL, &der, der_len);
+    if (!obj) {
+        P11PROV_debug("P11 KEY DECODER d2i_P11PROV_PK11_URI failed");
+        goto done;
+    }
+
+    if (!obj_desc_verify(obj)) {
+        goto done;
+    }
+
```
simo5, line 283: Perhaps before decoding the URI we should check that obj->desc corresponds to our identification string?
Or is this assumed by the fact the decoder was called at all ?

src/decoder.c
```diff
@@ -0,0 +1,393 @@
...
+        return CKR_GENERAL_ERROR;
+    }
+
+    P11PROV_debug("Adding object (handle:%lu)", p11prov_obj_get_handle(obj));
+    ctx->object = obj;
+
+    return CKR_OK;
+}
+
+static CK_RV p11prov_decoder_load_obj(P11PROV_DECODER_CTX *ctx,
```
simo5, line 146: Why have you re-implemented store logic in here instead of just calling p11prov_store_open() followed by p11prov_store_load() ?

I do not think replicating the store logic outside of the store is good for maintainability.

[Of course you can have a helper wrapper function in src/store.c so that we do not need to export all of these static functions individually]

src/decoder.c
```diff
@@ -0,0 +1,393 @@
...
+
+    uri = (const char *)ASN1_STRING_get0_data(obj->uri);
+    uri_len = ASN1_STRING_length(obj->uri);
+    /* todo check string ends in \0 */
+    if (!uri || uri_len <= 0) {
+        P11PROV_debug("Failed to get URI");
+        goto done;
+    }
+
+    if (p11prov_decoder_load_obj(ctx, uri, pw_cb, pw_cbarg) != CKR_OK) {
```
simo5, line 292: Actually if you call the openssl store API here that wouldn't that also remove the need for p11prov_obj_to_ossl_obj() as well?

tests/tpem_encoder
```diff
@@ -0,0 +1,161 @@
...
+    "${OPENSSL_CONF}" > "${OPENSSL_CONF}.encode_to_pem"
+OPENSSL_CONF=${OPENSSL_CONF}.encode_to_pem
+make-uri-pem() {
+    export LC_ALL=C
+
+    URI=$1
+    OUT=$2
+    DESC="${3:-PKCS#11 Provider URI v1.0}"
+
+    DESC_HEX=$(printf '%s' "${DESC}" | perl -ne 'print unpack "H*", $_')
```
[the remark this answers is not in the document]
0pq76r: Perl is a build dependency of openssl, so I thought it's ok to use. I can't remember why, but I had an issue with xxd (I think it behaves somehow differently on macos). I'm going to change it such that it only uses printf and od.

src/decoder.c
0pq76r: The `obj_desc_verify(obj)` compares the obj->desc with the identification string.

0pq76r:
> I realized that perhaps you should just call the store api directly via OSSL_STORE_open_ex() with the uri and the provider private libctx and avoid re-implementing a bunch of store functionality. Or have you already tried that and found a showstopper?

The OSSL_STORE returns an EVP_PKEY but the decoder is supposed to call the object_cb with an OSSL_PARAM array. The EVP_PKEY_todata calls the export method which for obvious reasons does not work for private keys.

I think calling the p11prov_store_*() is the way to go.

src/decoder.c
simo5: I had missed that one, thanks!

tests/tpem_encoder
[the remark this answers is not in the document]
simo5: I have a personal dislike of perl :-D

simo5:
> > I realized that perhaps you should just call the store api directly via OSSL_STORE_open_ex() with the uri and the provider private libctx and avoid re-implementing a bunch of store functionality. Or have you already tried that and found a showstopper?
> 
> The OSSL_STORE returns an EVP_PKEY but the decoder is supposed to call the object_cb with an OSSL_PARAM array. The EVP_PKEY_todata calls the export method which for obvious reasons does not work for private keys.
> 
> I think calling the p11prov_store_*() is the way to go.

Ah, that's right!

src/Makefile.am
```diff
@@ -19,6 +21,7 @@ noinst_HEADERS = \
 	signature.h \
 	slot.h \
 	store.h \
+	store_internals.h \
```
simo5, line 24: We already have a store.h, no need for a new header

src/decoder.c
```diff
@@ -0,0 +1,253 @@
...
+
+static char *obj_uri_get1(P11PROV_PK11_URI *obj)
+{
+    const unsigned char *uri = ASN1_STRING_get0_data(obj->uri);
+    int uri_len = ASN1_STRING_length(obj->uri);
+    if (!uri || uri_len <= 0) {
+        P11PROV_debug("Failed to get URI");
+        return NULL;
+    }
+    char *null_terminated_uri = OPENSSL_zalloc(uri_len + 1);
```
simo5, line 60: Use p11prov_alloc_sprintf here.

src/store.c
```diff
@@ -589,6 +590,26 @@ static int p11prov_store_set_ctx_params(void *pctx, const OSSL_PARAM params[])
...
+    if (!ctx) {
+        goto done;
+    }
+
+    while (
+        RET_OSSL_OK
+        == p11prov_store_load(ctx, object_cb, object_cbarg, pw_cb, pw_cbarg)) {
+        ;
+    }
+
```
simo5, line 608:
```
int ret = RET_OSSL_OK;
while (ret == RET_OSSL_OK) {
    ret = p11prov_store_load(ctx, object_cb, object_cbarg, pw_cb, pw_cbarg);
}
```

Also please return an error so you can bubble it up, there is no guarantee the load will succeed.

src/store.c
```diff
@@ -589,6 +590,26 @@ static int p11prov_store_set_ctx_params(void *pctx, const OSSL_PARAM params[])
     return RET_OSSL_OK;
 }
 
+void p11prov_store_quickload(void *provctx, const char *uri,
```
simo5, line 593: There isn't really anything quick about this call :-)
This is borderline bike-shedding but I would like something more clear like:
p11prov_store_direct_fetch() or maybe p11prov_store_internal() ...

simo5 (commented):
I think we are very close, I like the changes you made, and I really only found nits.

tests/tpem_encoder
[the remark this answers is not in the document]
0pq76r: od and printf are used instead of perl pack/unpack.
```
DESC_HEX=$(printf '%s' "${DESC}" | od -An -t x1)
...
# shellcheck disable=SC2059 # printf should treat variable as format string
printf "${PEM_HEX}" | base64
```

src/decoder.c
0pq76r: Re-implementation is gone. p11prov_store_direct_fetch() is exposed by the store.c to do a open+load.

src/Makefile.am
0pq76r: I moved the p11prov_store_direct_fetch declaration to store.h

src/store.c
0pq76r: I think it would be better to loop until p11prov_store_eof() is 1. Is there a clean way to accumulate the return values of store_load()?

src/decoder.c
0pq76r: I'm not sure if p11prov_alloc_sprintf is the right solution. The whole point of copying is to ensure it is null terminated because of the note in OSSL:

> In general it cannot be assumed that the data returned by ASN1_STRING_data() is *null terminated* or does not contain embedded nulls. 
-- https://www.openssl.org/docs/man1.1.1/man3/ASN1_STRING_get0_data.html

src/store.c
simo5: The only way, I think, is to create a linked list/array in the data structure you use to hold the callbacks.

0pq76r: I'm returning a STACK_OF(int)* containing the return value of the store_load function calls. I cannot put the return value into the callback data structure because the store_load does not know the structure and the callback function is only called if the store was able to load something.

simo5: Oh I misunderstood your question I guess.
So you were asking how to return the error from each call?
I do not think that is necessary, the reason to return the error, was to use
p11prov_set_error_mark(ctx->provctx); before the while, and after the while is over either call p11prov_pop_error_to_mark(ctx->provctx); if there was no error or p11prov_clear_last_error_mark(ctx->provctx); if an error happened.

src/decoder.c
simo5: p11prov_alloc_sprintf("%.*s", length, string);

simo5: The above takes care of adding a null terminator and will not overflow

src/store.c
0pq76r: I'm glad that I misunderstood you and that I don't have to return all the return values as list. However, I'm still a bit confused about how the error handling should actually work.

Is there any reason for the decoder to do anything with the value returned by store_load?
If an error occurs the callback function should not be called (this is already part of the store's behaviour). But the decoders return value should be True(1) to allow OpenSSL to use different decoders.

simo5: it is just to return debug logging to the caller which can then retrieve and paste it into their logs or discard as they wish.

0pq76r: If `open` or `load` fail `p11prov_store_direct_fetch()` returns an error code and leaves the error stack intact.

0pq76r:
I gave a build of this PR to some other devs. They found out that I forgot to add the encoder/decoder for Edwards curves. I added a commit to fix that.

src/decoder.c
```diff
@@ -189,6 +205,16 @@ static int p11prov_pem_decoder_p11prov_der_decode(
...
+
+        /* special case certificates */
+        struct desired_data_type_cbdata cbdata = {
+            .desired_data_type = P11PROV_NAME_CERTIFICATE,
+            .cb = object_cb,
+            .cbarg = object_cbarg,
+        };
+        load_obj(ctx, der_data, der_len, &cbdata, pw_cb, pw_cbarg);
+        /* */
+
```
simo5, line 185: Why do you need to handle certificates here?

simo5 (commented):
I think we have all that is needed here, I am just confused about the patch to handle certificates.
My understanding is that this feature exclusively deals with private key pem<->uri ?

simo5:
I guess the only way to make CI pass on debian, is to rebase on master, I have merged your PR for the t64 madness

0pq76r:
> I think we have all that is needed here, I am just confused about the patch to handle certificates. My understanding is that this feature exclusively deals with private key pem<->uri ?

The encoder is only implemented for private keys. In the decoder, the support for public keys comes for free. Explicitly filtering for private keys seems like an arbitrary restriction.

Unfortunately, the `ossl_store_handle_load_result` (https://github.com/openssl/openssl/blob/master/crypto/store/store_result.c#L85) only calls a second decoder for the keys (der->evp_pkey) . To have a similar decoding behaviour for the certificates as for the keys, the certificate needs to be returned while processing the pem file.

simo5:
are you thinking of a certificate stored in the token, and a uri only be present in the PEM file ?

So basically this means you can use PEM files for any object in the token ?

0pq76r:
Well, ideally the application would accept a pkcs11-uri for both key and certificate in the configuration.
Having a PEM file that supports any token object is more of a means to an end.

In the end, the goal is to only have to touch the token when doing a key/cert renewal. When an application only uses references to these objects the token can be used to distribute the new key/cert pairs.

src/decoder.c
```diff
@@ -189,6 +205,16 @@ static int p11prov_pem_decoder_p11prov_der_decode(
...
         && strcmp(pem_label, P11PROV_PEM_LABEL) == 0) {
+
+        /* special case certificates */
+        struct desired_data_type_cbdata cbdata = {
+            .desired_data_type = P11PROV_NAME_CERTIFICATE,
+            .cb = object_cb,
+            .cbarg = object_cbarg,
+        };
+        load_obj(ctx, der_data, der_len, &cbdata, pw_cb, pw_cbarg);
+        /* */
```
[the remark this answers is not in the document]
simo5: I guess I do not understand why you need this special case here.

The following code seem to work by letting openssl reinvoke the der decoder, why would that not work for certificates?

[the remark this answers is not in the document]
0pq76r: The empty comment is a left over from when I tried to get OSSL to decode the cert from the der.
For the keys the `load_obj()` is called in `try_key()` (frame 7). But, I cannot find an equivalent call for certificates.

```
#0  load_obj (ctx=ctx@entry=0x55555573cff0,
    der=0x5555556801f0 "0\202\001\004\032\031PKCS#11 Provider URI v1.0\f\201\346pkcs11:model=NSS%203;manufacturer=Mozilla%20Foundation;serial=", '0' <repeats 16 times>, ";token=NSS%20Certificate%20DB;id=%20%1E%EA%76%94%A0%EF%A5%A5%AB%9D%39%E1%9E%03%81;object"..., der_len=264, cbdata=cbdata@entry=0x7fffffffc220,
    pw_cb=pw_cb@entry=0x7ffff7c039b0 <ossl_pw_passphrase_callback_dec>, pw_cbarg=0x5555556fdc28) at decoder.c:92
#1  0x00007ffff77bb0c9 in p11prov_der_decoder_p11prov_obj_decode (desired_data_type=0x7ffff77e0171 "ED448", inctx=0x55555573cff0, cin=0x55555573e460, selection=<optimized out>,
    object_cb=0x7ffff7ba3cb0 <decoder_process>, object_cbarg=0x7fffffffc2c0, pw_cb=0x7ffff7c039b0 <ossl_pw_passphrase_callback_dec>, pw_cbarg=0x5555556fdc28) at decoder.c:144
#2  0x00007ffff77badac in p11prov_der_decoder_p11prov_ed448_decode (inctx=0x7fffffffc1d0, cin=0x108, selection=<optimized out>, object_cb=0x7fffffffc220,
    object_cbarg=0x7ffff7c039b0 <ossl_pw_passphrase_callback_dec>, pw_cb=0x5555556fdc28, pw_cbarg=0x5555556fdc28) at decoder.c:205
#3  0x00007ffff7ba40f1 in decoder_process (params=params@entry=0x0, arg=arg@entry=0x7fffffffc360) at crypto/encode_decode/decoder_lib.c:1000
#4  0x00007ffff7ba3b72 in OSSL_DECODER_from_bio (ctx=ctx@entry=0x5555556fdbf0, in=0x55555573e350) at crypto/encode_decode/decoder_lib.c:82
#5  0x00007ffff7ba42b9 in OSSL_DECODER_from_data (ctx=ctx@entry=0x5555556fdbf0, pdata=pdata@entry=0x7fffffffc480, pdata_len=pdata_len@entry=0x7fffffffc410)
    at crypto/encode_decode/decoder_lib.c:157
#6  0x00007ffff7d1c04e in try_key_value (data=<optimized out>, ctx=0x555555674aa0, cbarg=0x555555674ae8, libctx=0x7ffff7edb288 <default_context_int>, propq=0x0, cb=<optimized out>)
    at crypto/store/store_result.c:295
#7  try_key (data=<optimized out>, v=0x7fffffffccf8, ctx=0x555555674aa0, provider=<optimized out>, libctx=0x7ffff7edb288 <default_context_int>, propq=0x0)
    at crypto/store/store_result.c:412
#8  ossl_store_handle_load_result (params=<optimized out>, arg=0x7fffffffccf8) at crypto/store/store_result.c:134
#9  0x00007ffff7ba3d67 in decoder_process (params=params@entry=0x7fffffffca30, arg=arg@entry=0x7fffffffcb20) at crypto/encode_decode/decoder_lib.c:774
#10 0x00007ffff77bac2d in p11prov_pem_decoder_p11prov_der_decode (inctx=0x55555567e9c0, cin=<optimized out>, selection=<optimized out>, object_cb=0x7ffff7ba3cb0 <decoder_process>,
    object_cbarg=0x7fffffffcb20, pw_cb=0x7ffff7c039b0 <ossl_pw_passphrase_callback_dec>, pw_cbarg=0x555555666dc8) at decoder.c:190
#11 0x00007ffff7ba40f1 in decoder_process (params=params@entry=0x0, arg=arg@entry=0x7fffffffcbc0) at crypto/encode_decode/decoder_lib.c:1000
#12 0x00007ffff7ba3b72 in OSSL_DECODER_from_bio (ctx=0x555555666d90, in=0x555555674a10) at crypto/encode_decode/decoder_lib.c:82
#13 0x00007ffff7d9fd79 in file_load_file (ctx=0x555555669c50, object_cb=0x7ffff7d1b8c0 <ossl_store_handle_load_result>, object_cbarg=0x7fffffffccf8,
    pw_cb=0x7ffff7c039b0 <ossl_pw_passphrase_callback_dec>, pw_cbarg=0x555555674ae8) at providers/implementations/storemgmt/file_store.c:532
#14 file_load (loaderctx=0x555555669c50, object_cb=0x7ffff7d1b8c0 <ossl_store_handle_load_result>, object_cbarg=0x7fffffffccf8, pw_cb=0x7ffff7c039b0 <ossl_pw_passphrase_callback_dec>,
    pw_cbarg=0x555555674ae8) at providers/implementations/storemgmt/file_store.c:698
#15 0x00007ffff7d196d8 in OSSL_STORE_load (ctx=ctx@entry=0x555555674aa0) at crypto/store/store_lib.c:447
```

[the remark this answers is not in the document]
simo5: Certs are loaded in crypto/store/store_result.c:try_cert()

Cert should be returned only if expected_type is 0 (no type specified) or OSSL_STORE_INFO_CERT

simo5:
Note, the code currently only handles OSSL_STORE_INFO_PKEY in expect, so it will have to be adjusted, but we can do as a followup to this PR, I think we should merge this one once good enough, to avoid dragging it forever

simo5:
Sounds like Fedora address sanitizer is really mad, I kicked a rerun but see it it was not a fluke, unfortunately the logs are not very enlightening of where the problem lies ...

numinit:
How do I export a key for this? I've been trying something like this with my Yubikey:

`./result/bin/openssl pkeyutl -propquery provider=pkcs11 -in /dev/null -inkey 'pkcs11:serial=12345;object=Private key for Digital Signature' -out pk.pem`

But get:

```
Enter pass phrase for PKCS#11 Token (Slot 2 - Yubico YubiKey OTP+FIDO+CCID 02 00):
Enter PIN for PKCS#11 Token (Slot 2 - Yubico YubiKey OTP+FIDO+CCID 02 00):
Public Key operation error
40A790B9707F0000:error:030000CD:digital envelope routines:evp_pkey_copy_downgraded:keymgmt export failure:crypto/evp/p_lib.c:2080:key type = id-ecPublicKey
40A790B9707F0000:error:030000CD:digital envelope routines:evp_pkey_copy_downgraded:keymgmt export failure:crypto/evp/p_lib.c:2080:key type = id-ecPublicKey
Segmentation fault (core dumped)
```

numinit:
OK, I kind of got it working, but am still having trouble signing and get a segfault at the end...

```
./result/bin/openssl storeutl -keys -text 'pkcs11:serial=6460026;object-type=private'
Enter pass phrase for PKCS#11 Token (Slot 2 - Yubico YubiKey OTP+FIDO+CCID 02 00):
Enter PIN for PKCS#11 Token (Slot 2 - Yubico YubiKey OTP+FIDO+CCID 02 00):
0: Pkey
PKCS11 EC Private Key (384 bits)
[Can't export and print private key data]
URI pkcs11:model=YubiKey%20YK4;manufacturer=Yubico%20(www.yubico.com);serial=12345;token=YubiKey%20PIV%20%2312345;id=%02;object=Private%20key%20for%20Digital%20Signature;type=private
-----BEGIN PKCS#11 PROVIDER URI-----
(redacted)
-----END PKCS#11 PROVIDER URI-----
1: Pkey
PKCS11 RSA Private Key (2048 bits)
[Can't export and print private key data]
URI pkcs11:model=YubiKey%20YK4;manufacturer=Yubico%20(www.yubico.com);serial=12345;token=YubiKey%20PIV%20%2312345;id=%19;object=Private%20key%20for%20PIV%20Attestation;type=private
-----BEGIN PKCS#11 PROVIDER URI-----
(redacted)
-----END PKCS#11 PROVIDER URI-----
Total found: 2
Segmentation fault (core dumped)
```

numinit:
The no-deinit quirk was the solution, thanks.

To be clear: for some reason, it appears to be having trouble exporting the public key during sign(!) operations, even when I'm not using the pem exported key reference, which is strange:

```
./result/bin/openssl pkeyutl -sign -inkey provider.pem -keyform pem -rawin -digest sha256 -in default.nix -out default.sig
Enter pass phrase for PKCS#11 Token (Slot 2 - Yubico YubiKey OTP+FIDO+CCID 02 00):
Enter PIN for PKCS#11 Token (Slot 2 - Yubico YubiKey OTP+FIDO+CCID 02 00):
Public Key operation error
4047A0C25C7F0000:error:030000CD:digital envelope routines:evp_pkey_copy_downgraded:keymgmt export failure:crypto/evp/p_lib.c:2080:key type = id-ecPublicKey
4047A0C25C7F0000:error:030000CD:digital envelope routines:evp_pkey_copy_downgraded:keymgmt export failure:crypto/evp/p_lib.c:2080:key type = id-ecPublicKey
```

```
./result/bin/openssl pkeyutl -sign -inkey 'pkcs11:model=YubiKey%20YK4;manufacturer=Yubico%20(www.yubico.com);serial=12345;token=YubiKey%20PIV%20%2312345;id=%02;object=Private%20key%20for%20Digital%20Signature;type=private' -rawin -digest sha256 -in default.nix -out default.sig
Enter pass phrase for PKCS#11 Token (Slot 2 - Yubico YubiKey OTP+FIDO+CCID 02 00):
Enter PIN for PKCS#11 Token (Slot 2 - Yubico YubiKey OTP+FIDO+CCID 02 00):
Public Key operation error
4007B944FE7E0000:error:030000CD:digital envelope routines:evp_pkey_copy_downgraded:keymgmt export failure:crypto/evp/p_lib.c:2080:key type = id-ecPublicKey
4007B944FE7E0000:error:030000CD:digital envelope routines:evp_pkey_copy_downgraded:keymgmt export failure:crypto/evp/p_lib.c:2080:key type = id-ecPublicKey
```

Additionally, this occurs on the 0.3 branch, leaving me to believe it isn't related to this changeset. So I'll keep further commentary off this ticket.

numinit:
Got it working. Problem was I was doing shenanigans with a PKCS#11 engine, too. Confirmed (mostly) working:

```
$ ./result/bin/openssl pkeyutl -sign -inkey provider.pem -keyform pem -rawin -digest sha256 -in default.nix -out default.sig
Enter pass phrase for PKCS#11 Token (Slot 2 - Yubico YubiKey OTP+FIDO+CCID 02 00):
Enter PIN for PKCS#11 Token (Slot 2 - Yubico YubiKey OTP+FIDO+CCID 02 00):

$ hexdump -C default.sig
00000000  30 64 02 30 70 df 9f 72  f7 2a af f4 b9 f4 0b 64  |0d.0p..r.*.....d|
00000010  d4 89 c2 e0 f8 c2 de 97  bd b6 9e ef 5d 6c 46 9d  |............]lF.|
```

Then, in Ruby, which _only_ supports loading PEM files:

```
key = OpenSSL::PKey.read(File.read('provider.pem'))
Enter PEM pass phrase:
Enter PIN for PKCS#11 Token (Slot 2 - Yubico YubiKey OTP+FIDO+CCID 02 00):
=> #<OpenSSL::PKey::EC:0x00007f9949099fb8 oid=id-ecPublicKey>
irb(main):003> sig = key.sign 'SHA256', 'hello world'
(irb):3: [BUG] Segmentation fault at 0x00000000000000281
```

Oops, a ruby bug :-)

simo5:
Thanks for this contribution, looking forward to use it.

numinit:
This is awesome, well done.

numinit:
Node:

```
const crypto = require('node:crypto');
const fs = require('node:fs');
const privkey = crypto.createPrivateKey(fs.readFileSync('provider.pem').toString('ascii'))
// Enter PIN for PKCS#11 Token (Slot 2 - Yubico YubiKey OTP+FIDO+CCID 02 00):
const sign = crypto.createSign('SHA256'); sign.update('hello, node!'); sign.end();
const signature = sign.sign(privkey);
// <Buffer 30 65 02 31 00 99 8f e7 43 1a a4 90 4d cd 0c 15 bf 98 db 41 dd a3 a0 e5 5c 95 19 fe 4d 2d 33 d1 32 6a 5b 7b c4 05 f5 22 a1 ea 67 65 6a e0 0f ef 2f 13 ... 53 more bytes>
```

PR #333, opened 2024-01-09
Do not load profiles when they can't be available

simo5:
v2.40 (so anything be,low 3.0) spec did not have profiles.

Try to load them only for tokens that expose a 3+ version.


Jakuje (approved):
looks simple enough that it should not need separate test.

PR #334, opened 2024-01-17
tests: Check csr for keys in token with openssl

Jakuje:


simo5:
shellcheck complaint should be addressed otherwise it can be pushed.

Jakuje:
> shellcheck complaint should be addressed otherwise it can be pushed.

The report is weird. Its not on a line I would have touched and the `PRIURI` is used in several other places in the same file without any issue (though only in the `ossl` command which does expansion only later on.

I can try to workaround it though.

simo5:
Thanks, that worked.

PR #340, opened 2024-02-05
WIP: Run CI with integration test for libssh

The-Mule:
This is the first sketch of integration CI. For now it only runs libssh upstream tests relevant for pkcs11 provider. It only work on fedora and takes about 2 minutes to complete the job. The idea is to add jobs for other provider consumers. It requires additional polishing (e.g. ctest passes even if it runs no tests).

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,59 @@
+---
+name: Integration tests
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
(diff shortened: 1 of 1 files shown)


The-Mule:
PR should be final now. Expected ouput of ctest is now checked and we also check the the provider is used (using PKCS11_PROVIDER_DEBUG).

simo5:
Thanks @The-Mule

Issue #345, opened 2024-02-21
s_server fail: PEM cert with HSM private key fail as it looks for public key on HSM


space88man:
**Describe the bug**
I am using `openssl s_server` with `-cert APemFile -key "pkcs11:type=private;object=testCert"` .
The HSM does not have a public key or cert corresponding to the private key.

At `X509_check_private_key` it calls `p11prov_obj_key_cmp` which then tries to match a public key on the HSM which fails.

Is there a way to validate the private key with the public key from PEM cert instead of an (unsuccessful) round-trip to the HSM.

Started a discussion here: https://github.com/openssl/openssl/discussions/23652
**To Reproduce**
1. in tests/tmp.softhsm hide the ID of the public key:

        pkcs11-tool --module /usr/lib64/libsofthsm2.so --pin 12345678 -a testCert -y pubkey -e 0001FF

2.  export the PEM cert

        pkcs11-tool --module /usr/lib64/libsofthsm2.so --pin 12345678 -a testCert -y cert -r | openssl x509 -inform DER -out testCert.pem

3. Start s_server

        openssl s_server -cert testCert.pem -key 'pkcs11:type=private;object=testCert'

See error:

```
openssl s_server -cert testCert.pem -key 'pkcs11:type=private;object=testCert'
Using default temp DH parameters
error setting private key
007E33F06B7F0000:error:40800000:pkcs11:find_associated_obj:pkcs11:objects.c:1128:Error in C_FindObjects (count=0)
007E33F06B7F0000:error:40800005:pkcs11:match_public_keys:General Error:objects.c:2204:Could not find associated public key object
007E33F06B7F0000:error:05800074:x509 certificate routines:X509_check_private_key:key values mismatch:crypto/x509/x509_cmp.c:405:
```

**Expected behavior**

Can validate a HSM private key with a PEM cert in SSL_CTX_use_PrivateKey ... X509_check_private_key ...

**Operating environment (please complete the following information):**
 - OS: Fedora
 - Version: 39

**Token and application used (please complete the following information):**
- tests/tmp.softhsm from current master

**Additional context**
This worked in ye olde engine days with -keyform engine and -cert APemFile

```
Breakpoint 2, find_associated_obj (provctx=0x555555661820, obj=obj@entry=0x5555557127b0, class=class@entry=2) at objects.c:1078
1078        CK_ATTRIBUTE template[2] = { 0 };
(gdb) bt
[...]

PR #346, opened 2024-02-21
For RSA keys we can exit early after MODULUS / EXPONENT are verified.

space88man:
Addresses #345.

RSA keys: the  match operator was falling-through to the generic comparator (look for associated objects) instead of exiting early after EXPONENT/MODULUS were verified


simo5:
It is probably better to just add a check for the modulus early on for CKK_RSA just where we check the public exponent, and then we can immediately return w/o even calling cmp_public_keys.

Alternatively, move the check for CKA_PUBLIC_EXPONENT into cmp_public_key_values(), and the initial check for CKK_RSA would simply call cmp_public_key_values() unconditionally (with a comment that Private RSA key always include the public attributes.

And again just return later on.
I think I prefer the latter option, because currently cmp_public_key_values() is misleading as not checking public exponent in there is kind of bad.

space88man:
> And again just return later on. I think I prefer the latter option, because currently cmp_public_key_values() is misleading as not checking public exponent in there is kind of bad.

Ok - colocated both checks (MODULUS/PUBLIC_EXPONENT)

src/objects.c
```diff
@@ -2161,7 +2161,10 @@ static int cmp_public_key_values(P11PROV_OBJ *pub_key1, P11PROV_OBJ *pub_key2)
 
     switch (pub_key1->data.key.type) {
     case CKK_RSA:
-        ret = cmp_attr(pub_key1, pub_key2, CKA_MODULUS);
+        /* pub_key1 pub_key2 could be CKO_PRIVATE_KEY here but
+         *  nevertheless contain these two attributes
+         */
+        ret = cmp_attr(pub_key1, pub_key2, CKA_MODULUS) & cmp_attr(pub_key1, pub_key2, CKA_PUBLIC_EXPONENT);
```
simo5, line 2167: please split in two distinct lines with individual checks on the results.

src/objects.c
```diff
@@ -2180,8 +2183,12 @@ static int match_public_keys(P11PROV_OBJ *key1, P11PROV_OBJ *key2)
...
 
-    if (key1->class == CKO_PUBLIC_KEY && key2->class == CKO_PUBLIC_KEY) {
-        /* both keys are public, match directly their public values */
+    if ((key1->class == CKO_PUBLIC_KEY && key2->class == CKO_PUBLIC_KEY) || key1->data.key.type == CKK_RSA) {
+        /* either keys are public, match directly their public values
+         * OR
+         * CKA_RSA keys (private/public) contain CKA_MODULUS / CKA_PUBLIC_EXPONENT
+         * - no need to find_associated_obj
+         */
         return cmp_public_key_values(key1, key2);
```
simo5, line 2192: you probably want to urn make check-style-fix to automatically reformat according to our rules.

simo5 (changes requested):
Only a couple of style issues, then once CI is happy (ignore MacOS+softhsm if it fails, it if flaky) I'll merge.

space88man:
> Only a couple of style issues, then once CI is happy (ignore MacOS+softhsm if it fails, it if flaky) I'll merge.

Done - please check

src/objects.c
```diff
@@ -2161,7 +2161,12 @@ static int cmp_public_key_values(P11PROV_OBJ *pub_key1, P11PROV_OBJ *pub_key2)
 
     switch (pub_key1->data.key.type) {
     case CKK_RSA:
+        /* pub_key1 pub_key2 could be CKO_PRIVATE_KEY here but
+         *  nevertheless contain these two attributes
+         */
         ret = cmp_attr(pub_key1, pub_key2, CKA_MODULUS);
+        if (ret == RET_OSSL_ERR) break;
```
simo5, line 2168: checker will probably complain about this line :-)

simo5: The *other* checker :-D

simo5 (approved):
LGTM

PR #347, opened 2024-02-22
EC keys: implement vendor optimization when private key contains CKA_EC_POINT

space88man:
This is an optimization to avoid having a copy of the EC public key on HSM: some vendor implementations add CKA_EC_POINT to CKK_EC/CKO_PRIVATE_KEY to facilitate `EVP_PKEY_eq()`-type comparisons—skip round-tripping to the HSM to find_associated_obj.

This optimization is in OpenSC/libp11 see: : https://github.com/OpenSC/libp11/commit/281ccb3d60e77366bd6ece4b8a18d5ef28cb4f45

For both PR#346 PR#347 I found a discussion on libp11 about the round-tripping issue: https://github.com/OpenSC/libp11/issues/209

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -796,6 +796,12 @@ static CK_RV fetch_ec_key(P11PROV_CTX *ctx, P11PROV_SESSION *session,
     FA_SET_BUF_ALLOC(attrs, num, CKA_EC_PARAMS, true);
     if (key->class == CKO_PUBLIC_KEY) {
         FA_SET_BUF_ALLOC(attrs, num, CKA_EC_POINT, true);
+    } else {
+        /* known vendor optimization to avoid storing
+         * EC public key on HSM; can avoid
+         * find_associated_obj later
+         */
+        FA_SET_BUF_ALLOC(attrs, num, CKA_EC_POINT, false);
     }
     FA_SET_BUF_ALLOC(attrs, num, CKA_ID, false);
     FA_SET_BUF_ALLOC(attrs, num, CKA_LABEL, false);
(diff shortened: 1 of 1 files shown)


src/objects.c
```diff
@@ -2185,6 +2191,15 @@ static int match_public_keys(P11PROV_OBJ *key1, P11PROV_OBJ *key2)
...
 
+    if (key1->data.key.type == CKK_EC) {
+        /* Vendor implementation where EC private key may contain
+         * CKA_EC_POINT
+         */
+        ret = cmp_public_key_values(key1, key2);
+        if (ret != RET_OSSL_ERR)
+            return ret;
+    }
+
```
simo5, line 2202: Here we should probably just call cmp_public_key_values() unconditionally at the top of the function, if it succeed we return positively, if it doesn't we continue.
Returning an error is wrong because it will prevent tokens that work per spec *and* have an associated public key, from working.

simo5 (changes requested):
also please run the format checks via make before resubmitting

src/objects.c
space88man: Clarification - this returns on success - (!= RET_OSSL_ERR) - if there is an error then if will fall-through to the find_associated_obj. I.e. leave early on success, on failure continue with the original business logic.

I will wait to rebase after the previous PR is completed.

simo5: My bad, I read the check backwards, behavior is correct on error, the rest of the comment stands.
Ok to merge the other PR first.

space88man: Squashed with the comments incorporated

simo5:
Thanks, looks good, merging!

Issue #348, opened 2024-02-22
Is there a workaround for RHEL 9.3/4 OpenSSL 3.0.7 EC export compressed bug?


space88man:
**Describe the bug**
To be clear this is a bug in RHEL 9.3 OpenSSL 3.0.7 `ec_export` but cannot be fixed in time for 9.4.

I am wondering if there is a workaround that pkcs11-provider can do during keymgmt import.

* https://github.com/openssl/openssl/issues/16595
* https://issues.redhat.com/browse/RHEL-26325 — any fix cannot make it for 9.4
* https://github.com/openssl/openssl/commit/999509c235a59f0f65bc52522235db4c65297775

With OpenSSL 3.0.7 bug EC public key comparison will always fail: during export the  point-format says "uncompressed" but the actual data in the OSSL_PARAMS is compressed (about half the size). 
Thus `cmp_attr(key1, key2, CKA_P11PROV_PUB_KEY)` fails. I have observed that the data is always about half the size(compressed) compared to the uncompressed data from pkcs11-provider.

**To Reproduce**
1. Use OpenSSL 3.0.7 (e.g. from RHEL, compiled from upstream shows the same bug)
2. softhsm2 with an EC key pair
3. use s_server

**Expected behavior**
s_server loads

**Operating environment (please complete the following information):**
 - RHEL
 - 9.3 and soon 9.4

**Token and application used (please complete the following information):**
 - softhsm2 with EC keypair

**Additional context**

space88man:
Shifting the discussion to the PR https://github.com/latchset/pkcs11-provider/pull/349

PR #349, opened 2024-02-22
RFC: Workaround compressed EC points from OpenSSL 3.0.7

space88man:
Detect compressed point and import with the uncompressed point.

Use the heuristic starts-with '\x02'

Addresses #348


src/objects.c
```diff
@@ -2494,11 +2494,19 @@ static CK_RV prep_rsa_find(P11PROV_CTX *ctx, const OSSL_PARAM params[],
...
                           struct pool_find_ctx *findctx)
 {
     EC_GROUP *group = NULL;
+    EC_POINT *point = NULL;
+    BN_CTX *bn_ctx = NULL;
+    int ret, plen;
+
     OSSL_PARAM tmp;
+    const OSSL_PARAM *p;
+    OSSL_PARAM pub_key[2] = { { NULL, 0, NULL, 0, 0 },
```
simo5, line 2508: OSSL_PARAM pub_key[2] = { 0 }; should suffice.

src/objects.c
```diff
@@ -2494,11 +2494,19 @@ static CK_RV prep_rsa_find(P11PROV_CTX *ctx, const OSSL_PARAM params[],
     return CKR_OK;
 }
 
+#define PUB_KEY_SIZE 200
```
simo5, line 2497: This looks odd, is this max length for all curves?
If so maybe rename to MAX_EC_PUB_KEY_SIZE

src/objects.c
```diff
@@ -2510,20 +2518,60 @@ static CK_RV prep_ec_find(P11PROV_CTX *ctx, const OSSL_PARAM params[],
...
+                      OSSL_PKEY_PARAM_PUB_KEY);
+        EC_GROUP_free(group);
+        return CKR_KEY_INDIGESTIBLE;
+    }
+
+    if (((char *)p->data)[0] == '\x02') {
+        P11PROV_debug("OpenSSL 3.0.7 BUG - received compressed EC public key");
+        pub_key[0].key = OSSL_PKEY_PARAM_PUB_KEY;
+        pub_key[0].data_type = p->data_type;
+        pub_key[0].data = OPENSSL_malloc(PUB_KEY_SIZE);
```
simo5, line 2540: Instead of using  fixed value, just call EC_POINT_point2oct() with a NULL buffer, and it will give you the needed lenght, so then we can malloc just the necessary data.

Alternatively if 200 bytes is always enough space, just allocate the space on the stack via something like `unit8_t pub_data[MAX_EC_PUB_KEY_SIZE];`

simo5 (changes requested):
Pretty good, just minor tweaks needed.

simo5 (approved):

simo5:
Thanks for this one too, exactly how I liked it to be built.
Merging.

PR #353, opened 2024-03-04
GH workflow install libssl3t64-dgbsym on debian

0pq76r:
Debian is getting ready for 2038 (https://lwn.net/Articles/938149/). Apparently they have converted the libssl package. It is now necessary to install the *t64 version of the debug symbols.


simo5:
smh ...

PR #356, opened 2024-03-07
Fallback to a read lock on fork preparation

### LINKED ISSUE #355: Deadlock on P11PROV_SLOTS_CTX->rwlock while in passphrase callback
I have encountered a deadlock on the rwlock of a P11PROV_SLOTS_CTX.

This is an Apache server using the PKCS#11 provider, and no PIN specified in the URI nor anywhere else. Thus it calls the passphrase callback to prompt for the PIN.

It deadlocks inside the passphrase callback because apache seems to perform a fork while in the passphrase callback. `p11prov_slot_fork_prepare()` tries to get a write lock in the slot context's rwlock. But this is already read locked, which was taken by `p11prov_take_slots()` which was called from `p11prov_get_session()` which itself was called in `store_fetch()`. 

```
__futex_abstimed_wait_common (futex_word=futex_word@entry=0x2aa00105d60, expected=expected@entry=2, clockid=clockid@entry=0,
    abstime=abstime@entry=0x0, private=private@entry=0, cancel=false) at futex-internal.c:103
103     futex-internal.c: No such file or directory.
(gdb) bt
#0  __futex_abstimed_wait_common (futex_word=futex_word@entry=0x2aa00105d60, expected=expected@entry=2, clockid=clockid@entry=0,
    abstime=abstime@entry=0x0, private=private@entry=0, cancel=false) at futex-internal.c:103
#1  0x000003fff7a98918 in __GI___futex_abstimed_wait64 (futex_word=futex_word@entry=0x2aa00105d60, expected=expected@entry=2,
    clockid=clockid@entry=0, abstime=abstime@entry=0x0, private=private@entry=0) at futex-internal.c:128
#2  0x000003fff7aa2964 in __pthread_rwlock_wrlock_full64 (abstime=0x0, clockid=0, rwlock=0x2aa00105d58) at pthread_rwlock_common.c:829
#3  ___pthread_rwlock_wrlock (rwlock=0x2aa00105d58) at pthread_rwlock_wrlock.c:26
#4  0x000003fff6db0396 in p11prov_slot_fork_prepare (sctx=<optimized out>) at slot.c:243
#5  0x000003fff6da362e in fork_prepare () at provider.c:80
#6  0x000003fff7b09290 in __run_prefork_handlers (do_locking=do_locking@entry=true) at register-atfork.c:141
#7  0x000003fff7aeb93e in __libc_fork () at fork.c:51
#8  0x000003fff7d31a76 in apr_proc_create () from /lib/s390x-linux-gnu/libapr-1.so.0
#9  0x000003fff6fac440 in ?? () from /usr/lib/apache2/modules/mod_ssl.so
#10 0x000003fff6faf83a in ?? () from /usr/lib/apache2/modules/mod_ssl.so
#11 0x000003fff6fa26ce in ?? () from /usr/lib/apache2/modules/mod_ssl.so
#12 0x000003fff6b2bda2 in UI_process () from /lib/s390x-linux-gnu/libcrypto.so.3
#13 0x000003fff6acee56 in ?? () from /lib/s390x-linux-gnu/libcrypto.so.3
#14 0x000003fff6acf37c in ?? () from /lib/s390x-linux-gnu/libcrypto.so.3
#15 0x000003fff6da972c in token_login (session=session@entry=0x2aa0014f5d0, uri=uri@entry=0x2aa000f13d0, pw_cb=0x3fff6acf360,
    pw_cb@entry=<error reading variable: value has been optimized out>, pw_cbarg=0x2aa0014f598,
    pw_cbarg@entry=<error reading variable: value has been optimized out>, slot=slot@entry=0x2aa001064d0, user_type=1)
    at session.c:475
#16 0x000003fff6daafa0 in slot_login (slot=<optimized out>, slot@entry=0x2aa001064d0, uri=uri@entry=0x2aa000f13d0,
    pw_cb=<optimized out>, pw_cb@entry=0x3fff6acf360, pw_cbarg=<optimized out>, pw_cbarg@entry=0x2aa0014f598,
    reqlogin=reqlogin@entry=true, _session=0x0) at session.c:781
#17 0x000003fff6dab4fe in p11prov_get_session (provctx=0x2aa000f6860, slotid=slotid@entry=0x3ffffff96c0,
    next_slotid=next_slotid@entry=0x3ffffff96c8, uri=0x2aa000f13d0, mechtype=mechtype@entry=18446744073709551615,
    pw_cb=0x3fff6acf360, pw_cbarg=0x2aa0014f598, reqlogin=true, rw=false, _session=0x2aa0014f4a8) at session.c:915
#18 0x000003fff6db21a8 in store_fetch (pw_cbarg=0x2aa0014f598, pw_cb=0x3fff6acf360, ctx=0x2aa0014f420) at store.c:122
#19 p11prov_store_load (pctx=0x2aa0014f420, object_cb=0x3fff6b20430, object_cbarg=<optimized out>, pw_cb=0x3fff6acf360,
    pw_cbarg=0x2aa0014f598) at store.c:253
#20 0x000003fff6b1ebf8 in OSSL_STORE_load () from /lib/s390x-linux-gnu/libcrypto.so.3
#21 0x000003fff6fab692 in ?? () from /usr/lib/apache2/modules/mod_ssl.so
#22 0x000003fff6f96988 in ?? () from /usr/lib/apache2/modules/mod_ssl.so
#23 0x000003fff6f999c2 in ?? () from /usr/lib/apache2/modules/mod_ssl.so
#24 0x000002aa00062590 in ap_run_post_config ()
#25 0x000002aa0003b32e in main ()
```

I don't know if there is anything you could do about this, or if its rather a bug or misbehavior of apache. I don't know why it does perform a fork while in its passphrase callback.... This does not sound OK to me...

Anyway, would it be possible to make sure to not hold the lock while calling the passphrase callback ?
You never know how the callback function behaves, so not holding any locks would probably be safer.

simo5:
The reason to acquire a write lock is that there is a corner case where
    the forking application is going to actually create additional threads
    in an atfork() handler that runs before ours instead of doing something
    more sensible like an immediate exec().
    If such an application were to create threads then there is a chance
    one of those threads could cause a call into pkcs11-provider that would
    end up acquiring a read lock, which we will destroy as part of the post
    fork atfork() handler.
    Note that pthread creation or even pthread lock initialization functions
    are not considered async-sginal-safe fucntions, so this is all a shot in
    the dark anyway ...
    
    Given this premise, falling back to a read lock, which will prevent
    modification of the slots is sufficient in most cases.
    
    Although there is a slight chance again that resetting the locks can
    cause a thread to obtain a write lock while another thread holds a now
    defunct read lock, such that there is modification of the slots while
    a separate thread is reading this information, this is so rare it is
    not worth dealing with (and there is nothing we can do anyway).
    At least two threads won't be writing at the same time, that could be
    really problematic.
    
    Use of atfork() is also so rare that the chance of this happening in any
    reasonable application can be considered infinitesimally small.

- [x] test fork deadlock

Fixes #355

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ tests/*.log
 tests/*.trs
 tests/pincache
 tests/tfork
+tests/tfork_deadlock
 tests/tlsctx
 tests/tcmpkeys
 
--- a/src/encoder.c
+++ b/src/encoder.c
@@ -470,37 +470,38 @@ static P11PROV_PK11_URI *p11prov_encoder_private_key_to_asn1(P11PROV_CTX *pctx,
                                                              P11PROV_OBJ *key)
 {
     P11PROV_PK11_URI *out = NULL;
+    char *uri = NULL;
+    size_t uri_len;
+    int ret = RET_OSSL_ERR;
 
-    char *uri = p11prov_key_to_uri(pctx, key);
+    uri = p11prov_key_to_uri(pctx, key);
     if (!uri) {
-        goto error;
+        goto done;
     }
 
-    size_t uri_len = strlen(uri);
+    uri_len = strlen(uri);
     P11PROV_debug("uri=%s", uri);
 
     out = P11PROV_PK11_URI_new();
     if (!out) {
-        goto error;
+        goto done;
     }
 
     if (!ASN1_STRING_set(out->desc, P11PROV_DESCS_URI_FILE,
                          sizeof(P11PROV_DESCS_URI_FILE) - 1)) {
-        goto error;
+        goto done;
     }
     if (!ASN1_STRING_set(out->uri, uri, uri_len)) {
-        goto error;
+        goto done;
     }
 
-    goto done;
-
-error:
-    P11PROV_PK11_URI_free(out);
-    out = NULL;
+    ret = RET_OSSL_OK;
 
 done:
-    if (uri) {
-        free(uri);
+    OPENSSL_free(uri);
+    if (ret != RET_OSSL_OK) {
+        P11PROV_PK11_URI_free(out);
+        out = NULL;
     }
     return out;
 }
--- a/src/slot.c
+++ b/src/slot.c
@@ -246,11 +246,27 @@ void p11prov_slot_fork_prepare(P11PROV_SLOTS_CTX *sctx)
 {
     int err;
 
-    err = pthread_rwlock_wrlock(&sctx->rwlock);
+    /* attempt to get a write lock if possible, but fall back to a mere
+     * read lock if not possible (for example because it would cause
+     * a deadlock).
+     * Holding a write lock here is slightly preferable in case the
+     * application decides to create threads immediately after the fork
+     * within an atfork handler that runs before ours.
+     * Holding a write lock will prevent other threads from grabbing a
+     * read lock before we can reset the locks. However we assume this
+     * scenario to be mostly hypothetical and exceedingly rare given most
+     * forks result in a exec(), and atfork() is also a rarely used
+     * function, so falling back to a read lock to avoid deadlocks is ok
+     * in the vast majority of use cases.
+     */
+    err = pthread_rwlock_trywrlock(&sctx->rwlock);
     if (err != 0) {
-        err = errno;
-        P11PROV_debug("Failed to get slots lock (errno:%d)", err);
-        return;
+        err = pthread_rwlock_rdlock(&sctx->rwlock);
+        if (err != 0) {
+            err = errno;
+            P11PROV_debug("Failed to get slots lock (errno:%d)", err);
+            return;
+        }
     }
 }
 
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -10,7 +10,8 @@ testssrcdir=@abs_srcdir@
 #VALGRIND_SUPPRESSIONS_FILES = $(top_srcdir)/tests/pkcs11-provider.supp
 VALGRIND_FLAGS = --num-callers=30 -q --keep-debuginfo=yes
 
-check_PROGRAMS = tsession tgenkey tlsctx tdigests tdigest_dupctx treadkeys tcmpkeys tfork pincache
+check_PROGRAMS = tsession tgenkey tlsctx tdigests tdigest_dupctx treadkeys \
+		 tcmpkeys tfork pincache tfork_deadlock
 
 tsession_SOURCES = tsession.c
 tsession_CFLAGS = $(STD_CFLAGS) $(OPENSSL_CFLAGS)
@@ -48,6 +49,10 @@ pincache_SOURCES = pincache.c
 pincache_CFLAGS = $(STD_CFLAGS) $(OPENSSL_CFLAGS)
 pincache_LDADD = $(OPENSSL_LIBS)
 
+tfork_deadlock_SOURCES = tfork_deadlock.c
+tfork_deadlock_CFLAGS = $(STD_CFLAGS) $(OPENSSL_CFLAGS)
+tfork_deadlock_LDADD = $(OPENSSL_LIBS)
+
 tmp.softokn:
 	LIBSPATH=$(libspath) \
 	TESTSSRCDIR=$(testssrcdir) \
@@ -66,7 +71,7 @@ tmp.softhsm:
 dist_check_SCRIPTS = \
 	helpers.sh setup-softhsm.sh setup-softokn.sh softhsm-proxy.sh \
 	test-wrapper tbasic tcerts tecc tecdh tedwards tdemoca thkdf \
-	toaepsha2 trsapss tdigest ttls tpubkey tfork turi trand tecxc \
+	toaepsha2 trsapss tdigest ttls tpubkey tforking turi trand tecxc \
 	tcms top_state tpem_encoder
 
 test_LIST = \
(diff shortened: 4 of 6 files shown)


simo5:
Actually this patch may still not be ok, because the read lock could be nested in the same thread as the code calling the callback and already holding a read lock ...

ifranzki:
I have tested this change with Apache and yes, it does avoid the deadlock.

Jakuje (approved):
looks good

ifranzki:
Small typo in commit title:
"Just acquire a read lock on for preparation"
should probably read "... on for**k** preparation"

simo5:
@ifranzki does apache fork from the thread where the callback is running, or does it fork from another thread?

simo5:
Ok so I verified that holding a read lock multiple times with pthread_rwlock_rdlock() is just fine, therefore the code will work as is.

However I think I may still change this patch to attempt to acquire a write lock, and fallback to a mere read lock only if that fails ...

simo5:
I added a trywrlock() and also added a much deeper explanation in the commit message of my reasoning why I do what I do now in the code.
Will modify the PR description accordingly as well.

ifranzki:
> does apache fork from the thread where the callback is running, or does it fork from another thread?

It does fork from the thread where the callback is running. You can see this in the backtrace in the issue:
`#8  0x000003fff7d31a76 in apr_proc_create () from /lib/s390x-linux-gnu/libapr-1.so.0`
This is doing the fork, it is a function of the Apache portable runtime (APR), here is the code (look for `apr_proc_create`):
https://svn.apache.org/viewvc/apr/apr/trunk/threadproc/unix/proc.c?view=markup

Also, the backtrace shows that its executing in the process's main thread, i.e. the backtrack shows `main()` as very first function in the backtrace.

simo5:
Thanks for confirming, I think I am ok with the current code.
I just nee dot find time to write a test so we do not regress.

Jakuje:
>  Address sanitizer / CI with Address Sanitizer (fedora) (pull_request) Failing after 360m 

I ran diff through the logs for fedora in this run and last working in main https://github.com/latchset/pkcs11-provider/actions/runs/8266635712/job/22615078760 and the only difference in the dependencies is `unbound` from `1.19.1-2` to `1.19.1-4`.

Running the address sanitizer checks locally on master shows no issues, while running it in this branch shows two memory leaks:
```
Direct leak of 88 byte(s) in 1 object(s) allocated from:
    #0 0x7f742b2d92ef in malloc (/usr/lib64/libasan.so.8.0.0+0xd92ef) (BuildId: 7fcb7759bc17ef47f9682414b6d99732d6a6ab0c)
    #1 0x7f742adb5334 in CRYPTO_zalloc (/lib64/libcrypto.so.3+0x1b5334) (BuildId: 5d012dcc6f62f35dabb8a129d641e3df7b731f11)
    #2 0x7f742ae172dc in UI_create_method (/lib64/libcrypto.so.3+0x2172dc) (BuildId: 5d012dcc6f62f35dabb8a129d641e3df7b731f11)
    #3 0x4023f7 in main /home/jjelen/devel/pkcs11-provider/tests/tfork_deadlock.c:127
    #4 0x7f742aa46149 in __libc_start_call_main (/lib64/libc.so.6+0x28149) (BuildId: 7ea8d85df0e89b90c63ac7ed2b3578b2e7728756)
    #5 0x7f742aa4620a in __libc_start_main_impl (/lib64/libc.so.6+0x2820a) (BuildId: 7ea8d85df0e89b90c63ac7ed2b3578b2e7728756)
    #6 0x402cd4 in _start (/home/jjelen/devel/pkcs11-provider/tests/tfork_deadlock+0x402cd4) (BuildId: 4227a9452263f96357af137df2d8392b07b00ea6)

Indirect leak of 14 byte(s) in 1 object(s) allocated from:
    #0 0x7f742b2d92ef in malloc (/usr/lib64/libasan.so.8.0.0+0xd92ef) (BuildId: 7fcb7759bc17ef47f9682414b6d99732d6a6ab0c)
    #1 0x7f742adb6672 in CRYPTO_strdup (/lib64/libcrypto.so.3+0x1b6672) (BuildId: 5d012dcc6f62f35dabb8a129d641e3df7b731f11)
    #2 0x7f742ae172f4 in UI_create_method (/lib64/libcrypto.so.3+0x2172f4) (BuildId: 5d012dcc6f62f35dabb8a129d641e3df7b731f11)
    #3 0x4023f7 in main /home/jjelen/devel/pkcs11-provider/tests/tfork_deadlock.c:127
    #4 0x7f742aa46149 in __libc_start_call_main (/lib64/libc.so.6+0x28149) (BuildId: 7ea8d85df0e89b90c63ac7ed2b3578b2e7728756)
    #5 0x7f742aa4620a in __libc_start_main_impl (/lib64/libc.so.6+0x2820a) (BuildId: 7ea8d85df0e89b90c63ac7ed2b3578b2e7728756)
    #6 0x402cd4 in _start (/home/jjelen/devel/pkcs11-provider/tests/tfork_deadlock+0x402cd4) (BuildId: 4227a9452263f96357af137df2d8392b07b00ea6)
```
Can you try to fix them and rebuild it?

simo5:
@Jakuje I sneaked in a commit to fix a Coverity regression in another PR I merged today.
However the Fedora Sanitizer seem randomly failing again.
Are you ok to push regardless ?

Jakuje (approved):

simo5:
Merging as the address sanitizer issue is a problem in the Fedora build not in the code.

PR #358, opened 2024-03-12
Integration tests (libssh, httpd, bind)

The-Mule:
This PR adds two new tests (for httpd and bind) and re-factors previously added test for libssh. Tests are intended to run in CI (github actions) but they can be executed on any Fedora (39 and newer) as follows (`GIT_URL`  defaults to this repository and `GIT_REF` to the main branch):

```
# cd tests/integration
# GIT_URL=https://github.com/The-Mule/pkcs11-provider.git GIT_REF=integration-tests bash -e libssh.sh
```

Testing requires bash -e option. Non-zero exit code means failure. Provider debug logs are always printed out at the end (regardless of the result). All three tests uses softhsm token (bind test does so via p11-kit server proxy). Please notice that until Fedora 40 is released both httpd and bind test need to use a specific docker image.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -7,53 +7,80 @@ on:
   pull_request:
     branches: ["main"]
 
-env:
-    P_LIBSSH_DEBUG: /tmp/libssh-provider-debug.log
-
 jobs:
-  build:
-    name: libssh
+  libssh:
     runs-on: ubuntu-22.04
-    container: fedora:latest
+    container: docker.io/dokken/fedora-40
     steps:
-      - name: Install Dependencies
-        run: |
-          dnf -y install gcc g++ git cmake libcmocka libcmocka-devel \
-          autoconf automake autoconf-archive libtool softhsm nss-tools \
-          gnutls-utils p11-kit p11-kit-devel p11-kit-server opensc \
-          softhsm-devel socket_wrapper nss_wrapper uid_wrapper \
-          pam_wrapper priv_wrapper openssh-server zlib-devel
       - name: Checkout Repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
+
+      - name: Install Build Requirements
+        run: |
+          dnf -y install git autoconf autoconf-archive automake libtool \
+            openssl-devel
+
       - name: Setup, Build and Install pkcs11-provider
         run: |
           autoreconf -fiv
           ./configure --libdir=/usr/lib64
           make
           make install
-      - name: Clone, Setup and Build libssh
+
+      - name: Test
         run: |
-          git clone https://gitlab.com/libssh/libssh-mirror.git
-          cd libssh-mirror/
-          mkdir build
-          cd build/
-          cmake \
-            -DUNIT_TESTING=ON \
-            -DCLIENT_TESTING=ON \
-            -DCMAKE_BUILD_TYPE=Debug \
-            -DWITH_PKCS11_URI=ON \
-            -DWITH_PKCS11_PROVIDER=ON \
-            -DPKCS11_PROVIDER=/usr/lib64/ossl-modules/pkcs11.so ..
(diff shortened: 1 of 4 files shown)


The-Mule:
I need to process shellcheck errors. 

New integration tests results can be found in the checks too (all passes).

.github/workflows/integration.yml
```diff
@@ -7,53 +7,80 @@ on:
...
-env:
-    P_LIBSSH_DEBUG: /tmp/libssh-provider-debug.log
-
 jobs:
-  build:
-    name: libssh
+  libssh:
     runs-on: ubuntu-22.04
-    container: fedora:latest
+    container: docker.io/dokken/fedora-40
```
simo5, line 13: can we rather use fedora:rawhide ?
Otherwise this will have to be constantly maintained as time apsses

tests/integration/libssh.sh
```diff
@@ -0,0 +1,105 @@
...
+    title PARA "Install dependencies" 
+    
+    dnf install -y --skip-broken cmake libcmocka libcmocka-devel softhsm \
+      nss-tools gnutls-utils p11-kit p11-kit-devel p11-kit-server opensc \
+      softhsm-devel socket_wrapper nss_wrapper uid_wrapper pam_wrapper \
+      priv_wrapper openssh-server zlib-devel git autoconf autoconf-archive \
+      automake libtool openssl-devel gcc g++ libcmocka-devel 
+}
+
+pkcs11_provider_setup()
```
simo5, line 23: Two comments here:
- why do you re-checkout and build pkcs11-provider when the git hub action can do that  (in a cached way) for you already ?

- a few of these steps seem to be repeated in the various scripts, wouldn't ii make sense to have a common.sh library imported in the various scripts and execute from there so that we reduce the maintenance cost later ?

simo5 (commented):
Mostly LGTM, but I am afarid of the time it may take to run this is there are a lot of steps that pull lots of data from the network.

simo5:
So the integration step took 6 minutes.
4.42 minutes were spent in downloading/installing stuff.
We should probably use caches aggressively.

See this for inspiration: 
https://github.com/latchset/kryoptic/blob/15bd875f6ce5acbaa93c909074f1da2812215a68/.github/workflows/build.yml#L27

tests/integration/libssh.sh
The-Mule: This pkcs11_provider_setup() is skipped in GH Actions, it is only intended to run when the test is executed manually. When it fails in CI it is useful for debugging to be able to run the test locally, that is now possible. Some code de-duplication might be useful though.

.github/workflows/integration.yml
The-Mule: Once F40 is released we can use fedora:latest everywhere. I can try to use fedora:rawhide. I think I picked dokken/fedora-40 for some specific reason I resolved later anyway. Let me check...

simo5:
> I'll take a look how to make it faster, a lot of time is spent on installing F40 packages now. But maybe that can be cached too somehow.

Se previous comment, you just need to copy that code :-)

.github/workflows/integration.yml
The-Mule: Alright, not it works fine with fedora:rawhide.

tests/integration/libssh.sh
The-Mule: I simplified the workflow code a lot with matrix strategy.

The-Mule:
With matrix strategy integration tests workflow only takes ~2 minutes (1:57) now. I tried to add actions/cache on top of that but it had no effect in this case (it looks like dnf step is now running at the same time in all three jobs).

simo5:
Awesome thanks,
I think we should still cache because some tests are still dominated by install time, and it is all in download time. But we can do that in a followup PR.

PR #366, opened 2024-03-25
Try to run Coverity Scan on demand

simo5:
#### Description

PR to change Coverity Scan to run on demand (by setting the covscan label) instead of running it after merge to main.
This way we can run the scan when the PR is still open and catch issues before merging.

#### Checklist

N/A

#### Reviewer's checklist:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -26,3 +26,4 @@ want to know about year later when sorting thorough as commits are the changelog
 - [ ] This feature/change has adequate documentation added
 - [ ] Code conform to coding style that today cannot yet be enforced via the check style test
 - [ ] Commits have short titles and sensible commit messages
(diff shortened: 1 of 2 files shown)


simo5:
The conditional running of covscan works, as well as the auto-removal of labels (which requires pull-request write permissions).

This all *should* work with pull_request_target events as well, which is an event that runs the workflows *from* the repository, instead of possibly changed workflows from the PR from a fork or separate branch.

By running the workflow stored on the main branch, rather the one on the PR github should grant access to secretes needed to upload coverity data for analysis and allow writes to PR for removing the label automatically, but of course this can't be tested until the PR is merged.

simo5:
From covscan:

```
    Build ID: 600979

    Analysis Summary:
       New defects found: 0
       Defects eliminated: 0
```

simo5:
Merging w/o running another covscan because covscan is rate limited so we need to avoid spurious runs.

PR #369, opened 2024-04-05
ci: Run tests also against CentOS 9 (OpenSSL 3.0)

Jakuje:
#### Description

The Debian is unstable so it is moving quite fast with the OpenSSL versions. Fedora will update from OpenSSL 3.1 to 3.2 in coming month so we will not be able to check the changes keep working with older OpenSSL versions, which is our interest for RHEL9.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        name: [fedora, debian]
+        name: [fedora, debian, centos]
         compiler: [gcc, clang]
         token: [softokn, softhsm]
         include:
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -14,19 +14,21 @@ jobs:
...
           - name: debian
             container: debian:sid
+          - name: centos
+            container: quay.io/centos/centos:stream9
     container: ${{ matrix.container }}
     steps:
       - name: Install Dependencies
         run: |
-            if [ -f /etc/fedora-release ]; then
+            if [ -f /etc/redhat-release ]; then
```
simo5, line 31: we should probably use /etc/system-release here?

simo5 (approved):
LGTM, let me know if you want to change the code to use the system-release file, I do not think it is a deal breaker, it works as is as well.

.github/workflows/build.yml
Jakuje: I do not have preference. If you believe the `system-release` is more appropriate, I can change it. I see its not in debian containers. But I thought the `redhat-release` was more self-documented.

simo5: if debian does not have it no real need then.

PR #376, opened 2024-04-11
Set correct pin-value format in tests

### LINKED ISSUE #360: the pkcs11-provider takes pin-value query option from the PKCS#11 URI path
>  nc ERROR: fail: OSSL_STORE_INFO_PKEY lookup [uri=pkcs11:token=token_client;object=my_key;pin-value=1234]

The PKCS#11 URI says the `pin-value` is a part of `pk11-query` so it should come after `?` instead of the `;`. But I see this used in the tests in pkcs11-provider everywhere so its likely something we should be more strict, but likely not an issue in your test.

https://www.rfc-editor.org/rfc/rfc7512.html#section-2.3

_Originally posted by @Jakuje in https://github.com/latchset/pkcs11-provider/issues/357#issuecomment-1999428574_

I think we should fix at least tests to use valid PKCS#11 URIs and maybe fix the parser to be more strict to distinguish the path and query part of the URI.

sarroutbi:
#### Description
As described in #360, correct format for pin-value option in URI is for it to be after '?', rather than as an additional URI path. This change sets the values appropriately in tests to identify possible parser issues

Fixes #360

#### Checklist

- [x] Test suite updated for functionality tests

#### Reviewer's checklist:

### DIFF
--- a/tests/setup-softhsm.sh
+++ b/tests/setup-softhsm.sh
@@ -179,7 +179,7 @@ pkcs11-tool --keypairgen --key-type="RSA:2048" --login --pin=$PINVALUE \
 	--module="$P11LIB" --label="${TSTCRTN}" --id="$KEYID"
 ca_sign $TSTCRT $TSTCRTN "My Test Cert" $KEYID
 
-BASEURIWITHPIN="pkcs11:id=${URIKEYID};pin-value=${PINVALUE}"
+BASEURIWITHPIN="pkcs11:id=${URIKEYID}?pin-value=${PINVALUE}"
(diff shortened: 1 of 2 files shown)


simo5:
Thanks, looks good!

Jakuje:
Should we also update the parser to not accept the "bad" values or at least warn the user he did something "wrong"?

simo5:
To be clear,
I would not reject a PR to emit a warning in the debug/openssl log for pin-value and/or path values in the incorrect place, but it is not a priority.

PR #377, opened 2024-04-11
ci: Switch to macOS 14 on M series chips

neverpanic:
#### Description
macOS 12 was released in October 21, and since then 13 has been out in October 22, and 14 in September 23. Switch to a more recent Mac operating system since that's likely what users will be running.

See [\[1\]][1] for GitHub's list of available shared runners. From the macos-12 label to the macos-14 label, the number of CPU cores stays the same (3), but the CPU architecture switches from x86_64 to arm64. All other parameters stay the same except for RAM, which changes from 14 GB to 7 GB. Since pkcs11-provider is not a memory intensive project, this likely has negligible impact on us. The new CPU architecture can be helpful to identify hardware-specific issues, so is probably a net positive. The M series chips are also typically much faster than the older Intel counterparts.

[1]: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

#### Checklist

- [ ] ~Documentation updated~

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,7 +89,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-12]
+        os: [macos-14]
         token: [softokn, softhsm]
     steps:
       - name: Install Dependencies


neverpanic:
This is also nice because the macOS-14 run is now actually the first CI step to finish :)

Jakuje (approved):
lgtm.

Does it make sense to run also on the older one for some compatibility?

neverpanic:
> Does it make sense to run also on the older one for some compatibility?

IIRC Simo only wanted one.

I also think we don't really need to continue supporting older macOS releases for a longer time, plus there's really nothing in pkcs11-provider that would behave in a different way on an older macOS. I don't think there's a lot of added value by testing older macOS.

PR #378, opened 2024-04-12
Use a genereic mechanism to block calls to tokens

### LINKED ISSUE #327: Extend operation state quirk to cover all operation state operations.
PR #324 introduced the ability to set a configuration quirk so that trying to save the state is not even attempted.

This should be extended to all uses of the operation state function.
Ideally by checking for the quirk within the body of the operation state wrapper itself.

simo5:
#### Description

Introduce a generic mechanism to block calls to a token.
Use this mechanism to reimplement the no_operation_state quirk.

Fixes #327 

#### Checklist

- [ ] ~Documentation updated~

#### Reviewer's checklist:

### DIFF
--- a/src/interface.h
+++ b/src/interface.h
@@ -135,4 +135,63 @@ CK_RV side_channel_free_Decrypt(P11PROV_CTX *ctx, CK_SESSION_HANDLE hSession,
 
 CK_INFO p11prov_module_ck_info(P11PROV_MODULE *mctx);
 
+/* The following defines are needed for a generic mask for any of the functions
+ * we generate via interface.pre, however there is no need to assign a blocking
+ * value until we'll have a configuration option that allows to set blocks, so
+ * most of these are defined to 0 which won't block anything.
+ * Additionally we reserve the lower 4 bits to future "group" blocking. For
+ * example we may introduce a way to block all the PCS#11 v3 function calls to
+ * simulate a 2.40 token */
+#define P11PROV_BLOCK_Initialize 0b0000000000000000
+#define P11PROV_BLOCK_Finalize 0b0000000000000000
+#define P11PROV_BLOCK_GetInfo 0b0000000000000000
+#define P11PROV_BLOCK_GetFunctionList 0b0000000000000000
+#define P11PROV_BLOCK_GetSlotList 0b0000000000000000
+#define P11PROV_BLOCK_GetSlotInfo 0b0000000000000000
+#define P11PROV_BLOCK_GetTokenInfo 0b0000000000000000
+#define P11PROV_BLOCK_GetMechanismList 0b0000000000000000
+#define P11PROV_BLOCK_GetMechanismInfo 0b0000000000000000
+#define P11PROV_BLOCK_OpenSession 0b0000000000000000
+#define P11PROV_BLOCK_CloseSession 0b0000000000000000
+#define P11PROV_BLOCK_GetSessionInfo 0b0000000000000000
+#define P11PROV_BLOCK_GetOperationState 0b0000000000001000
+#define P11PROV_BLOCK_SetOperationState 0b0000000000001000
+#define P11PROV_BLOCK_Login 0b0000000000000000
+#define P11PROV_BLOCK_Logout 0b0000000000000000
+#define P11PROV_BLOCK_CreateObject 0b0000000000000000
+#define P11PROV_BLOCK_CopyObject 0b0000000000000000
+#define P11PROV_BLOCK_DestroyObject 0b0000000000000000
+#define P11PROV_BLOCK_GetAttributeValue 0b0000000000000000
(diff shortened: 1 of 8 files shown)


src/pk11_uri.gen.c
```diff
@@ -7,8 +7,9 @@ extern P11PROV_PK11_URI *
 d2i_P11PROV_PK11_URI(P11PROV_PK11_URI **a, const unsigned char **in, long len);
 extern int i2d_P11PROV_PK11_URI(const P11PROV_PK11_URI *a, unsigned char **out);
 extern const ASN1_ITEM *P11PROV_PK11_URI_it(void);
-P11PROV_PK11_URI *d2i_P11PROV_PK11_URI(P11PROV_PK11_URI **a,
-                                       const unsigned char **in, long len)
+
+P11PROV_PK11_URI
+*d2i_P11PROV_PK11_URI(P11PROV_PK11_URI **a, const unsigned char **in, long len)
```
Jakuje, line 12: ugh ... is this really the preferred formatting of splitting the type and `*` to different lines?

src/provider.c
```diff
@@ -613,9 +613,12 @@ int p11prov_ctx_cache_sessions(P11PROV_CTX *ctx)
...
 }
 
-bool p11prov_ctx_no_operation_state(P11PROV_CTX *ctx)
+bool p11prov_ctx_is_call_blocked(P11PROV_CTX *ctx, uint64_t mask)
 {
-    return ctx->no_operation_state;
+    if (ctx->blocked_calls & mask) {
+        return true;
+    }
+    return false;
```
Jakuje, line 621: could be simpler like this:
```suggestion
    return (ctx->blocked_calls & mask);
```

src/interface.h
```diff
@@ -135,4 +135,58 @@ CK_RV side_channel_free_Decrypt(P11PROV_CTX *ctx, CK_SESSION_HANDLE hSession,
 
 CK_INFO p11prov_module_ck_info(P11PROV_MODULE *mctx);
 
+/* never blocked calls are masked to 0 by default */
+/* 2.40 */
```
Jakuje, line 139: I think the values here deserve some more explanation:
 * whats the point adding the check for the calls that can never be blocked? Do we assume compiler will compile the check out?
 * Why the only value (other than 0) so far is `0b1000` (8) and not `1` to start with?

src/interface.pre
```diff
@@ -10,6 +10,10 @@ BEGIN:
         P11PROV_raise(ctx, ret, "Can't get module interfaces"); \
         return ret; \
     } \
+    if (p11prov_ctx_is_call_blocked(ctx, P11PROV_BLOCK_##name)) { \
+        P11PROV_raise(ctx, ret, "C_" #name " is blocked"); \
```
Jakuje, line 14: This is super-ugly in the generated code. Can we do something like even though it would cause some more copying in runtime in theory?
```suggestion
        P11PROV_raise(ctx, ret, "C_%s is blocked", #name); \
```
(I would hope the compiler can optimize these though)

src/provider.c
```diff
@@ -1532,7 +1535,7 @@ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
             } else if (strncmp(str, "no-allowed-mechanisms", toklen) == 0) {
                 ctx->no_allowed_mechanisms = true;
             } else if (strncmp(str, "no-operation-state", toklen) == 0) {
```
Jakuje, line 1537: with this change, this option now affects also the `C_SetOperationState`. Should we update the documentation regarding this change?

src/pk11_uri.gen.c
simo5: clang-format decides, I don't :-)

src/provider.c
simo5: true, and without the parens even... will change

src/interface.h
simo5: The pre-generated code from interface.pre needs all the definitions, but we need to  assign an actual value only once we provide a knob to turn that value on/off

I thought about pre-assigning values too, but not sure that is useful.

src/pk11_uri.gen.c
Jakuje: But it has configuration ;)

src/interface.pre
simo5: it probably won't, but I am also changing this raise to a debug, as one of the reasons we had to block the operation was also unnecessary noise in the openssl logs

src/provider.c
simo5: Set and Get are a pair the no_opertion_state flag always meant to block both (which are always used together in the code, currently) so I do not think we need a doc change for that.

simo5:
Uhmm I am not sure why the test fails ... investigating ...

Jakuje (approved):
lgtm

PR #385, opened 2024-04-23
Address various issues with meson builds

simo5:
#### Description

The change to the meson build system regressed some minor things that I am attempting to fix here:
- Output also individual log files
  Meson defaults to a ginormous log file with the output of all tests mashed together, this works but makes looking at specific test failures harder.
- Turn the custom targets that werre setting up tests into "tests", this allows us to run them only during the tests phase and not during the compile phase, which is unwanted
- Add dependencies from tests to test executables so they are rebuilt if missing
- Change make targets so we only build pkcs11.so and not all of the tests as well, this prevents coverity from flagging stuff in the test suite among other things
- Do not run tests in parallel, there are some unstated dependencies between tests so they will flake randomly via race conditions. Until we can set proper dependencies between tests just disable parallel runs, it doesn't change the total run by much
- Restore collection of individual log files

#### Checklist

- [ ] ~Code modified for feature~
- [ ] ~Documentation updated~

#### Reviewer's checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,10 @@
 .PHONY: all check check-style check-style-show check-style-fix clean generate-code generate-docs
 
 all:
-	meson setup builddir
-	meson compile -C builddir
+	if [ ! -d "builddir" ]; then \
+		meson setup builddir; \
+	fi; \
+	meson compile -C builddir pkcs11
 
 check:
 	meson test -C builddir
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -148,7 +148,7 @@ foreach t, extra_args : tests
       suite: suite,
       env: test_env,
       depends: test_executables,
-      is_parallel: is_parallel,
+      is_parallel: false,
     )
   endforeach
 endforeach
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -104,10 +104,13 @@ test_programs = [
   'pincache',
 ]
 
+test_executables = []
 foreach t : test_programs
-  executable(t, '@0@.c'.format(t),
-             include_directories: [configinc],
-             dependencies: [libcrypto, libssl])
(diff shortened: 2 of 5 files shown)


simo5:
@ueno can you take a look at this one?

Makefile
```diff
@@ -1,8 +1,10 @@
 .PHONY: all check check-style check-style-show check-style-fix clean generate-code generate-docs
 
 all:
-	meson setup builddir
-	meson compile -C builddir
+	if [ ! -d "builddir" ]; then \
+		meson setup builddir; \
+	fi; \
```
ueno, line 6:
```suggestion
	meson setup --wipe builddir
```

tests/meson.build
```diff
@@ -30,25 +30,20 @@ if nss_softokn.found()
...
-  env: conf_env,
-  command: [
-    find_program('setup-softhsm.sh'),
-  ],
-  capture: true,
-)
+foreach name, targs : test_setup
+  test(
+    name,
+    targs.get('exe'),
```
ueno, line 41: No need to capture the output of `setup-*.sh`?

ueno (approved):
Looks good to me, only a couple of nits.

Makefile
simo5: why would I wipe the setup every time I run make ?

tests/meson.build
simo5: I thought about it, but it is not as important as individual tests, and the output of setup is always at the top of the meson log file.
And to capture the outpup I would have to run the setup scripts through a wrapper or change the scripts to redirect all their output and that was more change than I wanted to do with this PR, so I decided to just let it be for now.

neverpanic (approved):

PR #392, opened 2024-05-14
feat: RSA_PKCS1_WITH_TLS_PADDING padding developped in the provider

sebastienandert:
Hi,  

I work in a company and for these 3 last months, we have worked on the embedding of the opensslv3 functions in our application.  
The aim of these implementation is to access to our HSM (Eviden) via opensslv3 functions in PKCS11 using your provider.  

Unfortunately, a test failed due to a function not implemented in your provider.  

# Missing function in the provider

When we connect an openssl s_client to an openssl s_server using these 2 commands :  
`./openssl s_server -port 4456 -provider pkcs11 -key "pkcs11:object=nom;id=1122334455-000000000000000000000000-98" -cert /int1/pki/store/user/crt/C2P_HSM_20261205.crt`  
and  
`./openssl s_client -connect 127.0.0.1:4456 -CApath /int1/pki/store/ac/crt_rehash -cipher AES128-GCM-SHA256`

We get the following error coming from Openssl libraries :

```c++
-----------------------------------------------------------------
ACCEPT
ERROR
0011F0435A7F0000:error:1C8000A5:Provider routines:p11prov_rsaenc_set_ctx_params:illegal or unsupported padding mode:asymmetric_cipher.c:491:
0011F0435A7F0000:error:0A000093:SSL routines:tls_process_cke_rsa:decryption failed:ssl/statem/statem_srvr.c:2946:
CONNECTION CLOSED

In the openssl sources, this error comes from the function : 

openssl-3.0.11/ssl/statem/statem_srvr.c:tls_process_cke_rsa:2909
    if (EVP_PKEY_decrypt_init(ctx) <= 0
            || EVP_PKEY_CTX_set_rsa_padding(ctx, RSA_PKCS1_WITH_TLS_PADDING) <= 0) {
        SLfatal(s, SSL_AD_DECRYPT_ERROR, SSL_R_DECRYPTION_FAILED);
        goto err;
    }
-----------------------------------------------------------------
```

After inverstigating on this issue, it appears that ciphers using RSA key exchange type involves to use the padding : RSA_PKCS1_WITH_TLS_PADDING, which is not supported by your provider.
hence this error raises by your provider :

```c++
-----------------------------------------------------------------
0011F0435A7F0000:error:1C8000A5:Provider routines:p11prov_rsaenc_set_ctx_params:illegal or unsupported padding mode:asymmetric_cipher.c:491:

if (mechtype == CK_UNAVAILABLE_INFORMATION) {
    P11PROV_debug("mechtype == CK_UNAVAILABLE_INFORMATION");
    ERR_raise(ERR_LIB_PROV, PROV_R_ILLEGAL_OR_UNSUPPORTED_PADDING_MODE);
    return RET_OSSL_ERR;
}
-----------------------------------------------------------------
```

Then, we have implemented this unsupported padding (mandatory for our needs) in order to continue using you provider.  

It's now 1 month our improved provider runs under our UAT environment without error anymore.  
Of course we would like to share you our implementation and to know if you could integrate it in your sources in a near futur ?  

Thanks,
Kind regards,

### DIFF
--- a/src/asymmetric_cipher.c
+++ b/src/asymmetric_cipher.c
@@ -4,6 +4,9 @@
 #include "provider.h"
 #include <string.h>
 #include "openssl/rsa.h"
+#include "openssl/prov_ssl.h"
+#include <fcntl.h>
+#include <unistd.h>
 
 DISPATCH_RSAENC_FN(newctx);
 DISPATCH_RSAENC_FN(freectx);
@@ -24,6 +27,9 @@ struct p11prov_rsaenc_ctx {
 
     CK_MECHANISM_TYPE mechtype;
     CK_RSA_PKCS_OAEP_PARAMS oaep_params;
+    int ossl_id;            /* real OSSL padding mode */
+    int client_version;     /* RSA_PKCS1_WITH_TLS_PADDING: client_version */
+    int alt_version;        /* RSA_PKCS1_WITH_TLS_PADDING: negociated_version */
 };
 
 static void *p11prov_rsaenc_newctx(void *provctx)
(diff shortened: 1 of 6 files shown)


src/asymmetric_cipher.c
```diff
@@ -24,6 +27,9 @@ struct p11prov_rsaenc_ctx {
 
     CK_MECHANISM_TYPE mechtype;
     CK_RSA_PKCS_OAEP_PARAMS oaep_params;
+    int ossl_id;            /* real OSSL padding mode */
+    int client_version;     /* RSA_PKCS1_WITH_TLS_PADDING: client_version */
+    int alt_version;        /* RSA_PKCS1_WITH_TLS_PADDING: negociated_version */
```
simo5, line 32: These are passed as int by openssl for convenience, but are actually each a pair of 1 byte values.
I would either declare them as uint16_t or even better already split thme into the component bytes and declare a pair of uint8_t for each.

src/asymmetric_cipher.c
```diff
@@ -508,16 +671,34 @@ static int p11prov_rsaenc_set_ctx_params(void *ctx, const OSSL_PARAM params[])
...
+    if (p) {
+        ret = OSSL_PARAM_get_int(p, &encctx->client_version);
+        if (ret != RET_OSSL_OK) {
+            return ret;
+        }
+    }
+   
+    p = OSSL_PARAM_locate_const(params, OSSL_ASYM_CIPHER_PARAM_TLS_NEGOTIATED_VERSION);
+    if (p) {
+        ret = OSSL_PARAM_get_int(p, &encctx->alt_version);
```
simo5, line 696: same as above

src/encoder.c
```diff
@@ -655,10 +655,11 @@ static int p11prov_ec_set_keypoint_data(const OSSL_PARAM *params, void *key)
 
 static X509_PUBKEY *p11prov_ec_pubkey_to_x509(P11PROV_OBJ *key)
 {
-    struct ecdsa_key_point keypoint = { 0 };
+    struct ecdsa_key_point keypoint;
     X509_PUBKEY *pubkey;
     int ret;
-
+    
+    memset (&keypoint, 0, sizeof (keypoint));
```
simo5, line 662: please don't do this, the above initialization is just fine.

src/interface.c
```diff
@@ -168,11 +169,7 @@ static CK_RV p11prov_interface_init(P11PROV_MODULE *mctx)
     ret = CKR_FUNCTION_NOT_SUPPORTED;
     intf->GetInterface = dlsym(mctx->dlhandle, "C_GetInterface");
     if (!intf->GetInterface) {
-        char *err = dlerror();
-        P11PROV_debug(
-            "C_GetInterface() not available. Falling back to "
-            "C_GetFunctionList(): %s",
-            err);
+        err = dlerror();
```
simo5, line 172: this change does not belong in this PR, please revert it.

simo5: this comment applies to all the changes in this file

src/interface.c
```diff
@@ -373,14 +380,16 @@ void p11prov_module_mark_reinit(P11PROV_MODULE *mctx)
...
+    CK_INFO ck_info;
     CK_RV ret;
 
     if (!mctx) {
         return CKR_GENERAL_ERROR;
     }
 
+    memset (&args, 0, sizeof (args));
+    memset (&ck_info, 0, sizeof (ck_info));
     ret = MUTEX_LOCK(mctx);
```
simo5, line 393: don't do this the above initializations are just fine

src/kdf.c
```diff
@@ -349,7 +349,11 @@ static int p11prov_hkdf_get_ctx_params(void *ctx, OSSL_PARAM *params)
     if (p) {
         size_t ret_size = 0;
         if (hkdfctx->params.bExpand != CK_FALSE) {
-            ret_size = SIZE_MAX;
+#ifndef SIZE_MAX
```
simo5, line 352: If you do not have SIZE_MAX define it in a header, not in here.
What platform lacks this in a header?

src/objects.c
```diff
@@ -991,7 +991,8 @@ CK_RV p11prov_obj_find(P11PROV_CTX *provctx, P11PROV_SESSION *session,
     P11PROV_debug("Find objects [class=%lu, id-len=%lu, label=%s]", class,
                   id.ulValueLen,
                   label.type == CKA_LABEL ? (char *)label.pValue : "None");
-
+    
+    memset (template, 0, sizeof (template));
```
simo5, line 995: again no, the original initialization is just fine.

simo5: I won't repeat it again for each occurence, but please revert all of these changes

src/provider.c
```diff
@@ -1358,6 +1359,16 @@ static struct p11prov_cfg_names {
...
+static void pkcs11prov_dbg_init (void)
+{
+    /* for debugging only
+     * Do nothing else than giving the possibility to set a breakpoint
+     * NOTE: don't compile with '-g -Og', use either '-g' or '-g -O0'
+     *       with -Og, the optimizer remove this call witch does nothing
+     */
+    return ;
+}
+
```
simo5, line 1371: Please remove this

simo5 (changes requested):
Thanks for this PR, there is a lot to adjust, especially the actual core of the function is completely side-channel unsafe and will require careful rewrite.

Please avoid churn unnecessary to implement the actual feature when submitting a PR, it just makes it harded to review.

(Also I will not accept regressing from basic initialization features, if you need to support an old compiler that can't deal with initializers, then please keep that as a local patch to your environment for now)

tomato42:
@sebastienandert wrote:
> It's now 1 month our improved provider runs under our UAT environment without error anymore.

there's a subtle difference between "it works" and "it's secure": https://people.redhat.com/~hkario/marvin/
this code is almost certainly vulnerable to side-channel attacks even if the HSM performs the private key operations in constant time (and that's an "if" in itself)

simo5:
Given I saw no answers to the review, and the proposed change required major overhauls pretty much everywhere I went ahead and implemented it diffrently in PR #402 

Closing this one, thanks for the contribution.

PR #397, opened 2024-05-23
Fix CKA_ALLOWED_MECHANISMS attribute generation

simo5:
#### Description
The spec says the length is the size of the data
The current code sets the number of elements as length, but that seems incorrect.

#### Checklist

- [ ] ~Documentation updated~

#### Reviewer's checklist:


Jakuje (dismissed):
Do you have some tests with this attribute? I know softhsm supports this but I do not recall if we used this in the tests somewhere. I am not aware of any other implementation supporting this attribute for now.

simo5:
@Jakuje 
I do not have a test and this bothers me, do you know if we can enable this on softhsm easily?
I think we should have a test to ensure we are doing the right thing instead of blindly following the spec. The spec sometimes *is* wrong or confusing, or under-specified, although this specific case is probably straightforward enough.

simo5:
Soo it looks like Ubuntu has a version of softhsm that is too old ?

simo5:
> Soo it looks like Ubuntu has a version of softhsm that is too old ?

Ugh no that is not the problem, somehow Ubuntu is getting confused by the request to use CTR-DRBG via provider ...
```
genpkey: Error generating RSA-PSS key
40173885727F0000:error:13000084:engine routines:dynamic_load:dso not found:../crypto/engine/eng_dyn.c:442:
40173885727F0000:error:13000074:engine routines:ENGINE_by_id:no such engine:../crypto/engine/eng_list.c:433:id=rdrand
40173885727F0000:error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:../crypto/evp/evp_fetch.c:386:Global default library context, Algorithm (CTR-DRBG : 122), Properties (provider=pkcs11)
```

@beldmit have you seen this before perchance?

beldmit:
Not for rdrand...

src/keymgmt.c
```diff
@@ -986,90 +1004,124 @@ const OSSL_DISPATCH p11prov_rsa_keymgmt_functions[] = {
...
-        ret = p11prov_obj_set_attributes(ctx->provctx, NULL, key, template,
-                                         tsize);
-        if (ret != CKR_OK) {
-            P11PROV_debug("Failed to add RSAPSS restrictions (%lu)", ret);
-            if (ret == CKR_ATTRIBUTE_TYPE_INVALID) {
-                /* set quirk to disable future attempts for this token */
                 token_supports_allowed_mechs = CK_FALSE;
-                (void)p11prov_token_sup_attr(
-                    ctx->provctx, p11prov_obj_get_slotid(key), SET_ATTR,
+                /* if he quirk has never been set we'll get back what we
```
Jakuje, line 1036:
```suggestion
                /* if the quirk has never been set we'll get back what we
```

src/keymgmt.c
```diff
@@ -986,90 +1004,124 @@ const OSSL_DISPATCH p11prov_rsa_keymgmt_functions[] = {
...
+    if (!ctx) {
+        return RET_OSSL_ERR;
+    }
+
+    if (params == NULL) {
+        return RET_OSSL_OK;
+    }
+
+    p = OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_RSA_DIGEST);
+    if (p) {
```
Jakuje, line 1087: Should we try to set all PSS mechanisms in case there is no digest specified here in the else branch as we did previously with `set_default_rsapss_mechanisms()`, which is now deleted?

simo5: No, I deleted it on purpose because setting all mechanisms is not really useful, omitting the attribute completely is more efficient.

Jakuje: This was setting just the PSS mechanisms technically preventing to use RAW rsa and PKCS#1.5, which was the main point, you will not be able to achieve without that (unless you force specific digest).

Jakuje: I think that was main point of introducing RSA-PSS as separate key type in openssl was to not allow any other operations on these keys so that would be my preference to do this enforcing, if the token supports it.

simo5:
Fixed a minor issue found by coverity and rebase on main.

src/keymgmt.c
```diff
@@ -1013,63 +1041,149 @@ static void *p11prov_rsapss_gen(void *genctx, OSSL_CALLBACK *cb_fn,
...
-    if (token_supports_allowed_mechs == CK_TRUE) {
-        CK_ATTRIBUTE template[] = {
-            { CKA_ALLOWED_MECHANISMS, ctx->data.rsa.allowed_types,
-              ctx->data.rsa.allowed_types_size },
-        };
-        CK_ULONG tsize = 1;
+    if (token_supports_allowed_mechs) {
+        /* We always want to restrict PSS keys to just the PSS mechanisms.
+         * If a specific restriction is not already set then set defaults */
+        if (!ctx->data.rsa.allowed_types) {
```
Jakuje, line 1061: This is just a nit regarding the naming, the `allowed_types` might sounds a bit confusing. I would prefer `allowed_mechs` to be clear what it is on the first sight.

tests/trsapssam
```diff
@@ -0,0 +1,58 @@
...
+
+#Need to enable this or genpkey fails because it can't output the private key
+sed -e "s/#pkcs11-module-encode-provider-uri-to-pem/pkcs11-module-encode-provider-uri-to-pem = true/" \
+    "${OPENSSL_CONF}" > "${OPENSSL_CONF}.rsapss_genpkey"
+OPENSSL_CONF=${OPENSSL_CONF}.rsapss_genpkey
+
+ossl '
+genpkey -propquery "?provider=pkcs11"
+        -algorithm "RSA-PSS" -pkeyopt "rsa_pss_keygen_md:SHA256"
+        -pkeyopt "pkcs11_uri:pkcs11:object=Test-RSA-PSS-Restrictions"'
```
Jakuje, line 15: This tests the case where we have explicit hash. I would prefer to have also test generating a key without the explicit hash and make sure the key can not be used for RSA PKCS#1.5 signature.

src/keymgmt.c
simo5: Well we have "mechanism types" listed here, so "types" is just fine imo.
Not planning on changing that structure in this PR at any rate.

tests/trsapssam
simo5: Yeah I will add that and respin

Jakuje (approved):
lgtm!

simo5:
Uhmm I need to fix the CI scripts to recognize that the no-covscan label is already present when a PR gets updated ... and as usual Bind test fails, but that test is not binding (pun fully intended :-) ).

Thanks for the review @Jakuje

PR #398, opened 2024-05-23
Fix session count types

bith3ad:
#### Description

Since commit f1e85afeb0fe ("Session locking fixes") the session count variables are int. This is wrong since max_sessions is initialized by the token ulMaxSessionCount member so it of type CK_ULONG. Interpreting it as int is wrong since all values >= BIT(31) are interpreted as negative value.

Fix this by converting all session counting vars back to CK_ULONG.

#### Reviewer's checklist:

- [x] ~Any issues marked for closing are addressed~
- [x] ~This feature/change has adequate documentation added~

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -26,10 +26,10 @@ struct p11prov_session_pool {
     P11PROV_CTX *provctx;
     CK_SLOT_ID slotid;
 
-    int num_sessions;
-    int max_sessions;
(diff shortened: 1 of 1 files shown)


simo5:
The bind failure is a known fail, an unrelated.
The rest looks great, and coverity shows all clear.

Thanks for the contribution @bith3ad

PR #401, opened 2024-05-30
Use OpenSSL random functions

### LINKED ISSUE #280: Key generation should not use token GenerateRandom()
When generating a key on token, and if ID is not specified, pkcs11-provider falls back to generating random bytes in https://github.com/latchset/pkcs11-provider/blob/main/src/keymgmt.c#L375-L382.

This is problematic because a token may support key generation, might not implement the pkcs#11 random generator interface. Using the token random generator may also be unexpectedly slow.

While object ID is a mandatory object attribute, it is not a required attribute in key generation template. Omitting it from the template is valid, and usually the tokens will generate deterministic ID by hashing the public key components (or possibly just randomizing it for secret keys).  The token generated ID can be read back from the generated key object.

simo5:
Let *that* cycle back into the provider and call C_GenerateRandom() if that's how the properties end up wiring things.

Fixes #280

#### Description

After some consideration I think using the openssl RAND functions is just fine here.
First of all it is not a security sensitive thing, and secondarily those functions will still cycle back to use the C_GenerateRandom function if we probed that the token does have a random generator and properties force it.

The existing generation test should eb sufficient to cover this change, and no documentation needs to be updated.

#### Checklist

- [ ] ~Documentation updated~

#### Reviewer's checklist:

### DIFF
--- a/src/keymgmt.c
+++ b/src/keymgmt.c
@@ -3,6 +3,7 @@
 
 #include "provider.h"
 #include "platform/endian.h"
+#include "openssl/rand.h"
 #include <string.h>
 
 #define DFLT_DIGEST "SHA256"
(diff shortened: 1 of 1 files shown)


Jakuje (dismissed):

src/keymgmt.c
```diff
@@ -498,9 +499,13 @@ static int p11prov_common_gen(struct key_generator *ctx,
     sh = p11prov_session_handle(session);
 
     if (cka_id.ulValueLen == 0) {
+        int err = RET_OSSL_ERR;
         /* generate unique id for the key */
-        ret = p11prov_GenerateRandom(ctx->provctx, sh, id, sizeof(id));
-        if (ret != CKR_OK) {
+        err = RAND_priv_bytes_ex(p11prov_ctx_get_libctx(ctx->provctx), id,
+                                 sizeof(id), 0);
```
beldmit, line 505: Why do we need priv_bytes here?

simo5: Uhmm indeed we do not actually need that .. let me change it to the public one.

Jakuje (approved):

PR #403, opened 2024-06-04
CI Maintenance

simo5:
Some CI maintenance
- Disabled failing Bind integration tests
- Better covscan related label/status management

## New Covscan Behavior/Process

# Labels
Only two labels exist:
- covscan
  This label triggers an actual covscan run, do not abuse we have only 4 scans per day total
- covscan-ok
  Once one of the admins receives the report set this label if passed

# Behavior
1. On setting covscan label a coverity scan is triggered and results are sent to an admin, the covscan label is removed to prevent unwanted re-runs and to avoid confusion on the status of the PR
2. On setting the covscan-ok label the covertiy test passes (test is green)
3. On pushing new commits if no labels are present:
   - if no files in src/* are touched in the whole PR covscan will be deemed no neccessary (test is green)
   - if any file in src/* is changed in any commit then covscan will be deemed required (test is red)
4. On pushing new commits if covscan-ok label is present:
   - if no *new* changes in src/ then covscan is considered ok (test is green)
   - if a rebase happened (tooling can't properly check for changes on rebase) or *new* changes appear in src/ then covscan will be required, and the covscan-ok label is removed

Note that just removing the covscan-ok label will not rerun the covscan test and will not cause a test failure

# Required status
The covscan test is a required status, if the test is not explicitly green the PR will not be pushable without admin override.

### DIFF
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -5,6 +5,7 @@ on:
   pull_request_target:
     branches: ["main"]
     types:
+      - synchronize
       - labeled
   schedule:
     - cron: '41 3 * * 0'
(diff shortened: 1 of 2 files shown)


simo5:
Last use of no-covscan before I make it go away.

PR #415, opened 2024-07-08
Fix CID 500198: Integer handling issues

simo5:
#### Description
Coverity seem to have updated its rules and is now more concerned about integer overflows than before. This place has no change of overflowing, but lets play ball and silence it with a check.
CID 500198: Integer handling issues (INTEGER_OVERFLOW)

#### Checklist

- ~[ ] Test suite updated with functionality tests~
- ~[ ] Test suite updated with negative tests~
- ~[ ] Documentation updated~

#### Reviewer's checklist:

- ~[ ] Any issues marked for closing are addressed~
- ~[x] Commits have short titles and sensible commit messages~

### DIFF
--- a/src/keymgmt.c
+++ b/src/keymgmt.c
@@ -1461,6 +1461,10 @@ static const char *p11prov_ec_query_operation_name(int operation_id)
     return NULL;
 }
 
+#define CURVE_521_BITS 521
+#define MAX_CURVE_BITS CURVE_521_BITS
+#define MAX_CURVE_SIZE ((MAX_CURVE_BITS + 7) / 8)
+
 static int p11prov_ec_secbits(int bits)
 {
     /* common values from various NIST documents */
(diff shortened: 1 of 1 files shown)


src/keymgmt.c
```diff
@@ -1473,7 +1477,7 @@ static int p11prov_ec_secbits(int bits)
     if (bits < 384) {
         return 128;
     }
-    if (bits < 512) {
+    if (bits < CURVE_521_BITS) {
```
manison, line 1480: Is the change of the value (512 -> 521) intentional here or is it a typo? OpenSSL's default implementation uses 512 as the threshold as was the case before this change.

simo5: Well given there is no 512 bit curve the distinction doesn't really exist.
But I can revert this change if you think this can be too confusing.

simo5: I actually reverted this change, given it does not really belong in a coverity related fix anyway.

simo5:
Integration tests are currently broken due to rawhide changes :/

PR #417, opened 2024-07-08
Set the raw point for ECDH public data params

### LINKED ISSUE #416: ECDH Params public data should be raw format by default, not DER
**Describe the bug**
When the pPublicData parameter is filled we source CKA_EC_POINT which is the DER encoded format of the EC public point, however the spec says:

```
pPublicData       pointer to other party’s EC public key value. For short Weierstrass
                  EC keys: a token MUST be able to accept this value encoded as a raw
                  octet string (as per section A.5.2 of [ANSI X9.62]).  A token MAY, 
                  in addition, support accepting this value as a DER-encoded ECPoint 
                  (as per section E.6 of [ANSI X9.62]) i.e. the same as a CKA_EC_POINT
                  encoding.  The calling application is responsible for converting the
                  offered public key to the compressed or uncompressed forms of these
                  encodings if the token does not support the offered form.
```

Both softhsm and NSS autodetect and convert as needed, but kryoptic is failing ECDH tests because of this.

**To Reproduce**
Enable ECDH test for kryoptic

**Expected behavior**
We should provide the format that the spec lists as MUST for this parameter, not what is listed as MAY, as some tokens MAY not support it.

 - OS: Fedora
 - Version: rawhide

 - Device: Kryoptic
 - PKCS11 Driver version: main tree
 - Application: Pkcs11-provider's CI
 - Version: main tree

**Additional context**
Related Kryoptic bug at https://github.com/latchset/kryoptic/issues/66

simo5:
#### Description

Changes the format of the point passed into ECDH Params

Enables Kryoptic ECDH test, which now passes.

Fixes #416 

#### Checklist

- ~[ ] Documentation updated~

#### Reviewer's checklist:

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -2175,6 +2175,65 @@ done:
     return ret;
 }
 
+CK_ATTRIBUTE *p11prov_obj_get_ec_public_raw(P11PROV_OBJ *key)
+{
+    CK_ATTRIBUTE *pub_key;
+
+    if (!key) {
+        return RET_OSSL_ERR;
+    }
+
+    if (key->data.key.type != CKK_EC) {
+        P11PROV_raise(key->ctx, CKR_GENERAL_ERROR, "Unsupported key type");
+        return RET_OSSL_ERR;
+    }
+
+    if (key->class != CKO_PRIVATE_KEY && key->class != CKO_PUBLIC_KEY) {
+        P11PROV_raise(key->ctx, CKR_GENERAL_ERROR, "Invalid Object Class");
+        return RET_OSSL_ERR;
+    }
+
+    pub_key = p11prov_obj_get_attr(key, CKA_P11PROV_PUB_KEY);
+    if (!pub_key) {
+        CK_ATTRIBUTE *ec_point;
+
+        ec_point = p11prov_obj_get_attr(key, CKA_EC_POINT);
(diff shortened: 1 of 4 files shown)


src/objects.c
```diff
@@ -2175,6 +2175,65 @@ int p11prov_obj_get_ec_public_x_y(P11PROV_OBJ *obj, CK_ATTRIBUTE **pub_x,
...
+        P11PROV_raise(key->ctx, CKR_GENERAL_ERROR, "Invalid Object Class");
+        return RET_OSSL_ERR;
+    }
+
+    pub_key = p11prov_obj_get_attr(key, CKA_P11PROV_PUB_KEY);
+    if (!pub_key) {
+        CK_ATTRIBUTE *ec_point;
+
+        ec_point = p11prov_obj_get_attr(key, CKA_EC_POINT);
+        if (ec_point) {
```
Jakuje, line 2201: Would not be nicer to reverse this condition to avoid nesting the whole processing below?

Jakuje (approved):
otherwise lgtm

src/objects.c
simo5: Maybe, but given this works and it is not terribly ugly, I'll take your LGTM and just merge as is :)

PR #420, opened 2024-07-10
Off-by-one error in pool consistency check

glguy:
I'm getting a segmentation fault on this line because pool->objects can be NULL. I suspect this > should be >=, though I'm not quite sure.

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -234,7 +234,7 @@ static void obj_rm_from_pool(P11PROV_OBJ *obj)
     }
 
     /* LOCKED SECTION ------------- */
-    if (obj->poolid > pool->size || pool->objects[obj->poolid] != obj) {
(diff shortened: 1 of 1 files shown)


simo5:
Good catch, but odd that you hit that condition

glguy:
Thanks for the merge. I'm able to generate that condition in some cases using the PIV application on my YubiKey when I use a sufficiently general URI to query the private key. I haven't narrowed down the actual failure and I understand that ideally this consistency condition should never fail. I know that I can cause it to stop happening when I include `;type=private` in my URI. So my hunch is that one of the *other* object types that `pkcs11:id=%01` or `pkcs11:id=%02` on my device matches isn't completely supported.

PR #422, opened 2024-07-17
Extend the ttls test to be able to test different configurations

Jakuje:
#### Description

The TLS tests now tests just the default parameters with RSA key. This PR extends it to test also TLS 1.2, ECDSA keys and specific ciphersuites.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/address-sanitizer.yml
+++ b/.github/workflows/address-sanitizer.yml
@@ -57,7 +57,7 @@ jobs:
         run: |
           meson compile -C builddir
           meson test --num-processes 1 -C builddir
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         if: failure()
         with:
           name: Address sanitizer logs on ${{ matrix.name }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,7 +65,7 @@ jobs:
         run: |
           meson compile -C builddir
           meson test --num-processes 1 -C builddir
-      - uses: actions/upload-artifact@v3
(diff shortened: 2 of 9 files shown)


Jakuje:
Removed the commit that made it failing all over the place and keeping just the fixups that should make the CI pass (except for mac?) and that should be generally useful.

The commit making all of that failing is available in the wip branch https://github.com/Jakuje/pkcs11-provider/commits/ttls-test-wip/ and I believe they might be useful for reproducing further TLS issues.

simo5:
Uhm in the second macos test faile di see that softhsm setup fails, but softokn and kryoptic setups succeed, yet *all* tests are skipped, would you happen to know why ?

neverpanic:
@Jakuje Sorry for the delay on this – I checked out this branch, built it locally and ran the tests, and they all pass on my system. This either means that the issue isn't easily reproducible, or it only happens with components from Homebrew (I use MacPorts).

If you don't have any idea what's going on I can spend some more time to investigate and attempt to reproduce, but that will involve actually setting up a macOS VM, so unless we have no further avenues to follow up on, I'd like to avoid this.

Jakuje:
Sounds like the issue was with the server that I did not kill it after each test. For some reason, it did not matter for linux, but did for macos. Now all looks green, except for the coverity, which I think is not relevant for this change.

simo5:
The coverity scan is "failed" because it detected a code change after the covscan-ok label was set.

simo5:
But there is no code change, so that's odd, I will have to recheck the logic in the CI automation, I will manually set the label after review

simo5 (approved):
LGTM

Issue #395, opened 2024-05-22
ECDH TLS groups support in Keymgmt


simo5:
Hi

This is my openssl.cnf after applying EVP configuration:-
```
[openssl_init]
providers = provider_sect
alg_section = algorithm_sect

# List of providers to load
[provider_sect]
default = default_sect
pkcs11 = pkcs11_sect

[default_sect]
activate = 1
[pkcs11_sect]
module = /usr/lib/pkcs11.so
pkcs11-module-path = /usr/lib/libckteec.so.0
pkcs11-module-cache-keys = false
pkcs11-module-quirks = no-operation-state
activate = 1

[algorithm_sect]
default_properties = ?provider=pkcs11
```

**Server side command:-**
`openssl s_server -key  "pkcs11:model=OP-TEE%20TA;manufacturer=Linaro;serial=0000000000000000;token=kshitiz_test;id=%02;object=ecc-key-521;type=private?pin-value=1234" -cert req_client.crt -accept 443 -Verify 3 -named_curve secp521r1 -ciphersuites 'TLS_AES_256_GCM_SHA384' -tls1_3`

**Client side command:-**
`./openssl s_client -connect 10.232.132.242:443 -tls1_3 -ciphersuites 'TLS_AES_256_GCM_SHA384' -cert server.crt -key server.key  -CAfile ../ca.crt`

**Error recieved on server side:-**
```
verify depth is 3, must return a certificate
Using default temp DH parameters
ACCEPT
ERROR
20904DB5FFFF0000:error:40800007:pkcs11:p11prov_common_gen_init:Invalid or improper arguments were provided to the invoked function:/usr/src/debug/pkcs11-provider/0.3/src/keymgmt.c:150:Unsupported selection
20904DB5FFFF0000:error:03000086:digital envelope routines:gen_init:initialization error:/usr/src/debug/openssl/3.2.1/crypto/evp/pmeth_gn.c:52:
20904DB5FFFF0000:error:0A00013A:SSL routines:tls_parse_ctos_key_share:unable to find ecdh parameters:/usr/src/debug/openssl/3.2.1/ssl/statem/extensions_srvr.c:684:
```
Am I doing something wrong here? And one more thing, that if we are able to reach to optee in sign verification operation without applying EVP configuration change, so why this change is required for exchange operation? Any specific reason?

_Originally posted by @kshitizvars in https://github.com/latchset/pkcs11-provider/discussions/389#discussioncomment-9527061_

[...]

PR #424, opened 2024-07-29
Add CKA_DERIVE flag in server's private key template

kshitizvars:
#### Description

 This commit adds CKA_DERIVE flag in server's private key template which gets checked by optee subsystem to derive shared secret. Tested TLS1.2 with the change.
Without this flag enabled, tls connection fails while deriving shared secret.
This change is in common code, please let me know where I can change instead of changing common code. 

#### Checklist

- [x] Code modified for feature

#### Reviewer's checklist:

- [ ] Any issues marked for closing are addressed
- [ ] There is a test suite reasonably covering new functionality or modifications
- [ ] This feature/change has adequate documentation added
- [ ] Code conform to coding style that today cannot yet be enforced via the check style test
- [ ] Commits have short titles and sensible commit messages
- [ ] Coverity Scan has run if needed (code PR) and no new defects were found

PR #427, opened 2024-07-30
tests: Run TLS tests also when forcing all server operations on token

Jakuje:
#### Description

This is likely the missing bit we had when trying to reproduce the issue #395. These are changes forcing all the TLS operations on token for the s_server. The s_client fails the basic key verification, because it is likely not correctly imported to the provider (?).

With this configuration the TLS 1.3 with RSA and ECDSA host certs work ok, TLS 1.2 with RSA too, but TLS 1.2 with ECDSA fails with 
```
00EE1FD59A7F0000:error:0A0000C1:SSL routines:tls_post_process_client_hello:no shared cipher:ssl/statem/statem_srvr.c:2220:^M
```
I was not able to figure out what OpenSSL needs at this point. I did not see any failures on the pkcs11 nor on the pkcs11 provider layer.

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/tests/openssl.cnf.in
+++ b/tests/openssl.cnf.in
@@ -28,6 +28,7 @@ pkcs11-module-token-pin = file:@PINFILE@
 #pkcs11-module-allow-export
 #pkcs11-module-load-behavior
 #pkcs11-module-block-operations
+#pkcs11-module-cache-keys
 ##QUIRKS
 activate = 1
 
--- a/tests/ttls
+++ b/tests/ttls
@@ -48,11 +48,13 @@ run_test() {
         expect {
             eof {exit 0;};
             default {exit 1;};
-        }" > "${TMPPDIR}/s_server_output" &
+        }" 2>&1 | tee "${TMPPDIR}/s_server_output" &
     SERVER_PID=$!
 
     read -r < "${TMPPDIR}/s_server_ready"
 
+    # The point is to force the server to use the pkcs11-provider for all operations, not the client now
+    OPENSSL_CONF="${ORIG_OPENSSL_CONF}" \
     expect -c "spawn $CHECKER openssl s_client -connect \"localhost:${PORT}\" -CAfile \"${CACRT}\" $CLNT_ARGS;
         set timeout 60;
         expect {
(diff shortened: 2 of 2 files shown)


simo5:
Have you tried importing #424 to see if that improves something ?

simo5:
If that does not help it may be some issue with the group aliases, OpenSSL is quite baroque and uses multiple names for ciphers in some places, and then insists on specific names among the aliases in other places.

Jakuje:
> Have you tried importing #424 to see if that improves something ?

Good point. Tried now, but the error is the same.

Jakuje:
> This has basically been done in #446, so I am going to close this PR.

This is actually not true. The #446 added just one test forcing the operation on the token, while here I added all the existing TLS tests to be executed. But all of them seems to work now in the way how they are written.

When I add the test from #449, it fails as described in the issue, but the `expect` is not catching the error for some reason (which needs some more digging), even though we got the failure as described:
```
spawn openssl s_client -connect localhost:23456 -CAfile /home/jjelen/devel/pkcs11-provider/builddir/tests/softokn/CAcert.pem -tls1_3 -ciphersuites TLS_AES_256_GCM_SHA384 -groups secp256r1
Connecting to ::1
CONNECTED(00000004)
80220C337F7F0000:error:0A000438:SSL routines:ssl3_read_bytes:tlsv1 alert internal error:ssl/record/rec_layer_s3.c:909:SSL alert number 80
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 7 bytes and written 247 bytes
Verification: OK
---
New, (NONE), Cipher is (NONE)
This TLS version forbids renegotiation.
Compression: NONE
Expansion: NONE
No ALPN negotiated
Early data was not sent
Verify return code: 0 (ok)
---
Server output:
spawn openssl s_server -accept 23456 -naccept 1 -key pkcs11:type=private;id=%00%02 -cert pkcs11:type=cert;object=ecCert
ACCEPT
ERROR
80F22F2D037F0000:error:0A00013A:SSL routines:tls_parse_ctos_key_share:unable to find ecdh parameters:ssl/statem/extensions_srvr.c:685:
shutting down SSL
CONNECTION CLOSED
   0 items in the session cache
   0 client connects (SSL_connect())
   0 client renegotiates (SSL_connect())
   0 client connects that finished
   1 server accepts (SSL_accept())
   0 server renegotiates (SSL_accept())
   0 server accepts that finished
   0 session cache hits
   0 session cache misses
   0 session cache timeouts
   0 callback cache hits
   0 cache full overflows (128 allowed)
send: spawn id exp3 not open
    while executing
"send " NO SESSION PARAMETERS \n""
    invoked from within
"expect {
            "END SSL SESSION PARAMETERS" {};
            default {
                send " NO SESSION PARAMETERS \n";
                exit 1;
..."

```

PR #435, opened 2024-08-26
Allow fallback to pulling cert when checking private/public key consistency

### LINKED ISSUE #430: Fallback to fetching public key from certificate not always applied?
(moving this from a comment on a closed issue to a new thread)
Request: Make the fallback to fetching public key from certificate work with OpenSSL's SSL_CTX_use_Privatekey().

This fallback was apparently implemented by #129 but that does not seem to work when an EC private key is passed to `SSL_CTX_use_Privatekey()` in OpenSSL even if matching certificate is present with same ID. In this case `p11prov_ec_match()` gets called which fails to find the public key via `find_associated_obj()`. Would be nice to get the fallback applied here too for wider usability. 

Whereas, when signing using "openssl pkeyutl" the fallback does work and appears to follow a different code path.

Tested using git master branch and a soft token (softhsm2) where only private key (EC) and certificate were imported into the token using `p11tool` -- public key was not explicitly imported. The same token works with `pkcs11-helper` library. 

Unlike `p11tool`, `softhsm2-util` imports keypair in one step and creates both public and private key objects in the token --- in that case there are no such errors, obviously.

simo5:
#### Description

This behavior is needed to pass some openssl consistency checks with tokens that store the public key only in a cert object.

Fixes #430 

#### Checklist

- ~[ ] Documentation updated~

#### Reviewer's checklist:

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -2279,6 +2279,81 @@ static int cmp_public_key_values(P11PROV_OBJ *pub_key1, P11PROV_OBJ *pub_key2)
     return ret;
 }
 
+static int match_key_with_cert(P11PROV_OBJ *priv_key, P11PROV_OBJ *pub_key)
+{
+    P11PROV_OBJ *cert;
+    CK_ATTRIBUTE attrs[2] = { 0 };
+    CK_ATTRIBUTE *x;
+    int num = 0;
+    int ret = RET_OSSL_ERR;
+
+    cert = find_associated_obj(priv_key->ctx, priv_key, CKO_CERTIFICATE);
+    if (!cert) {
+        P11PROV_raise(priv_key->ctx, CKR_GENERAL_ERROR,
+                      "Could not find associated certificate object");
+        return RET_OSSL_ERR;
+    }
+
+    switch (pub_key->data.key.type) {
+    case CKK_RSA:
+        attrs[0].type = CKA_MODULUS;
+        attrs[1].type = CKA_PUBLIC_EXPONENT;
+        num = 2;
+        break;
+    case CKK_EC:
+    case CKK_EC_EDWARDS:
+        attrs[0].type = CKA_P11PROV_PUB_KEY;
+        num = 1;
+        break;
+    }
+
+    ret = get_all_from_cert(cert, attrs, num);
+    if (ret != CKR_OK) {
+        P11PROV_raise(priv_key->ctx, ret,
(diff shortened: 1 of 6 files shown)


src/objects.c
```diff
@@ -2313,7 +2388,10 @@ static int match_public_keys(P11PROV_OBJ *key1, P11PROV_OBJ *key2)
     if (!assoc_pub_key) {
         P11PROV_raise(priv_key->ctx, CKR_GENERAL_ERROR,
                       "Could not find associated public key object");
-        return RET_OSSL_ERR;
+
+        /* some tokens only store the public key o a cert and not in a separate
```
Jakuje, line 2392:
```suggestion
        /* some tokens only store the public key in a cert and not in a separate
```
?

tests/tlssetkey.c
```diff
@@ -0,0 +1,53 @@
+/* Copyright (C) 2022 Simo Sorce <simo@redhat.com>
```
Jakuje, line 1:
```suggestion
/* Copyright (C) 2024 Simo Sorce <simo@redhat.com>
```

tests/ttls
```diff
@@ -7,6 +7,12 @@ source "${TESTSSRCDIR}/helpers.sh"
 title PARA "Test SSL_CTX creation"
 $CHECKER "${TESTBLDDIR}/tlsctx"
 
+if [ -n "$ECBASE2URI" ]; then
+    title PARA "Test setting cert/keys on TLS Context"
+    $CHECKER "${TESTBLDDIR}/tlssetkey" "${ECCRT2URI}" "${ECPRI2URI}"
+fi
```
Jakuje, line 13: would it make sense to add a sanity test with full keys? Either in else branch or just with normal keys? And with RSA keys?

simo5: RSA keys always have the public exponent so we'd be testing nothing really, not sure we need any other test really, this specific behavior with EC keys is the corner case that needs testing, other fallback behaviors are already tested independently as needed, and the general behavior is also already working.

simo5:
Ok added test with URIs for private key that has the corresponding pub key too, it cost us nothing after all.

simo5:
@Jakuje given you told me (privately) that there were no makor issues with this PR, all test pass and cosvcan is happy, I am going to merge.

I need this in for my next PR.

PR #436, opened 2024-08-26
Refactor setup

simo5:
Move only the token specific initialization into custom setup files, and the use pkcs11-tool and a single script to set up all the keys and certificates.

#### Description

The main work of setting up tokens is now in test/setup.sh and each token type has a <name>-init.sh file that does the bare minimum custom initialization needed.

#### Checklist

- ~[ ] Code modified for feature~
- ~[ ] Test suite updated with negative tests~
- ~[ ] Documentation updated~

#### Reviewer's checklist:

- ~[ ] Any issues marked for closing are addressed~
- ~[ ] Coverity Scan has run if needed (code PR) and no new defects were found~


simo5:
NOTE: Kryoptic stuff will fail until https://github.com/latchset/kryoptic/pull/84 is merged

simo5:
NOTE: NSS fails on Centos9 because it does not support NSS_LIB_PARAMS yet I think ...

tests/setup.sh
```diff
@@ -1,15 +1,51 @@
-#!/bin/bash -e
-# Copyright (C) 2024 Simo Sorce <simo@redhat.com>
+#!/bin/bash -ex
+# Copyright (C) 2024 Simo Sorcen <simo@redhat.com>
```
Jakuje, line 2:
```suggestion
# Copyright (C) 2024 Simo Sorce <simo@redhat.com>
```

tests/setup.sh
```diff
@@ -1,15 +1,51 @@
...
+if [ -d "${TMPPDIR}" ]; then
+    rm -fr "${TMPPDIR}"
+fi
+mkdir "${TMPPDIR}"
+
+# Token data dir
+TOKDIR="$TMPPDIR/tokens"
+if [ -d "${TOKDIR}" ]; then
+    rm -fr "${TOKDIR}"
+fi
```
Jakuje, line 24: Is this check needed when the tmpdir is removed and freshly created few lines above anyway?

simo5: ha no, I just copied the block without questioning it :)

simo5:
@Jakuje what should we do with NSS on CentOS 9 ?
Test the version and skip it if it is not new enough ?

Jakuje:
> @Jakuje what should we do with NSS on CentOS 9 ? Test the version and skip it if it is not new enough ?

Probably yes. I think CentOS 9 should get the NSS update any time soon (I thought this already some months back though).

simo5:
@Jakuje Best I could do is to make the C9s test "pass" without actually doing anything if the NSS version is too old.

.github/workflows/build.yml
```diff
@@ -37,31 +37,45 @@ jobs:
...
-                diffutils expect valgrind
+                diffutils expect valgrind opensc gnutls-utils
               if [ "${{ matrix.token }}" = "softokn" ]; then
-                dnf -y install nss-softokn nss-tools nss-softokn-devel
+                dnf -y install nss-softokn nss-tools nss-softokn-devel \
+                  nss-devel
               elif [ "${{ matrix.token }}" = "softhsm" ]; then
-                dnf -y install softhsm opensc p11-kit-devel p11-kit-server \
-                  gnutls-utils
+                dnf -y install softhsm p11-kit-devel p11-kit-server
```
Jakuje, line 45: once we touch it, we could probably remove the `p11-kit-server` from this line as we no longer use the p11-kit remoting. We should probably remove it from the `packaging/pkcs11-provider.spec` too.

tests/softokn-init.sh
```diff
@@ -0,0 +1,27 @@
...
+certutil -N -d "${TOKDIR}" -f "${PINFILE}"
+
+export P11LIB="${SOFTOKNPATH%%/}/libsoftokn3${SHARED_EXT}"
+
+# NSS hardcodes the Token Label ...
+export TOKENLABEL="NSS Certificate DB"
+export TOKENLABELURI="NSS%20Certificate%20DB"
+
+# Older NSS version need a custom init arg
+#TOKENOPTIONS="pkcs11-module-init-args = configDir=${TOKDIR}"
```
Jakuje, line 23: Would not it be easier to enable this line for old NSS instead of doing the skipping in the github workflow?

simo5: There is no way to pass the argument to pkcs11-tool afaik, so it wouldn't help, I will just remove that comment.

.github/workflows/build.yml
simo5: ok

simo5: should we also remove the whole softhsm-proxy.sh thing ?

tests/softokn-init.sh
Jakuje: Indeed. My bad.

.github/workflows/build.yml
Jakuje: Yes. I think it only confuses things. But note that the p11-kit itself should stay around for the default proxy module.

simo5: Done, PTAL

tests/setup.sh
```diff
@@ -1,15 +1,46 @@
...
 # SPDX-License-Identifier: Apache-2.0
 
 source "${TESTSSRCDIR}/helpers.sh"
 
-if ! command -v p11tool &> /dev/null
-then
-    echo "GnuTLS utils are required"
-    exit 0
+if [ $# -ne 1 ]; then
+    echo "Usage setup.sh <tokentype>"
```
Jakuje, line 8: missing `exit 1` here.

Jakuje (dismissed):
one last comment, otherwise consider approved.

tests/setup.sh
simo5: fixed, thanks!

simo5:
Thanks for the review @Jakuje

PR #453, opened 2024-10-18
tests: Run more TLS tests when forcing all server operations on token

Jakuje:
#### Description

This is rebase of previously closed PR #427, which adds also a test reproducer for #449.

This is trying to solve few issues:
 * The standard error is not recorded in the log files, which makes them hard to use for debugging
 * The expect script `default` does not catch `eof`, which makes some failures hidden
 * One of these failures happens in the client, when the provider is forced to be used, but it dies on verification failure. This is because the SSL/TLS code picks up errors left on stack that were not cleaned in several occasions
 * The error from reproducer from #449 was also being hidden because the server closed the connection.
 * The TLS tests when the provider is forced to be used were executing only one basic use case. This again attempts to run all of them.
 * The query for ALLOWED_MECHANISM during signature was leaving errors on the stack with softokn. I added the `no-allowed-mechanisms` for the softokn tests and documented this undocumented quirk
 * When doing operations on imported keys, we had undefined slot id, which caused the signature verification failures in client.

#### Checklist

- [X] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -107,6 +107,9 @@ void p11prov_debug_mechanism(P11PROV_CTX *ctx, CK_SLOT_ID slotid,
         }
     }
 
+    /* set error mark so we can clear spurious errors */
+    p11prov_set_error_mark(ctx);
+
     ret = p11prov_GetMechanismInfo(ctx, slotid, type, &info);
     if (ret != CKR_OK) {
         p11prov_debug(NULL, 0, NULL,
@@ -129,6 +132,8 @@ void p11prov_debug_mechanism(P11PROV_CTX *ctx, CK_SLOT_ID slotid,
             }
         }
     }
+    /* if there was any error, remove it, this is just a debug function */
+    p11prov_pop_error_to_mark(ctx);
 }
 
 extern struct ckmap token_flags[];
--- a/src/signature.c
+++ b/src/signature.c
@@ -430,9 +430,25 @@ static CK_RSA_PKCS_MGF_TYPE p11prov_sig_map_mgf(const char *digest_name)
 static CK_RV p11prov_sig_pss_restrictions(P11PROV_SIG_CTX *sigctx,
                                           CK_MECHANISM *mechanism)
 {
-    CK_ATTRIBUTE *allowed_mechs =
-        p11prov_obj_get_attr(sigctx->key, CKA_ALLOWED_MECHANISMS);
+    CK_BBOOL token_supports_allowed_mechs = CK_TRUE;
+    CK_ATTRIBUTE *allowed_mechs = NULL;
+    CK_RV ret;
 
+    /* check if token supports CKA_ALLOWED_MECHANISMS at all */
+    ret = p11prov_token_sup_attr(
+        sigctx->provctx, p11prov_obj_get_slotid(sigctx->key), GET_ATTR,
+        CKA_ALLOWED_MECHANISMS, &token_supports_allowed_mechs);
+    if (ret != CKR_OK) {
+        P11PROV_raise(sigctx->provctx, ret,
+                      "Failed to probe CKA_ALLOWED_MECHANISMS quirk");
+        return ret;
+    }
+    if (token_supports_allowed_mechs == CK_FALSE) {
+        /* Token does not support CKA_ALLOWED_MECHANISMS so there are no restrictions */
+        return CKR_OK;
+    }
+
+    allowed_mechs = p11prov_obj_get_attr(sigctx->key, CKA_ALLOWED_MECHANISMS);
     if (allowed_mechs) {
         CK_ATTRIBUTE_TYPE *mechs = (CK_ATTRIBUTE_TYPE *)allowed_mechs->pValue;
         int num_mechs = allowed_mechs->ulValueLen / sizeof(CK_MECHANISM_TYPE);
--- a/src/signature.c
+++ b/src/signature.c
@@ -1484,6 +1484,16 @@ static int p11prov_rsasig_set_ctx_params(void *ctx, const OSSL_PARAM params[])
     p = OSSL_PARAM_locate_const(params, OSSL_SIGNATURE_PARAM_PAD_MODE);
     if (p) {
         CK_MECHANISM_TYPE mechtype = CK_UNAVAILABLE_INFORMATION;
+        CK_SLOT_ID slotid = p11prov_obj_get_slotid(sigctx->key);
+
+        /* If the object is imported, use the default slot */
+        if (slotid == CK_UNAVAILABLE_INFORMATION) {
+            P11PROV_SLOTS_CTX *slots = p11prov_ctx_get_slots(sigctx->provctx);
+            if (!slots) {
+                return RET_OSSL_ERR;
+            }
+            slotid = p11prov_get_default_slot(slots);
+        }
         if (p->data_type == OSSL_PARAM_INTEGER) {
             int pad_mode;
             /* legacy pad mode number */
@@ -1525,8 +1535,7 @@ static int p11prov_rsasig_set_ctx_params(void *ctx, const OSSL_PARAM params[])
              * regardless, and this is not the case in PKCS#11 */
             CK_RV rv;
 
-            rv = p11prov_check_mechanism(sigctx->provctx,
-                                         p11prov_obj_get_slotid(sigctx->key),
+            rv = p11prov_check_mechanism(sigctx->provctx, slotid,
                                          CKM_RSA_PKCS_PSS);
             if (rv != CKR_OK) {
                 P11PROV_raise(sigctx->provctx, rv,
@@ -1537,9 +1546,7 @@ static int p11prov_rsasig_set_ctx_params(void *ctx, const OSSL_PARAM params[])
 
         sigctx->mechtype = mechtype;
 
-        P11PROV_debug_mechanism(sigctx->provctx,
-                                p11prov_obj_get_slotid(sigctx->key),
-                                sigctx->mechtype);
+        P11PROV_debug_mechanism(sigctx->provctx, slotid, sigctx->mechtype);
     }
 
     p = OSSL_PARAM_locate_const(params, OSSL_SIGNATURE_PARAM_PSS_SALTLEN);
--- a/tests/ttls
+++ b/tests/ttls
@@ -25,7 +25,7 @@ SERVER_PID=-1
 # Make sure we terminate programs if test fails in the middle
 # shellcheck disable=SC2317  # Shellcheck for some reason does not follow trap
 wait_for_server_at_exit() {
-    wait "$1"
+    wait "$1" || :
     echo "Server output:"
     cat "${TMPPDIR}/s_server_output"
 }
--- a/tests/ttls
+++ b/tests/ttls
@@ -112,6 +112,9 @@ run_tests() {
 
     title PARA "Run test with TLS 1.2 and ECDH"
     run_test "$ECPRIURI" "$ECCRTURI" "" "-tls1_2 -cipher ECDHE-ECDSA-AES128-GCM-SHA256 -groups secp256r1"
+
+    title PARA "Run test with TLS 1.3 and specific suite"
+    run_test "$ECPRIURI" "$ECCRTURI" "" "-tls1_3 -ciphersuites TLS_AES_256_GCM_SHA384 -groups secp256r1"
 }
 
 title SECTION "TLS with key in provider"
--- a/tests/ttls
+++ b/tests/ttls
@@ -42,6 +42,7 @@ run_test() {
         set timeout 60;
         expect {
             \"ACCEPT\" {};
+            eof { exit 2; }
             default {
                 send \" NO ACCEPT \n\";
                 exit 1;
@@ -52,6 +53,7 @@ run_test() {
         close \$server_ready;
         expect {
             \"END SSL SESSION PARAMETERS\" {};
+            eof { exit 2; }
             default {
                 send \" NO SESSION PARAMETERS \n\";
                 exit 1;
@@ -74,6 +76,7 @@ run_test() {
         set timeout 60;
         expect {
             \" TLS SUCCESSFUL \" {};
+            eof { exit 2; }
             default {
                 send \" NO TLS SUCCESSFUL MESSAGE \n\";
                 exit 1;
--- a/tests/ttls
+++ b/tests/ttls
@@ -88,7 +88,7 @@ run_test() {
                 send \" NO EOF \n\";
                 exit 1;
             };
-        }"
+        }" || (wait_for_server_at_exit $SERVER_PID; exit 1; )
 
     wait_for_server_at_exit $SERVER_PID
 }
--- a/tests/ttls
+++ b/tests/ttls
@@ -67,7 +67,7 @@ run_test() {
                 send \" NO EOF \n\";
                 exit 1;
             };
-        }" > "${TMPPDIR}/s_server_output" &
+        }" &> "${TMPPDIR}/s_server_output" &
     SERVER_PID=$!
 
     read -r < "${TMPPDIR}/s_server_ready"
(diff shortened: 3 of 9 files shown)


Jakuje:
ugh. In the CI it passes now, but locally it failed for me previously. After proper clean it works for me too locally.

tests/ttls
```diff
@@ -70,6 +70,9 @@ run_test() {
 
     read -r < "${TMPPDIR}/s_server_ready"
 
+    # FIXME For now we strive to test the server. The client causes some unrelated issues that will have to
+    # be handled later
+    OPENSSL_CONF="${ORIG_OPENSSL_CONF}" \
```
simo5, line 77: I am not sure we want this, we want to be able to test forcing both server an client.
If we do this then we'd have to add a test matrix where we try all combinations of forcing server and client.

tests/ttls
```diff
@@ -91,7 +91,7 @@ run_test() {
                 send \" NO EOF \n\";
                 exit 1;
             };
-        }"
+        }" || (wait_for_server_at_exit $SERVER_PID;  exit 1; )
 
     wait_for_server_at_exit $SERVER_PID
```
simo5, line 96: Isn't this redundant now?

simo5 (commented):
Largely LGTM, some comments.

That said maybe remove the last commit so we can see that CI fails, then I will merge the original PR, and then you can rebase on main and we ensure everything still works.

tests/ttls
Jakuje: I think this makes sure we wait for the server exit on success so I think it is still needed.

In case of the expect script fails, this line would not be executed and the script would terminate right away not waiting for the server and not giving the server logs.

tests/ttls
Jakuje: Probably yes. But it does not work without this, even the basic case fails now with:
```
########################################
## Forcing the provider for all server operations


## Run sanity test with default values (RSA)
spawn openssl s_client -connect localhost:23456 -CAfile /home/jjelen/devel/pkcs11-provider/builddir/tests/softokn/CAcert.pem
Connecting to ::1
CONNECTED(00000004)
Can't use SSL_get_servername
depth=1 CN=Issuer
verify return:1
depth=0 O=PKCS11 Provider, CN=My Test Cert
verify return:1
80D25E0D127F0000:error:40800070:pkcs11:p11prov_rsasig_set_ctx_params:An invalid mechanism was specified to the cryptographic operation:../src/signature.c:1532:CKM_RSA_PKCS_PSS unavailable
80D25E0D127F0000:error:0A080006:SSL routines:tls_process_cert_verify:EVP lib:ssl/statem/statem_lib.c:560:
---
Certificate chain
 0 s:O=PKCS11 Provider, CN=My Test Cert
   i:CN=Issuer
   a:PKEY: rsaEncryption, 2048 (bit); sigalg: RSA-SHA256
   v:NotBefore: Oct 21 07:58:11 2024 GMT; NotAfter: Oct 21 07:58:11 2025 GMT
---
Server certificate
-----BEGIN CERTIFICATE-----
MIIDPzCCAiegAwIBAgIBAzANBgkqhkiG9w0BAQsFADARMQ8wDQYDVQQDEwZJc3N1
ZXIwHhcNMjQxMDIxMDc1ODExWhcNMjUxMDIxMDc1ODExWjAxMRgwFgYDVQQKEw9Q
S0NTMTEgUHJvdmlkZXIxFTATBgNVBAMTDE15IFRlc3QgQ2VydDCCASIwDQYJKoZI
hvcNAQEBBQADggEPADCCAQoCggEBAMyqHVpQB9mfrimSiIdV34EU8FyVc+o1SIRq
jRsoo3F93VLhhaOpje1EQ3I0bDOKSr8fT+pjReHENkOycnbzw2eE6eAeaDlNYU51
MRNp0eZY0/K9WRZDppkc+i/Z9G5jj+wyZgNSUS2EmYvz9vTXcdG8cPLdyTBm2/U3
JIQDZPJ9UmavFa0Ojr2MiDDTY/dWxXPUIoMf4dQ/ppKa4e++XUteh9LvONqMY6RU
P9JhkbXGGmQiLyVVkD3pwSDZG7M3g3AdDl4BQhU5/v/G0ETHWbKmKnzpDCJ2FJah
VBWx0jZgbwNiGQ9j8shs12lEX5YMezDu/NCE3ZgaVORDwdtbpWUCAwEAAaOBgTB/
MAwGA1UdEwEB/wQCMAAwHwYDVR0RBBgwFoEUdGVzdGNlcnRAZXhhbXBsZS5vcmcw
DgYDVR0PAQH/BAQDAgWgMB0GA1UdDgQWBBSP4+rrZTwRXWi4OdqhdTseOJC2FTAf
BgNVHSMEGDAWgBRgYl8qthze3R6sKMpf02knjzgH9DANBgkqhkiG9w0BAQsFAAOC
AQEAA9qDTCeoOQox2oOkq/XlIek4XcCMjvCkiJdGmuPkDLit4sExalldkMPkds5W
n+BMAYObkPK7QmuZVXWtQYnMSm62N2ejNdSD+Dc/uUEp8KkznFhC6wsuScrjEwOh
oacfvKUS8BslUhRrTgDGQ9cvvI4gjHRzkQmNagmoR1vkbZGJGcQE9tte1WD+OtCZ
9t1uXr6q2hvb30jlsb/EUbfLYBEvC0sTyQ/n/ucmCcTg9GSBy/IcaiWYxng2grNl
Ox6zIE21D/5PUtD+nZsakdDZPuxAe/nfbys8yb5pPMGJ5mZQfUgdxvkO0+iGBtTP
pJrlJI/9/0yU94guAgco1Nwdbg==
-----END CERTIFICATE-----
subject=O=PKCS11 Provider, CN=My Test Cert
issuer=CN=Issuer
---
No client certificate CA names sent
Peer signing digest: SHA256
Peer signature type: RSA-PSS
Server Temp Key: X25519, 253 bits
---
SSL handshake has read 1317 bytes and written 321 bytes
Verification: OK
---
New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384
Server public key is 2048 bit
This TLS version forbids renegotiation.
Compression: NONE
Expansion: NONE
No ALPN negotiated
Early data was not sent
Verify return code: 0 (ok)
---
/home/jjelen/devel/pkcs11-provider/tests/ttls: line 28: wait: pid 380444 is not a child of this shell
Server output:
spawn openssl s_server -accept 23456 -naccept 1 -key pkcs11:type=private;id=%00%01 -cert pkcs11:type=cert;object=testCert
ACCEPT
ERROR
80D23BA1F87F0000:error:40800091:pkcs11:p11prov_GetOperationState:There is no active operation of appropriate type in the specified session:../src/interface.gen.c:335:Error returned by C_GetOperationState
80D23BA1F87F0000:error:0A000438:SSL routines:ssl3_read_bytes:tlsv1 alert internal error:ssl/record/rec_layer_s3.c:909:SSL alert number 80
shutting down SSL
CONNECTION CLOSED
   0 items in the session cache
   0 client connects (SSL_connect())
   0 client renegotiates (SSL_connect())
   0 client connects that finished
   1 server accepts (SSL_accept())
   0 server renegotiates (SSL_accept())
   0 server accepts that finished
   0 session cache hits
   0 session cache misses
   0 session cache timeouts
   0 callback cache hits
   0 cache full overflows (128 allowed)
```
I am not sure which of the failures is the problem, but I assume the mechanism issue? The debug log shows the following issues:
```
[../src/interface.gen.c:950] p11prov_VerifyFinal(): Calling C_VerifyFinal
[../src/objects.c:435] p11prov_obj_free(): Free Object: 0x555a26466fa0 (handle:1)
[../src/objects.c:442] p11prov_obj_free(): object free: reference held
[../src/objects.c:435] p11prov_obj_free(): Free Object: 0x555a26466fa0 (handle:1)
[../src/objects.c:442] p11prov_obj_free(): object free: reference held
[../src/keymgmt.c:713] p11prov_rsa_has(): rsa has 0x555a264a76d0 4
[../src/signature.c:1257] p11prov_rsasig_digest_verify_init(): rsa digest verify init (ctx=0x555a264b9ee0, key=0x555a264a76d0, params=(nil))
[../src/objects.c:404] p11prov_obj_ref_no_cache(): Ref Object: 0x555a264a76d0 (handle:18446744073709551614)
[../src/provider.c:607] p11prov_ctx_cache_keys(): cache_keys = 1
[../src/signature.c:1460] p11prov_rsasig_set_ctx_params(): rsasig set ctx params (ctx=0x555a264b9ee0, params=(nil))
[../src/signature.c:1460] p11prov_rsasig_set_ctx_params(): rsasig set ctx params (ctx=0x555a264b9ee0, params=0x7ffd3e343650)
[../src/signature.c:1494] p11prov_rsasig_set_ctx_params(): Set OSSL_SIGNATURE_PARAM_PAD_MODE to 6
[../src/signature.c:1532] p11prov_rsasig_set_ctx_params(): Error: 0x00000070; CKM_RSA_PKCS_PSS unavailable
```
This error condition has the following code comment:
https://github.com/latchset/pkcs11-provider/blob/c7fb177ab9f3e797f5a8c89e45e259ca74cd749b/src/signature.c#L1521-L1525

Jakuje:
Pushed without the commit from #449 to demonstrate the failure.

tests/ttls
simo5: The GetOperationState failure is suspicious, sounds like we are simply not raising on the correct error, and that's the only thing left on the stack.

So the problem here is that OpenSSL is trying to do RSA PSS, can we change the options to force ECDSA and see if that makes it work?

Also all three tokens should support PSS, is this failing with on or all of them ?

Jakuje: I did not try kryoptic, but the same test fails on both softhsm and softokn. If I switch to the second test (ecdsa), it fails with the test same way, but I do not see any obvious error except for the operation state error.

Jakuje:
It looks like I finally see what is going on there. The check on 
https://github.com/latchset/pkcs11-provider/blob/c7fb177ab9f3e797f5a8c89e45e259ca74cd749b/src/signature.c#L1528-L1530
checks if the mechanism is available on the given slot the key is using, but the key is imported so it has slot `-1` (`CK_UNAVAILABLE_INFORMATION`) which leads to conclusion the RSA-PSS mechanism is not supported by a slot. I tried to work around it, but I assume there will be a better way.

The fix in #449 changes the handle of the imported EC keys so this does not solve the client issue with imported ECDH keys so it is usable in both client and server operations.

But if I see right, it still fails with generic errors, which look like coming from OpenSSL during signature verification, which went through ok. But it looks like the OpenSSL itself is exploring the error stack and if it finds some error, it fails anyway:
```
80A20FED9A7F0000:error:40800012:pkcs11:p11prov_GetAttributeValue:Invalid attribute type specified in a template:../src/interface.gen.c:488:Error returned by C_GetAttributeValue^M
...
Verification error: unspecified certificate verification error^M
```
This leads me to the question https://github.com/openssl/openssl/discussions/23025 which points out that we should really clear these bogus/outdated/non-relevant-anymore messages from the error stack, as the OpenSSL (SSL/TLS API) really cares for the queue and if there is something left, it considers it as an error (or just make this error less verbose so it wont make it to the error queue -- likely not -- this is generic handling of the pkcs11 calls so I would prefer to keep it as it is).

The same issue has the `C_GetOperationState()` call -- it pushes the error on the openssl stack, which is than interpreted as an error during the TLS exchange and makes the exchange failed. I worked around this by using `no-operation-state` quirk in this PR, but we should either enable it by default for the pkcs11 tokens we know that are broken, again flush the error message after handling the error or just downgrade it to debug log.

With clearing the error stack, I can make all the tests pass (locally -- in CI it still fails for some reason I need to investigate) on both client and server.

Any comments/insights welcomed.

src/debug.c
```diff
@@ -129,6 +129,8 @@ void p11prov_debug_mechanism(P11PROV_CTX *ctx, CK_SLOT_ID slotid,
             }
         }
     }
+    /* XXX Not fatal */
+    ERR_clear_error();
```
simo5, line 133: I am confused by this clear here.

Generally we should set a new error stack up when we enter a function exposed to OpenSSL, and then pop/clear it if we need to suppress any error.

This clearing seem to happen in the middle of a call to a function that is not aware of what error stack is available.

If we want to always prevent any errors from showing up from this function I think we should set up a stack pointer at the start and just pop it at the end, otherwise we risk clearing previous errors that we did want to report.

See the functions: p11prov_set_error_mark(), p11prov_pop_error_to_mark(), p11prov_pop_error_to_mark() and p11prov_clear_last_error_mark()
An example of their use is in store_fetch() and should be extended to other places as needed.

src/signature.c
```diff
@@ -1526,9 +1527,9 @@ static int p11prov_rsasig_set_ctx_params(void *ctx, const OSSL_PARAM params[])
             CK_RV rv;
 
             rv = p11prov_check_mechanism(sigctx->provctx,
-                                         p11prov_obj_get_slotid(sigctx->key),
+                                         slotid,
                                          CKM_RSA_PKCS_PSS);
-            if (rv != CKR_OK) {
+            if (rv != CKR_OK && slotid != CK_UNAVAILABLE_INFORMATION) {
```
simo5, line 1532: While I can see why you are doing this I think we should take a different approach.
After line 1487 you should immediately check is slotid is valid, and if not fallback to pull the slotid of the login session which is the session we use to store the key if we ever need to perform an operation.

This way the check on the mechanism will be always done on the correct slot.

simo5: Use
```
slotid = p11prov_get_default_slot(slots);
```
in the fallback and fail the operation if no default slot is available..

src/util.c
```diff
@@ -69,6 +69,9 @@ CK_RV p11prov_fetch_attributes(P11PROV_CTX *ctx, P11PROV_SESSION *session,
     } else if (attrnums > 1
                && (ret == CKR_ATTRIBUTE_SENSITIVE
                    || ret == CKR_ATTRIBUTE_TYPE_INVALID)) {
+        /* The previous PKCS#11 call failed, which populated the OpenSSL error queue.
+         * But we working around this error so just flush the error now. */
+        ERR_clear_error();
```
simo5, line 74: See the other comment for error stack management

simo5 (changes requested):
Sounds like we are finally on the right path.
It should be relatively easy to address the 2 main comments (slot selection and error stack mgmt) I highlighted.

src/debug.c
Jakuje: Thank you for the pointer. I did not  see these. Indeed, these are the functions I was looking for!

src/signature.c
Jakuje: Thank you! Yes, this was mostly hackish solution to prove the code path is causing the issue not meant for final solution. Using login/default slot makes the most sense.

src/util.c
Jakuje: We are hitting the issue here in the corner case for softokn, where we ask for one attribute `CKA_ALLOWED_MECHANISMS`, so we do not fall back to the fetching of attributes per partes and stack resetting on non-required attributes:
```
56: C_GetAttributeValue^M
P:683169; T:0x139804941566592 2024-10-23 15:45:49.074^M
[in] hSession = 0x1000004^M
[in] hObject = 0x9a81bced^M
[in] pTemplate[1]: ^M     
    CKA_ALLOWED_MECHANISMS  0000000000000000 / 0^M
[out] pTemplate[1]: ^M    
    CKA_ALLOWED_MECHANISMS  0000000000000000 / -1^M
Returned:  18 CKR_ATTRIBUTE_TYPE_INVALID^M
```
Not sure if current solution is best, but it also works.

simo5: This is so annoying,
maybe we should have a quirk to turn off using CKA_ALLOWED_MECHANISM and conditionalize the template ?

simo5: Actually we already have a quirk, so we should just use it here to decide whether we want to add CKA_ALLOWED_MECHANISMS to the template, see p11prov_rsapss_gen() for an example.

Jakuje: We have just the trial-error quirk `token_supports_allowed_mechs` in `p11prov_rsapss_gen()` which does not extend outside of this function context. I will probably introduce a new configuration and turn this off for softokn altogether.
For the case in signature.c we can not use trial-error as it is the only attribute we have in the template.

Jakuje: oh, the quirk is implemented, but not documented, which confused me.

simo5 (approved):

Issue #455, opened 2024-10-24
Check PIN lockout status


simo5:
**Describe the feature**
The PKCS#11 specification allows a token to expose flags that indicate if the PIN authentication failure count is too high and might soon get locked.
pkcs11-provider should consult these flags on authentication and if the flag CKF_USER_PIN_FINAL_TRY is returned, refuse to attempt a login, and require the user to reset the counter by performing a manual unlock once via tools like pkcs11-tool.

**Expected behavior**
Never lockup a token by checking token flags before login.

**Additional context**
Might not work if tokens do not report the flags correctly.
There is also a TOCTOU problem if two threads concurrently try to login, but better than nothing.

PR #457, opened 2024-10-24
Add code to prevent locking the token by mistake

simo5:
#### Description

For tokens that properly report the status of the PIN authentication counter via token flags, check them out and refuse to attempt login if the token is on its last try.

A token should never be on its last try and finding this flags set is an indication that someone may have hardocded an in correct pin in the configuration or an URI. Proceeding would have a high chance of ending up blocking the token.

Fixes: #455

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -140,6 +140,7 @@ tests = {
   'uri': {'suites': ['softokn', 'softhsm', 'kryoptic']},
   'ecxc': {'suites': ['softhsm', 'kryoptic']},
   'cms': {'suites': ['softokn', 'kryoptic']},
+  'pinlock': {'suites': ['kryoptic']},
 }
 
 test_wrapper = find_program('test-wrapper')
--- a/tests/tpinlock
+++ b/tests/tpinlock
@@ -1,0 +1,81 @@
+#!/bin/bash -e
+# Copyright (C) 2024 Simo Sorce <simo@redhat.com>
+# SPDX-License-Identifier: Apache-2.0
+
+source "${TESTSSRCDIR}/helpers.sh"
+
+title PARA "Test PIN lock prevention"
+
+ORIG_OPENSSL_CONF=${OPENSSL_CONF}
+sed "s/^pkcs11-module-token-pin.*$/##nopin/" "${OPENSSL_CONF}" > "${OPENSSL_CONF}.nopin"
+OPENSSL_CONF=${OPENSSL_CONF}.nopin
+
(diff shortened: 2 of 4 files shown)


Jakuje (dismissed):
lgtm

simo5:
@Jakuje I have not added tests for this feature, do you think it is obvious enough, or do you think I should try to create a test? I might be able to do it with kryoptic at least.

Jakuje:
I think having a test coverage for this would be helpful to be sure it works as expected.

Just something simple trying to login several times using pkcs11 provider and bad pin (different processes so this is not caught by the mechanism preventing relogin with bad pin), then checking it is not locked (for example using pkcs11-tool), then using the right pin in pkcs11 tool and then verifying the functionality again with pkcs11 provider.

simo5:
Added a test!

tests/tpinlock
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash -e
+# Copyright (C) 2022 Simo Sorce <simo@redhat.com>
```
Jakuje, line 2: 2024

tests/meson.build
```diff
@@ -140,6 +140,7 @@ tests = {
   'uri': {'suites': ['softokn', 'softhsm', 'kryoptic']},
   'ecxc': {'suites': ['softhsm', 'kryoptic']},
   'cms': {'suites': ['softokn', 'kryoptic']},
+  'pinlock': {'suites': ['kryoptic']},
```
Jakuje, line 143: It does not work with others? At all? Or do they have just different amount of failed attempts?

simo5: I did not have time to test :-)

tests/tpinlock
simo5: meh :)

tests/meson.build
simo5: Ok tested now, but both softokn and softhsm fail the tests with an odd error:
error: PKCS11 function C_OpenSession failed: rv = CKR_SLOT_ID_INVALID (0x3)
and never reach final try of course.

Let me investigate if my setup is incorrect in those cases.

Jakuje: Interesting. I am not sure if the softokens support the pin counter though as its more like security through obscurity given that the DB can be always copied or backed up.
If it would be too complicated, just a comment that it does not work would be probably good.

Jakuje (approved):

PR #463, opened 2024-11-06
Fix memory leaks when tokens are missing

### LINKED ISSUE #462: Memory leak in slot.c
**Describe the bug**
On empty slot list whole array P11PROV_SLOTS_CTX::slots is leaked. Also
due to C_GetSlotInfo returning empty slots, memory for them is
allocated, but counter isn't incremented due to `continue` in cycle body thus creating another leak.

Suggested solution is to change P11PROV_SLOTS_CTX::slots from array of
poiners to array of structs to fight the second problem. 
And do not skip
`p11prov_free_slots` when number of present slots is 0 for the first problem.

I have commit with fix on my hands, and ready to present a pull request

simo5:
#### Description

In case we have slots advertized but the driver fails to return information or the token is not present we were leaking memory as the slot is not added to the array and the number of slots is not incremented.

Ensure the slot struct is freed in this case.

Fixes #462 

#### Checklist

- [ ] ~Documentation updated~

#### Reviewer's checklist:

### DIFF
--- a/src/slot.c
+++ b/src/slot.c
@@ -182,18 +182,22 @@ CK_RV p11prov_init_slots(P11PROV_CTX *ctx, P11PROV_SLOTS_CTX **slots)
             ret = CKR_HOST_MEMORY;
             goto done;
         }
-        sctx->slots[sctx->num] = slot;
 
         ret = p11prov_GetSlotInfo(ctx, slotid[i], &slot->slot);
         if (ret != CKR_OK || (slot->slot.flags & CKF_TOKEN_PRESENT) == 0) {
             /* skip slot */
+            OPENSSL_free(slot);
             continue;
         }
         ret = p11prov_GetTokenInfo(ctx, slotid[i], &slot->token);
         if (ret) {
             /* skip slot */
+            OPENSSL_free(slot);
             continue;
         }
+
+        sctx->slots[sctx->num] = slot;
+        sctx->num++;
 
         trim(slot->slot.slotDescription);
         trim(slot->slot.manufacturerID);
@@ -243,8 +247,6 @@ CK_RV p11prov_init_slots(P11PROV_CTX *ctx, P11PROV_SLOTS_CTX **slots)
 
         P11PROV_debug_slot(ctx, slot->id, &slot->slot, &slot->token,
                            slot->mechs, slot->nmechs, slot->profiles);
-
-        sctx->num++;
     }
 
 done:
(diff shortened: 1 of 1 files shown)


src/slot.c
```diff
@@ -187,11 +187,13 @@ CK_RV p11prov_init_slots(P11PROV_CTX *ctx, P11PROV_SLOTS_CTX **slots)
         ret = p11prov_GetSlotInfo(ctx, slotid[i], &slot->slot);
         if (ret != CKR_OK || (slot->slot.flags & CKF_TOKEN_PRESENT) == 0) {
             /* skip slot */
+            OPENSSL_free(slot);
```
Jakuje, line 190: at this moment, the `slot` is already stored in the `sctx->slots[sctx->num]`. Even though the `sctx->num` is incremented only at the end of this cycle, its possible that this could cause use after free. Would it be safer to move the assignment from lime 185 to the end of the cycle too?

Jakuje: The remaining `goto done` conditions should also free this memory, shouldn't it?

simo5: uhmm, yeah this requires a little bit more work than I thought, expect a new push

Jakuje (dismissed):

DEX1999:
And what about leak when you allocate `num`, which equals to numbers of slots in system, elements on string 169 and then if `sctx->num == 0` which is equals to number of slots with tokens you just return without freeing.

latal-1:
You still have a leak due to exiting from `p11prov_free_slots` [before](https://github.com/latchset/pkcs11-provider/blob/e0640320367ad5a2866bf4d77963bd5040d40d1d/src/slot.c#L351) [free](https://github.com/latchset/pkcs11-provider/blob/e0640320367ad5a2866bf4d77963bd5040d40d1d/src/slot.c#L369)

src/slot.c
```diff
@@ -182,19 +182,23 @@ CK_RV p11prov_init_slots(P11PROV_CTX *ctx, P11PROV_SLOTS_CTX **slots)
...
         }
         ret = p11prov_GetTokenInfo(ctx, slotid[i], &slot->token);
         if (ret) {
             /* skip slot */
+            OPENSSL_free(slot);
             continue;
         }
 
+        sctx->slots[sctx->num] = slot;
+        sctx->num++;
```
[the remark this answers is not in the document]
simo5: I prefer not.

simo5:
> > And what about leak when you allocate `num` elements on string 169, which equals to numbers of slots in system, and then if `sctx->num == 0`, which is equals to number of slots with tokens, you just return without freeing.
> 
> It is a corner case not worth optimizing for IMHO. Especially because I have a plan, in the future, to keep slots around that do not have token present as we have an interest in supporting token insertion in the future.

I misread the comment, sorry, pushed another change that drops the early return in the freeing function.

Jakuje (approved):
lgtm

simo5:
ty!

PR #471, opened 2024-11-22
Fix memory leak of ctx_pool.contexts

neverpanic:
#### Description

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2327985

#### Checklist

- [x] Code modified for ~feature~ bug
- [ ] ~Documentation updated~

#### Reviewer's checklist:

### DIFF
--- a/src/provider.c
+++ b/src/provider.c
@@ -219,6 +219,12 @@ static void context_rm_pool(struct p11prov_ctx *ctx)
     }
     if (found) {
         ctx_pool.num--;
+        if (ctx_pool.num == 0) {
+            /* This was the last context, free ctx_pool.contexts to avoid
+             * leaking memory. */
+            OPENSSL_free(ctx_pool.contexts);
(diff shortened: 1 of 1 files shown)


beldmit:
@Jakuje I think it's msan, not asan

Jakuje:
> @Jakuje I think it's msan, not asan

I think the asan has lsan module which should detect memory leaks too:

https://clang.llvm.org/docs/LeakSanitizer.html

But it looks like it needs `ASAN_OPTIONS=detect_leaks=1`, which is not present here. Let me add it in separate PR.

The msan on the other hand requires rebuilds of the whole module including dependencies to avoid false positives if I remember well.

neverpanic:
Are the CI failures expected or a side effect of the free?

simo5:
I broke kryoptic CI, I know how to fix it, we do not need to block on it.

PR #472, opened 2024-11-22
tests: Use LeakSanitizer to catch future memory leaks

Jakuje:
#### Description

Follow-up to #471. The address sanitizer should catch memory leaks, but it needs the configuration, which was missing

https://clang.llvm.org/docs/LeakSanitizer.html

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -115,7 +115,7 @@ jobs:
       - name: Build Kryoptic
         run: |
           cd kryoptic
-          cargo build
+          cargo build --features standard
 
       - name: Setup, Build and Install pkcs11-provider
         run: |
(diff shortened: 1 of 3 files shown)


neverpanic (approved):

simo5:
@jakuje as you are fixing tests, would you mind adding --features standard to kryoptic builds to fix the kryoptic CI as well ?

Jakuje:
> @Jakuje as you are fixing tests, would you mind adding --features standard to kryoptic builds to fix the kryoptic CI as well ?

I started in #473 simplifying the kryoptic build to use dynamic feature, but we need the standard too, which I forgot on first try. Changed now. But it still does not work. Trying now here only the `standard` feature.

simo5:
@Jakuje .github/workflows/integration.yml needs the same treatment

Jakuje:
> @Jakuje .github/workflows/integration.yml needs the same treatment

Added. Missed that one.

PR #479, opened 2024-11-25
Allow generated EVP_PKEY to be used to verify operation

### LINKED ISSUE #480: EVP_PKEY_verify doesn't work with generated EVP_PKEY
**Describe the bug**

`EVP_PKEY_verify` [returns error](https://github.com/latchset/pkcs11-provider/blob/b0cc1ff41b498458c3f4eacfdfc13c8e33af8076/src/signature.c#L702) when using with generated `EVP_PKEY`

**To Reproduce**

```c
EVP_PKEY* keyPair;
EVP_PKEY_CTX *ctx;
unsigned char* data = ...;
size_t dataLen = ...;
unsigned char* signature;
size_t signatureLen;

ctx = EVP_PKEY_CTX_new_from_name(NULL, "RSA", "provider=pkcs11");
EVP_PKEY_keygen_init(ctx);
EVP_PKEY_CTX_set_rsa_keygen_bits(ctx, 1024);
EVP_PKEY_generate(ctx, keyPair);

ctx = EVP_PKEY_CTX_new_from_pkey(NULL, keyPair, "provider=pkcs11");
EVP_PKEY_sign_init(ctx);
EVP_PKEY_sign(ctx, signature, &signatureLen, data, dataLen); // or make C_Sign routine instead

ctx = EVP_PKEY_CTX_new_from_pkey(NULL, keyPair, "provider=pkcs11");
EVP_PKEY_verify_init(ctx);
EVP_PKEY_verify(ctx, signature, signatureLen, data, dataLen); // ERROR
```

**Additional context**

Generated `EVP_PKEY` [contains only `P11PROV_OBJ` private key](https://github.com/latchset/pkcs11-provider/blob/b0cc1ff41b498458c3f4eacfdfc13c8e33af8076/src/keymgmt.c#L493)

latal-1:
#### Description

Previously generated EVP_PKEY could be used for operations that only require a private key

Fixes #480 

#### Checklist

- [ ] ~~Test suite updated with negative tests~~
- [ ] ~~Documentation updated~~

#### Reviewer's checklist:

### DIFF
--- a/src/keymgmt.c
+++ b/src/keymgmt.c
@@ -482,14 +482,19 @@ static int p11prov_common_gen(struct key_generator *ctx,
     }
 
     ret = p11prov_merge_pub_attrs_into_priv(pub_key, priv_key);
+    if (ret != CKR_OK) {
+        goto done;
+    }
+
+    ret = p11prov_set_pub(priv_key, pub_key);
 
 done:
     if (ret != CKR_OK) {
+        p11prov_obj_free(pub_key);
         p11prov_obj_free(priv_key);
-        priv_key = NULL;
+        pub_key = priv_key = NULL;
     }
     p11prov_return_session(session);
-    p11prov_obj_free(pub_key);
     *key = priv_key;
     return ret;
 }
--- a/src/objects.c
+++ b/src/objects.c
@@ -110,8 +110,8 @@ void p11prov_obj_pool_free(P11PROV_OBJ_POOL *pool)
         }
         OPENSSL_free(pool->objects);
         (void)MUTEX_UNLOCK(pool);
-        /* ------------- LOCKED SECTION */ }
-    else {
+        /* ------------- LOCKED SECTION */
+    } else {
         P11PROV_debug("Failed to lock object pool, leaking it!");
         return;
     }
@@ -4030,6 +4030,7 @@ CK_RV p11prov_obj_copy_specific_attr(P11PROV_OBJ *pub_key,
 #define RSA_PRIV_ATTRS_NUM 2
 
 #define EC_PRIV_ATTRS_NUM 3
+// TODO: maybe remove after public key object was added in P11PROV_OBJ
 CK_RV p11prov_merge_pub_attrs_into_priv(P11PROV_OBJ *pub_key,
                                         P11PROV_OBJ *priv_key)
 {
--- a/src/objects.c
+++ b/src/objects.c
@@ -440,6 +440,9 @@ void p11prov_obj_free(P11PROV_OBJ *obj)
     if (obj == NULL) {
         return;
     }
+
+    p11prov_obj_free(obj->pub_key_obj);
+
     if (__atomic_sub_fetch(&obj->refcnt, 1, __ATOMIC_SEQ_CST) != 0) {
         P11PROV_debug("object free: reference held");
         return;
--- a/src/objects.c
+++ b/src/objects.c
@@ -946,8 +946,6 @@ CK_RV p11prov_obj_from_handle(P11PROV_CTX *ctx, P11PROV_SESSION *session,
     if (obj == NULL) {
         return CKR_HOST_MEMORY;
     }
-    obj->handle = handle;
-    obj->slotid = p11prov_session_slotid(session);
     obj->data.key.type = CK_UNAVAILABLE_INFORMATION;
 
     num = 0;
--- a/src/objects.c
+++ b/src/objects.c
@@ -40,6 +40,8 @@ struct p11prov_obj {
     CK_BBOOL cka_copyable;
     CK_BBOOL cka_token;
 
+    P11PROV_OBJ *pub_key_obj;
+
     P11PROV_URI *refresh_uri;
 
     union {
@@ -4148,4 +4150,20 @@ P11PROV_OBJ *mock_pub_ec_key(P11PROV_CTX *ctx, CK_ATTRIBUTE_TYPE type,
     }
 
     return key;
+}
+
+CK_RV p11prov_set_pub(P11PROV_OBJ *priv, P11PROV_OBJ *pub)
+{
+    if (!priv || !pub) return CKR_ARGUMENTS_BAD;
+    if (priv->class != CKO_PRIVATE_KEY || pub->class != CKO_PUBLIC_KEY)
+        return CKR_ARGUMENTS_BAD;
+    if (priv->pub_key_obj) return CKR_ARGUMENTS_BAD;
+    priv->pub_key_obj = pub;
+    return CKR_OK;
+}
+
+P11PROV_OBJ *p11prov_get_pub(P11PROV_OBJ *priv)
+{
+    if (!priv || priv->class != CKO_PRIVATE_KEY) return NULL;
+    return priv->pub_key_obj;
 }
--- a/src/signature.c
+++ b/src/signature.c
@@ -686,11 +686,7 @@ static CK_RV p11prov_sig_op_init(void *ctx, void *provkey, CK_FLAGS operation,
         return ret;
     }
 
-    sigctx->key = p11prov_obj_ref(key);
-    if (sigctx->key == NULL) {
-        return CKR_KEY_NEEDED;
-    }
-    class = p11prov_obj_get_class(sigctx->key);
+    class = p11prov_obj_get_class(key);
     switch (operation) {
     case CKF_SIGN:
         if (class != CKO_PRIVATE_KEY) {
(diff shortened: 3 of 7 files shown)


simo5:
This proposed change has several problems.

First of all, could you open an issue describing what problem you encountered ?

Second, please do not mix extraneous changes (like the clang warning) or other reformatting with feature changes in the same commit.

Third, I do not understand the point of the code you are proposing.
It seems like you are generating a keypair, and then expecting to be able to use the key pair to verify a signature?

But this is failing because, due to how OpenSSL magaes key generation we can only return a single key type and you are left w/o the public key pair part.

If this is the case please clearly state it in the issue you need to open, because if that is the case I need to think about a different architecture, probably a new key abstraction that can hold a public and a private key pair and chose based on operation, in either case the kind of change you are proposing won't be acceptable, as it is a hack that falls short in other areas.

simo5:
Also a change like this *must* come with tests to be acceptable.

latal-1:
> First of all, could you open an issue describing what problem you encountered ?

https://github.com/latchset/pkcs11-provider/issues/480

src/objects.c
```diff
@@ -4107,3 +4115,11 @@ P11PROV_OBJ *mock_pub_ec_key(P11PROV_CTX *ctx, CK_ATTRIBUTE_TYPE type,
 
     return key;
 }
+
+P11PROV_OBJ *p11prov_obj_pub_from_generated(P11PROV_OBJ *obj)
+{
+    if (obj == NULL || obj->isGenerated == false) {
```
Jakuje, line 4121: Do we need to limit this functionality only to generated keys? I think it might be also useful for non-generated ones as the EVP_PKEY private key is mostly abstraction of a key pair in OpenSSL so searching for associated objects to do a verification/decryption would make sense to me also for existing key objects.

simo5: I think we should refactor the code to have a obj->public pointer to another P11PROV_OBJ, so that we can keep the object on generation or dynamically load the associated public key (if possible) as need.
We currently load the associated public key explicitly in just some places, but we should generalize this so more parts will work correctly automatically.

We may also have to set a sentinel value to *prevent* trying to load/expecting a public key if OpenSSL explicitly imported into the provider a key without public bits, and/or to signal that we already searched once and no associated public key was found ion the token.

latal-1: > I think we should refactor the code to have a obj->public pointer to another P11PROV_OBJ

Done

> We may also have to set a sentinel value to prevent trying to load/expecting a public key

Public key object is set only when generating key pairs, otherwise it will be null

src/objects.c
```diff
@@ -270,6 +272,8 @@ P11PROV_OBJ *p11prov_obj_new(P11PROV_CTX *ctx, CK_SLOT_ID slotid,
     obj->class = class;
     obj->cached = CK_INVALID_HANDLE;
 
+    obj->pub_key_obj = NULL;
```
simo5, line 275: no need for this the zalloc above guarantees everything is NULL by default.

src/keymgmt.c
```diff
@@ -482,18 +482,30 @@ static int p11prov_common_gen(struct key_generator *ctx,
     }
 
     ret = p11prov_merge_pub_attrs_into_priv(pub_key, priv_key);
```
simo5, line 484: perhaps we can remove this merging, given that now we can always carry the public key linked

simo5: No need to do it in this PR but you may add a comment to the effect that this call is redundant now that we set a public key

src/keymgmt.c
```diff
@@ -482,18 +482,30 @@ static int p11prov_common_gen(struct key_generator *ctx,
...
-        priv_key = NULL;
+        pub_key = priv_key = NULL;
     }
     p11prov_return_session(session);
-    p11prov_obj_free(pub_key);
     *key = priv_key;
     return ret;
 }
 
+static void p11prov_common_free(void *key)
```
simo5, line 502: Please do not do this, modify p11prov_obj_free() to do the right thing instead.

src/objects.c
```diff
@@ -903,8 +903,6 @@ CK_RV p11prov_obj_from_handle(P11PROV_CTX *ctx, P11PROV_SESSION *session,
     if (obj == NULL) {
         return CKR_HOST_MEMORY;
     }
-    obj->handle = handle;
-    obj->slotid = p11prov_session_slotid(session);
```
simo5, line 903: Please explain this change.

simo5 (changes requested):
It is a good start, some minor changes are required.
that said there is more needed to make this properly functional.

latal-1:
> Also a change like this _must_ come with tests to be acceptable.

Done

src/keymgmt.c
latal-1: Fixed

src/keymgmt.c
latal-1: Added `TODO` to `p11prov_merge_pub_attrs_into_priv` definition

src/objects.c
latal-1: Fixed

src/objects.c
latal-1: In `p11prov_obj_new` `handle` and `slotid` already init with other `P11PROV_OBJ` fields. It seems logical that initialization takes place there

latal-1:
1. Sorry for a long time reaction
2. I can't handle some problems:
* `SIGSEGV` in `tsigver.c` after move deletion from `p11prov_rsa_free` and `p11prov_ec_free` to `p11prov_obj_free`
* `pkcs11-provider:softokn / tls` and `pkcs11-provider:softokn / tlsfuzzer` get error

Can someone solve these problems? Authorship of pull request and commits is not important for me

src/objects.c
simo5: I will handle his later on my own, my comment were assuming a change where we search in the token if we are asked for a public key and the pointer is NULL.

src/keymgmt.c
```diff
@@ -482,14 +482,19 @@ static int p11prov_common_gen(struct key_generator *ctx,
...
     ret = p11prov_merge_pub_attrs_into_priv(pub_key, priv_key);
+    if (ret != CKR_OK) {
+        goto done;
+    }
+
+    ret = p11prov_set_pub(priv_key, pub_key);
 
 done:
     if (ret != CKR_OK) {
+        p11prov_obj_free(pub_key);
```
simo5, line 493: This is wrong now because the next line will already free the pub_key you set on it.
After you set the pub key on the priv key you need to set pub_key = NULL;

This is may be one of the segfault issue, because you are causing a double-free here.

src/keymgmt.c
```diff
@@ -482,14 +482,19 @@ static int p11prov_common_gen(struct key_generator *ctx,
...
         p11prov_obj_free(priv_key);
-        priv_key = NULL;
+        pub_key = priv_key = NULL;
```
simo5, line 495: This become useless once you correctly zero the pub_key right after you pass ownership of it to the priv_key object.

src/objects.c
```diff
@@ -440,6 +440,9 @@ void p11prov_obj_free(P11PROV_OBJ *obj)
     if (obj == NULL) {
         return;
     }
+
+    p11prov_obj_free(obj->pub_key_obj);
+
```
simo5, line 445: This is wrong, all freeing MUST happen *after* the refcounting reaches 0.
Move it after the following block and you'll solve most of your segfaults.

src/signature.c
```diff
@@ -699,17 +695,25 @@ static CK_RV p11prov_sig_op_init(void *ctx, void *provkey, CK_FLAGS operation,
...
+    sigctx->key = p11prov_obj_ref(key);
+    if (sigctx->key == NULL) {
+        return CKR_KEY_NEEDED;
+    }
     sigctx->operation = operation;
 
     if (digest) {
         ret = p11prov_digest_get_by_name(digest, &sigctx->digest);
         if (ret != CKR_OK) {
+            p11prov_obj_free(sigctx->key);
```
simo5, line 716: This is wrong, if you free the key here you also need to NULL the pointer or we will get a double free in p11prov_sig_freectx(), but you do not have to null anything here exactly because p11prov_sig_freectx(0 already takes care of this.

simo5: If you are concerned about someone potentially calling the init operation multiple times, just check if the sigctx->key is not NULL at the start of the function and immediately error if that is the case. But there is no need to add that, it is an application error to try to initialize multiple times and the worst case is some memory leakage.

src/objects.c
simo5: Ok the reason I asked is that this is an extraneous change to the rest of the PR, but it is isolated to a separate commit, so it is ok.

simo5:
Note that DCO is missing in some commit and this will prevent merging until fixed (no exceptions)

I also recommend you rebase interactively and edit each commit by running `make check-style-fix `and then `git add -u; git rebase --continue`, as the style is broken in several places.

simo5 (changes requested):
Please fix the errors I identified and all the CI failures regarding Copyright headers, commit signatures and styles.

We are almost there.

simo5:
Superseded by #516

PR #487, opened 2024-12-12
Test kryoptic with NSS DB

Jakuje:
#### Description

This is more value for the kryoptic to make sure all the operations on the newly introduced DB work as expected.

The NSSDB backend is now super-slow so I am bumping the timeout to 60s/test. The kryoptic build is also adjusted to include this non-default feature.

This change depends on kryoptic including one last fix to provide `CKF_USER_PIN_INITIALIZED` flag from NSS DB backend.

#### Checklist

#### Reviewer's checklist:


tests/kryoptic.nss-init.sh
```diff
@@ -0,0 +1,25 @@
...
+[[slots]]
+slot = 42
+dbtype = "nssdb"
+dbargs = "configDir='${TOKDIR}' flags='passwordRequired'"
+description = "Kryoptic Soft Token"
+_EOF
+
+# this overrides what we define in the generic init
+# the NSS DB can not store custom labels
+export TOKENLABEL="Kryoptic Soft Token"
```
simo5, line 17: You can pass in some of these labels via configuration like you did for the passwordRequired Flags

simo5:
It'd be nice to make the kryoptic.nss text somehow optional and not enabled by default when we run `make check`, I generally want to test with the KRYOPTIC env var set, but I do not want to always try to run the nssdb tests until their perf improves at least.

tests/kryoptic.nss-init.sh
Jakuje: I wanted to make sure that we will really work with the NSS DB so using different one than the SQL DB. I am not sure if these particular knobs are used, but I can double-check

simo5: Tokenlabel is needed by opensc pkcs11-tool to be able to find the right slot/token for some of the operations.
I also do not think they should be identical to the std kryoptic run, I would see it specific like "Kryoptic NSSDB Token"

Jakuje: My intention was to keep the most of the options default.

The `flags='passwordRequired'` is needed for gnutls p11tool to attempt to login when searching for private keys. Without that, it skips the login and fails. It might make sense to describe this in the comment in the script rather than here though.

simo5 (approved):
LGTM,
still I would have expected some better performance, we'll have to do some profiling running the tests on the same host for proper comparison

simo5:
Covscan not needed.

PR #490, opened 2024-12-20
tests: Create separate self-signed EC key for tlsfuzzer testing

Jakuje:
#### Description

Follow up from #488, where we found the tlsfuzzer has problems with heterogeneous CA + server keys. This creates separate EC key outside of the main CA chain (EC Peer key used for ECDH generally), which turned out to be easier than I expected. It also simplifies the setup a bit and removes the excluded test cases, that should be working.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/tests/setup.sh
+++ b/tests/setup.sh
@@ -64,7 +64,7 @@ fi
 P11DEFARGS=("--module=${P11LIB}" "--login" "--pin=${PINVALUE}" "--token-label=${TOKENLABEL}")
 
 # prepare certtool configuration
-cat >> "${TMPPDIR}/cert.cfg" <<HEREDOC
+cat >> "${TMPPDIR}/cacert.cfg" <<HEREDOC
 ca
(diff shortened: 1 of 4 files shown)


tests/setup.sh
```diff
@@ -75,63 +75,73 @@ encryption_key
 cert_signing_key
 HEREDOC
 
+# Serial = 1 is the CA
+SERIAL=1
+
+ca_selfsign() {
```
simo5, line 81: If I understand the command correctly this is not really a ca selfsigned cert, but just a self signed cert, rename to crt_selfsign ?

simo5 (approved):
Only one nit, otherwise LGTM

tests/setup.sh
Jakuje: You are right. Will change.

tomato42 (approved):
lgtm

Issue #494, opened 2024-12-27
Invalid object handles refreshing


bukka:
**Describe the bug**

Currently the pkcs11-provider is able to recreate a new session when it gets closed (implemented in `p11prov_get_session` and `slog_login` after calling the `session_check`). With the exception that it doesn't clear the errors, this works fine and the session gets recreated. The problem happens when some crypto operation (e.g. SignInit) is triggered on the key object that was created during the previous session. This object will use the object handle that no longer exist because the session is no longer existent. So the operation fails and there is no way to recover from this state (except restarting the application) so further operations fail as well.

**To Reproduce**
1. Start pkcs11-proxy daemon
2. Start application (e.g. nginx) using pkcs11-provider using pkcs11 proxy library connection to pkcs11-daemon 
3. Trigger some crypto operation (e.g. https request to nginx)
4. Restart pkcs11-proxy daemon (that clears all current sessions)
5. Send another crypto operation (this fails and any further requests fail as well due to using object handles attached to deleted session)

**Expected behavior**

Key objects should refresh when `CKR_OBJECT_HANDLE_INVALID` is returned and clear all errors if operation is successful.

**Operating environment (please complete the following information):**
 - OS: Ubuntu 20.04 but with custom build of OpenSSL 3.4

**Token and application used (please complete the following information):**
 - [pkcs11-proxy](https://github.com/bukka/pkcs11-proxy) proxying to SoftHSMv2

**Additional context**

My application uses pkcs11-provider with nginx that checks ssl errors so it is also important that all errors are clearer. The whole setup is a bit complex but I can provide more info if anyone needs to recreate it. I have actually already debug everything and have an initial fix so I'm going to create a PR shortly. If we agree on the approach I would look into some automated tests so not putting all details how to manually recreate it here yet.

bukka:
Draft of the fix proposed in #495

simo5:
This is more complex to deal with than it appears, while you could try to renew handles one by one, this is something that should somehow detected and corrected for all objects at the same time, otherwise you may have tokens that on restart reuse handles, and then you get refreshes coming in out of order, such that the first few "invalid" handles are reported as such, and then later a "valid" but incorrect handle ends up used ie you may have the incorrect keys used because in run one it had handles 42 but in run two handle 42 is associated with a different key ...

simo5:
Also if the proxy is restarted, wouldn't any session operation fail given the module is not re-initialized ?

[...]

PR #495, opened 2024-12-27
Fix handling of invalid object handles

bukka:
#### Description

This is an initial fix for https://github.com/latchset/pkcs11-provider/issues/494 which describes all the details about the actual issue.

The fix first removes the errors in session refreshing if the session was successfully re-opened.

It then exposes function for refreshing of invalid objects - this uses the same logic as after the fork refreshing triggered from store.

And it adds a logic in signature for handling `CKR_OBJECT_HANDLE_INVALID` error. It basically refreshes the object and repeat the operation. It also clears errors if refreshing and repeated operation is successful.

For my use case signInit is enough but I realise that this would be quite incomplete so other places should be addressed in the similar way. But before I look into it, I wanted to hear if this approach would be acceptable so please let me know your thoughts. Also if acceptable I will be happy to look into providing some tests. 

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:


src/signature.c
```diff
@@ -935,10 +935,22 @@ static CK_RV p11prov_sig_operate(P11PROV_SIG_CTX *sigctx, unsigned char *sig,
         tbslen += mech->der_digestinfo_len;
     }
 
+    p11prov_set_error_mark(sigctx->provctx);
     ret = p11prov_sig_operate_init(sigctx, false, &session);
     if (ret != CKR_OK) {
-        return ret;
+        if (ret == CKR_OBJECT_HANDLE_INVALID && p11prov_obj_refresh_invalid(sigctx->key) == CKR_OK) {
```
simo5, line 941: you should move this whole section directly in p11prov_sig_operate_init()

src/signature.c
```diff
@@ -870,7 +870,7 @@ static CK_RV p11prov_sig_operate_init(P11PROV_SIG_CTX *sigctx, bool digest_op,
     }
 
 done:
-    if (ret != CKR_OK) {
+    if (ret != CKR_OK && ret != CKR_OBJECT_HANDLE_INVALID) {
```
simo5, line 873: if you do the refreshing in this function, as you should, then this will go away, as it should.
This exception here is a trap for any of the callers.

simo5 (changes requested):
Sounds reasonable in principle, however this change would have to be added to any "_init()" for any operation.
So all the operation in asymmetric_cpiher.c and exchange.c etc...

simo5:
Note that this kind of change will need tests, or it will regress easily.

bukka:
@simo5 Cool, thanks. Ok I will look into adding that to other _init and make it more generic and will also take into account your comments.

In terms of test I could look into adding an integration test for nginx and pkcs11-proxy which should cover the sign one and would be probably useful for testing the fork reload as nginx forks workers. Would you also want tests for other _init? If so, I would probably have to look into some longer running app so I can restart daemon and then retry it. If you have an idea about some existing application for such testing, that would be great!

simo5:
> @simo5 Cool, thanks. Ok I will look into adding that to other _init and make it more generic and will also take into account your comments.
> 
> In terms of test I could look into adding an integration test for nginx and pkcs11-proxy which should cover the sign one and would be probably useful for testing the fork reload as nginx forks workers. Would you also want tests for other _init? If so, I would probably have to look into some longer running app so I can restart daemon and then retry it. If you have an idea about some existing application for such testing, that would be great!

Assuming we can make a generic function for this, then just one test application with the proxy should be fine. Let's see where it goes as you add code.

simo5:
Just a word of warning, closing sessions means destroying all ephemeral keys created as session objects.
This means some handles will never be recoverable. I think we need a way to destroy or mark those objects as unrecoverable otherwise we may end up trying to recover them over and over in some cases.

bukka:
Ah yeah, good point. I will look into it.

bukka:
So after doing various tests with it, I think this approach is not viable becuase it can result in a situation when sessions can get mixed up. In my case the problem is that when the proxy daemon is restarted it resets all sessions (and all handles). What it means is that the session handles will be indexed from 1 after the restart because that's how SoftHSMv2 session manager works. What might then happen is that it might try to use a handle created before the restart that was in the meantime (after the restart) recreated in another slot so there will be an existing handle available but it will represent a different key. I think it just cannot work with continuous refreshing on an invalid handle and it needs to reset all sessions as it's done after the fork.

What I'm thinking instead is the do a slot refresh if there is a device error (maybe having a config option for it so the current behaviour of trying to continue works fine in case there are device that can still work after the device error). @simo5 Would that be an acceptable approach?

simo5:
> So after doing various tests with it, I think this approach is not viable becuase it can result in a situation when sessions can get mixed up. In my case the problem is that when the proxy daemon is restarted it resets all sessions (and all handles).

Yes this is what I would expect.

> What it means is that the session handles will be indexed from 1 after the restart because that's how SoftHSMv2 session manager works. What might then happen is that it might try to use a handle created before the restart that was in the meantime (after the restart) recreated in another slot so there will be an existing handle available but it will represent a different key.

Sessions never represent keys, but sessions can hold ephemeral keys and they would simply lost. Also sessions are not hermetic, if you have a key with a handle you can use it on any sessions, even if the key was created as an ephemeral key o a different sessions, but then keys are lost.

More problematic is that key handles will also be reset, and now you may be using a completely random key, that does not match type or anything. This is why on fork we refresh all handles by re-fetching the keys. Unfortunately this means losing again all ephemeral keys because they are lost when the holding session is closed.

> I think it just cannot work with continuous refreshing on an invalid handle and it needs to reset all sessions as it's done after the fork.

Exactly.

> What I'm thinking instead is the do a slot refresh if there is a device error (maybe having a config option for it so the current behavior of trying to continue works fine in case there are device that can still work after the device error). @simo5 Would that be an acceptable approach?

I am not sure, I think that in order to have the ability to recover from a token error you may also need to cache any key data that was imported in the session, so that you can restore ephemeral keys that openssl copied in to do some operation. But this will still miss any session key generated directly on the token.. oh well.

I wonder if it wouldn't be better to just put pkcs11-provider in an error state and have the application restart ...

But perhaps it is sufficient to ensure that all the objects that went lost stay around but somehow are neutered (invalid handle) and always return errors when OpenSSL tries to use them, you can't just drop objects because they are pointed at by EVP_PEKY->keydata so they need to stay around until openssl frees the data.
I forget if we already properly handle this in the fork case, if we do, then maybe it is ok.

PR #497, opened 2025-01-09
Some fixes for EdDSA signatures + test coverage

Jakuje:
#### Description

There was a logic error in creating the eddsa signature params for pkcs11, that when `OSSL_SIGNATURE_PARAM_INSTANCE` OSSL_PARAM was not provided, the pkcs11 parameters were not created and Ed448 operations were failing.

For some reason, the pkcs11 parameters are required for the Ed448 even if no prehashing nor context is used.

Additionally, if only the context is provided by OpenSSL/caller, we also need to provide the correct parameters, which was not handled before.

Unfortunately, this is not the issue I was after during last days.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -2289,62 +2289,33 @@ static int p11prov_eddsa_get_ctx_params(void *ctx, OSSL_PARAM *params)
 static int p11prov_eddsa_set_ctx_params(void *ctx, const OSSL_PARAM params[])
 {
     P11PROV_SIG_CTX *sigctx = (P11PROV_SIG_CTX *)ctx;
+    const char *instance = "Ed25519";
     const OSSL_PARAM *p;
+    CK_ULONG size;
+    bool matched = false;
     int ret;
 
     P11PROV_debug("eddsa set ctx params (ctx=%p, params=%p)", sigctx, params);
 
-    if (params == NULL) {
-        return RET_OSSL_OK;
+    size = p11prov_obj_get_key_bit_size(sigctx->key);
+    if (size != ED25519_BIT_SIZE && size != ED448_BIT_SIZE) {
+        P11PROV_raise(sigctx->provctx, CKR_KEY_TYPE_INCONSISTENT,
+                      "Invalid EdDSA key size %lu", size);
+        return RET_OSSL_ERR;
+    }
+
+    /* PKCS #11 parameters are mandatory for Ed448 key type anyway */
+    if (size == ED448_BIT_SIZE) {
+        instance = "Ed448";
     }
 
     p = OSSL_PARAM_locate_const(params, OSSL_SIGNATURE_PARAM_INSTANCE);
     if (p) {
-        const char *instance = NULL;
-        bool matched = false;
-        CK_ULONG size;
-
         ret = OSSL_PARAM_get_utf8_string_ptr(p, &instance);
         if (ret != RET_OSSL_OK) {
             return ret;
         }
         P11PROV_debug("Set OSSL_SIGNATURE_PARAM_INSTANCE to %s", instance);
-
-        size = p11prov_obj_get_key_bit_size(sigctx->key);
-        if (size != ED25519_BIT_SIZE && size != ED448_BIT_SIZE) {
-            P11PROV_raise(sigctx->provctx, CKR_KEY_TYPE_INCONSISTENT,
-                          "Invalid EdDSA key size %lu", size);
-            return RET_OSSL_ERR;
-        }
-        if (size == ED25519_BIT_SIZE) {
-            if (OPENSSL_strcasecmp(instance, "Ed25519") == 0) {
-                matched = true;
-                sigctx->use_eddsa_params = CK_FALSE;
-            } else if (OPENSSL_strcasecmp(instance, "Ed25519ph") == 0) {
-                matched = true;
-                sigctx->use_eddsa_params = CK_TRUE;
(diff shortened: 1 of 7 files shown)


tests/tsession.c
```diff
@@ -91,7 +101,12 @@ int main(int argc, char *argv[])
         for (int c = 0; c < CNUM; c++) {
             size_t size = EVP_PKEY_get_size(prikey);
             unsigned char sig[size];
-            ret = EVP_DigestSignFinal(sign_md[c], sig, &size);
+            if (mdname == NULL) {
+                ret =
+                    EVP_DigestSign(sign_md[c], sig, &size, data, sizeof(data));
```
simo5, line 106: Given we do not really need multiple calls, would it mak sense to always just use the one shot call for all cases?

src/signature.c
```diff
@@ -2290,10 +2290,23 @@ static int p11prov_eddsa_set_ctx_params(void *ctx, const OSSL_PARAM params[])
...
+    size = p11prov_obj_get_key_bit_size(sigctx->key);
+    if (size != ED25519_BIT_SIZE && size != ED448_BIT_SIZE) {
+        P11PROV_raise(sigctx->provctx, CKR_KEY_TYPE_INCONSISTENT,
+                      "Invalid EdDSA key size %lu", size);
+        return RET_OSSL_ERR;
+    }
+
+    /* PKCS #11 parameters are mandatory for Ed448 key type anyway */
+    if (size == ED448_BIT_SIZE) {
+        sigctx->use_eddsa_params = CK_TRUE;
```
simo5, line 2307: I do not think this is correct, rather we should select an instance, presumably based on the default and bit lengths and then call through the same code at lines 2322 onward.

What are you doing here is effectively setting
```
    sigctx->use_eddsa_params = CK_TRUE;
    sigctx->eddsa_params.phFlag = CK_FALSE;
```
by omitting to set the phFlag.

I think the correct fix here involves spinning off the instance selection to a helper function, and if no params are set, calling it with "Ed25519" is size == ED25519_BIT_SIZE and with "Ed448" if ED448_BIT_SIZE

An option is to also pre-set instance constants, then if there are params they will overwrite the instance variable with whatever came from the OSSL_SIGNATURE_PARAM_INSTANCE

Once all params are handled, unconditionally call the cose that sets the various flags based on instance name.

simo5 (changes requested):
I think the fix needs changes, but it is correct in abstract

src/signature.c
Jakuje: Please, check the table in the pkcs#11 specs:
https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061191

> What are you doing here is effectively setting

this is intentional. The phFlag is false in this default default. I can set it explicitly if it will make it more clear

Missing instance name on the openssl level results in Ed448/Ed25519. I find it overcomplicated going through the openssl instance if we do not have any from the caller. The code from lines 2322 just really sets `use_eddsa_params` and `phFlag`. Moreover we return on the line 2311 when there are no parameters provided.

tests/tsession.c
Jakuje: Right, this would simplify things. I can change it as I hope no algorithm depends on having update.

Jakuje: Changed.

src/signature.c
simo5: It is a side effect, I anted it explicit, and that can be done easily, I can do it as a followup PR if you like

tests/setup.sh
```diff
@@ -232,34 +232,34 @@ if [ "${TOKENTYPE}" != "softokn" ]; then
...
-# ED2CRTURI="pkcs11:type=cert;object=${ED2CRTN}"
-#
-# title LINE "ED448 PKCS11 URIS"
-# echo "${EDBASEURIWITHPINVALUE}"
-# echo "${EDBASEURIWITHPINSOURCE}"
-# echo "${EDBASEURI}"
-# echo "${EDPUBURI}"
-# echo "${EDPRIURI}"
-# echo "${EDCRTURI}"
+    # this requires OpenSC 0.26.0, which is not available in Ubuntu yet
```
simo5, line 236: at this point we have many checks scattered through setup.sh that test for this or that OS and it makes it difficult to find and change them.

What about having an initial section in setup.sh that sets some generic variables like:
ED448SUPPORT=1
ED25519SUPPORT=1
ECASPARAMETERS=1

and then we have a set of checks that will turn these variables off if the specific system/library we are testing on does not support the primitive.

The actual setup code then just tests for the ED448SUPPORT (or other) variable to decide whether to run the specific setup part.

Jakuje: Its not that many. Just for Ed25519, Ed448 and the explicit curves (which are now moved to the meson). The Ed448 depends on the Ed25519 and the Ed448 needs specific OpenSC version (where I just figured out there is no easy way to find opensc version from any cli). Given that the mapping is sometimes based on the software token, sometimes on opensc version, I am not sure if moving the checks to some other place would make it more readable.

src/signature.c
Jakuje: I set the explicit phFlag now. Hope it will be enough.

Jakuje: Addressed as discussed through meet.

simo5 (approved):

simo5:
Thanks for this fix!

Issue #510, opened 2025-01-22
Meson option to allow override of DEFAULT_PKCS11_MODULE path


embetrix:
**Describe the feature**
the actual `meson.build` does not allow to overwrite the default `DEFAULT_PKCS11_MODULE` path

**Expected behavior**
Add an option to allow configure it at build time e.g 

` meson setup builddir -Ddefault_pkcs11_module="/some/path/mymodule.so"
`

PR #511, opened 2025-01-22
meson: add option to allow override default default_pkcs11_module

embetrix:
fix https://github.com/latchset/pkcs11-provider/issues/510

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -67,8 +67,12 @@ if host_machine.endian() == 'big'
 endif
 
 p11_kit = dependency('p11-kit-1', required: false)
-if p11_kit.found()
+default_pkcs11_module = get_option('default_pkcs11_module')
+if default_pkcs11_module == '' and p11_kit.found()
   default_pkcs11_module = p11_kit.get_variable(pkgconfig: 'proxy_module')
+endif
+
+if default_pkcs11_module != ''
   conf.set_quoted('DEFAULT_PKCS11_MODULE', default_pkcs11_module)
 endif
 
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -7,3 +7,8 @@ option('enable_explicit_EC_test',
        type: 'boolean',
        value: false,
        description: 'Enable explicit EC tests')
+
+option('default_pkcs11_module',
+    type : 'string',
+    value : '',
+    description : 'Path to the default PKCS11 module')


meson.build
```diff
@@ -67,10 +67,15 @@ if host_machine.endian() == 'big'
...
 p11_kit = dependency('p11-kit-1', required: false)
-if p11_kit.found()
-  default_pkcs11_module = p11_kit.get_variable(pkgconfig: 'proxy_module')
-  conf.set_quoted('DEFAULT_PKCS11_MODULE', default_pkcs11_module)
+default_pkcs11_module = get_option('default_pkcs11_module')
+if default_pkcs11_module == 'no'
+  if p11_kit.found()
+    default_pkcs11_module = p11_kit.get_variable(pkgconfig: 'proxy_module')
+  else
+    error('default_pkcs11_module is empty')
```
simo5, line 75: This will now fail to catch that default_pkcs11_module is empty because it has been moved inside a check that will not trigger if it is '' (or anything but 'no').

So the check that fails configuration if default_pkcs11_module is empty probably needs to be move out and done explicitly instead of via an else statement on p11_kit.found()

meson_options.txt
```diff
@@ -2,3 +2,9 @@ option('preload_libasan',
        type: 'string',
        value: 'no',
        description: 'Path to libasan.so to preload')
+
+
+option('default_pkcs11_module',
+    type : 'string',
+    value : 'no',
```
simo5, line 9: This should probably be an empty string and caught in meson.build, unless we really want to allow not setting any default (that would require C code changes too).

simo5 (changes requested):
Some changes are needed

meson.build
```diff
@@ -67,11 +67,17 @@ if host_machine.endian() == 'big'
...
 
 p11_kit = dependency('p11-kit-1', required: false)
-if p11_kit.found()
+default_pkcs11_module = get_option('default_pkcs11_module')
+if default_pkcs11_module == 'no' and p11_kit.found()
   default_pkcs11_module = p11_kit.get_variable(pkgconfig: 'proxy_module')
-  conf.set_quoted('DEFAULT_PKCS11_MODULE', default_pkcs11_module)
 endif
 
+if default_pkcs11_module == ''
```
simo5, line 75: This will now fail to catch that default_pkcs11_module is 'no' but p11_kit was not found to change it.

simo5 (approved):
LGTM

embetrix:
@simo5 looks like libp11-kit-dev dependency is missing in the CI

simo5:
> @simo5 looks like `libp11-kit-dev `dependency is missing in the CI

It may have been intentional, we do support building w/o p11-kit-dev ... in which case no default would be present so now that I think of it I guess we need to allow for no defaults

simo5:
Indeed we do support it not being defined:
https://github.com/latchset/pkcs11-provider/blob/2ef24025b919ca2a78821a4cb6fb5e081220ea34/src/interface.c#L234

So I guess we need to allow an undefined defult, I guess the code needs to be:
if default_pkcs11_module != ''
  conf.set_quoted('DEFAULT_PKCS11_MODULE', default_pkcs11_module)
endif

And lave the CI with default undefined

embetrix:
@simo5: done

simo5:
Lately MacOS CI is flakier and times out, I tried to respin it and timed out again, so I am just going to ignore it for now.
Thanks for the speedy contribution.

Issue #489, opened 2024-12-17
Support for RSA-PSS-only keys


Jakuje:
**Describe the feature**
The OpenSSL (mostly in the context of TLS 1.3) supports special RSA keys, that are designated only to the PSS operation. These could be distinguished on the PKCS#11 level as RSA keys with ALLOWED_MECHANISMS set to only RSA-PSS mechanisms or by having associated certificate with RSA-PSS usage (will have to go back to figure out the right name).

**Expected behavior**
Reading the PKCS#11 objects should try to derive the RSA/RSA-PSS and return different OpenSSL key types based on the allowed mechanisms (if supported by the token) or associated certificate.

**Additional context**
We already have a way to set the ALLOWED_MECHANISMS when we generate key, but it still results in generic RSA key instead of the RSA-PSS one.

https://docs.openssl.org/3.1/man7/RSA-PSS/

Jakuje:
I spent couple of days playing around this and here is my WIP branch with comments in top commit:

https://github.com/Jakuje/pkcs11-provider/tree/rsa-pss-wip

Copying also here:
```
This is work in progress. While the easy deteciton of the keys is in place, we
still miss the encoding the keys to ASN1/DER/PEM through exports and encoders,
where I got completely lost.

There are also WIP tests that excercise these code paths in tls and tlsfuzzer
tests. But the openssl command generating self-signed CA using this code leads
to certificates with invalid public keys, both in

        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption

in place where we should have

            Public Key Algorithm: rsassaPss

(and few more restrictions)

As well as in the signature field, we are getting invalid PSS parameters
(empty):

    Signature Algorithm: rsassaPss
    Signature Value:        (INVALID PSS PARAMETERS)

where we should get

    Signature Value:
        Hash Algorithm: sha256
        Mask Algorithm: mgf1 with sha256
         Salt Length: 0x20
        Trailer Field: 0x01 (default)

In OpenSSL, this is handled in function prepare_rsa_params() defined in

providers/implementations/encode_decode/encode_key2any.c
```

PR #514, opened 2025-01-23
Add support for RSA-PSS Keys

Jakuje:
#### Description

This is not as straightforward as it looked like initially and consists of several changes that surface on different occasions:

 * The private RSA keys can have CKA_ALLOWED_MECHANISMS which can be used to determine if a key is generic key (can be used for any operation) or RSA-PSS key, so that it could be used only for any or specific RSA-PSS operations.

 * These keys get different identifier on OpenSSL level. They also get different OIDs on the ASN.1 in various places of the X.509 certificates:

   * The signatures in X.509 has a OID + parameters describing the hashes, mgf and salt length used. This is mandatory.

   * The public key encoding can contain the RSA-PSS restrictions. These are not mandatory so it allows us to indicate the key is restricted to PSS operations without the need to stick to specific combination of parameters.

 * When we force all operation in pkcs11 provider, the rsapss table was missing the match() callback, making key comparison fail when using RSA-PSS keys.

Given that the certtool we use for signing certificates during setup can not distinguish RSA-PSS restricted keys and therefore generates unrestricted certificates, we need to generate the certificate later using openssl, making the tests a bit more ugly.

Fixes: #489

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:


Jakuje:
(sorry for huge one-commit, but it was pile of different interleaving changes I did not manage to separate to sensible smaller chunks yet)

src/signature.c
```diff
@@ -335,6 +351,7 @@ struct p11prov_mech {
     int der_ecdsa_algorithm_id_len;
     const unsigned char *der_digestinfo;
     int der_digestinfo_len;
+    const EVP_MD *(*hash_md)(void);
```
simo5, line 354: Why do we need this? We already have p11prov_digest_get_name() to get the name to be used in EVP_MD_fetch() like w already do in mech_fallback_init() or p11prov_digest_util() ...

src/signature.c
```diff
@@ -1330,16 +1349,73 @@ static struct {
...
+        goto err;
+    }
+
+    ret = p11prov_mech_by_mechanism(sigctx->pss_params.hashAlg, &mech);
+    if (ret != CKR_OK) {
+        P11PROV_raise(sigctx->provctx, ret, "Failed to get mech for digest %lx",
+                      sigctx->pss_params.hashAlg);
+        goto err;
+    }
+    md = mech->hash_md();
```
simo5, line 1375: these mds here are never actually used, so we go through several cycles of fetching stuff when all we need is just the digest OID to encode in the ASN.1 structure later.
I wonder if there is a way to just pass the OID or the name directly and avoid a bunch of churn.

src/store.c
```diff
@@ -357,27 +356,8 @@ static int p11prov_store_load(void *pctx, OSSL_CALLBACK *object_cb,
     case CKO_PUBLIC_KEY:
     case CKO_PRIVATE_KEY:
         object_type = OSSL_OBJECT_PKEY;
-        type = p11prov_obj_get_key_type(obj);
```
simo5, line 360: I would prefer to keep this code here and simply add a check with p11prov_obj_is_rsa_pss() in the CKK_RSA case.

simo5 (dismissed):
I wish the ASN.1 code was more straightforward and weren't obfuscated by all these OpenSSL Opaque structures.
For example all of the MD and MFG1 EVP_md stuff is ultimately only need to translate from the Hash name to the corresponding OID used in X509 ... I wonder if we shouldn't just store ASN1_OBJECT * pointers instead of the EVP_md ones in our structures to avoid all the churn. Even better if there was a way to just store the serialized ASN.1 once instead of reconstructing it from scratch at every invocation ...
After all there is nothing really dynamic in this information from our POV, the same fixed algorithm will always generate a very defined ASN.1 output.
Maybe we should just generate it once via some ASN.1 utility and store the pre-generated ASN.1 in the rsa-pss_map ?
Is there any need for runtime generation ?

src/objects.c
```diff
@@ -558,6 +558,69 @@ CK_KEY_TYPE p11prov_obj_get_key_type(P11PROV_OBJ *obj)
...
+    for (int i = 0; i < am_nmechs; i++) {
+        bool found = false;
+        for (int j = 0; j < P11PROV_N_RSAPSS_MECHS; j++) {
+            if (allowed[i] == p11prov_rsapss_mechs[j]) {
+                found = true;
+                break;
+            }
+        }
+        if (!found) {
+            non_pss_found = true;
```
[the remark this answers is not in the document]
Jakuje: No need. Its just an artifact of refactoring previous iteration of this function. Will update and simplify.

src/signature.c
Jakuje: I probably misunderstood the OpenSSL API. Now reading that the `EVP_sha1()` has performance impact, while I though it will just return const reference to the existing instance.
I wanted to avoid doing the `EVP_MD_fetch()` as it involves searching for the actual implementation with context, propq ...
Given that we really want just the algorithm name, I think this would be good to simplify. Will have a look what can be done.

src/store.c
Jakuje: Good point. This is also an artifact from previous iteration as  this predates the time I had the `p11prov_obj_is_rsa_pss()` function. Will change.

src/signature.c
Jakuje: Oh, thats the reason:

> Prior to OpenSSL 3.0, functions such as EVP_sha256() which return a "const" object were used directly to indicate the algorithm to use in various function calls.

Looking through the API, we could store the ASN1 DER format, convert it to ASN1_OBJECT using d2i_ASN1_OBJECT function and then pass the object to the algorithm with X509_ALGOR_set0(). Its not straightforward either, but avoids fetching the MD that many times. What do you think?

I would really really like to avoid constructing the Algorithm Identifier structure with PSS_PARAMS by hand. Its hell of nested sequences and I do not completely understand each of the encoding there.

Jakuje: Or we could go through the NID with nid having stored in the table, then calling `OBJ_nid2obj()` and `X509_ALGOR_set0()`, similarly as done in `ossl_X509_ALGOR_from_nid()`. That sounds quite reasonable from my point of view.

src/signature.c
simo5: TBH, unless PSS_PARAMS has any dynamic content, I would seriously considering pre-generating all the options once offline, storing the generated blobs in a static variable.
Then just copying them in place or do a quick DER -> ASN1_OBJECT dance once on first use if ASN1_OBJECT is all that OpenSSL can accept to add this data to the cert.

src/signature.c
Jakuje: I got rid of the EVP_MD now.

The problem is that really all of the fields have default value so they are all optional and we need them in two contexts -- the key restrictions (where we might want to omit some) + signature parameters (we need to have all of them and we might need to support whole matrix of each SHA for digest x SHA for MGF1 (even though it does not make sense as PKCS#11 support this -- not sure if there is way to configure this in openssl though). 

https://datatracker.ietf.org/doc/html/rfc8017#appendix-A.2.3

src/signature.c
Jakuje: got rid of the EVP_MD.

src/signature.c
simo5: After all that RFC says:
```
      it is RECOMMENDED
      that the underlying hash function be the same as the one
      identified by hashAlgorithm
```

Jakuje: Then I can see if I can get them pregenerated somehow reasonably. It should simplify things a bit ...

Jakuje:
I have the signature parameters rewritten to use static structures, but `X509_PUBKEY_set0_param()` requires again the `ASN1_STRING` structure as an input so we will have to convert it from the static structures again (with `d2i_ASN1_STRING` function likely), which makes it a bit defeating the point of having static structures. But I think we can live with it.

simo5:
You already amended your comment, yeah an ASN1_STRING is a lightweight wrapper for a "bunch of bytes", so is not a big deal.

simo5 (approved):
LGTM

simo5:
URI test still very slow on Mac + softhsm, but URI+softhsm is v. slow on other platforms too, pushing for now, and we'll investigate later

PR #531, opened 2025-03-10
Debug logs extra

bukka:
#### Description

This adds some debug logs that I have been using for debugging various issues with pkcs11-proxy. There are no changes in logic (except some renaming so logging can happen after the locked section). It is quite useful to see how the objects are getting to the pool which also uncovered one issue in the provider (different PR will be created for that).

I'm of course happy to remove the ones that you don't find useful - basically just trying to minimize the diff for my branch.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -212,6 +212,11 @@ done:
     (void)MUTEX_UNLOCK(pool);
     /* ------------- LOCKED SECTION */
 
+    if (ret == CKR_OK) {
+        P11PROV_debug("Object added to pool (idx=%d, obj=%p)", obj->poolid,
+                      obj);
+    }
+
     return ret;
 }
 
@@ -219,6 +224,7 @@ static void obj_rm_from_pool(P11PROV_OBJ *obj)
 {
     P11PROV_OBJ_POOL *pool;
     CK_RV ret;
+    const char *errstr = NULL;
 
     if (obj->slotid == CK_UNAVAILABLE_INFORMATION) {
         /* a mock object */
@@ -230,15 +236,21 @@ static void obj_rm_from_pool(P11PROV_OBJ *obj)
         return;
     }
 
+    P11PROV_debug("Object to be removed from pool (idx=%d, obj=%p)",
+                  obj->poolid, obj);
+
     ret = MUTEX_LOCK(pool);
     if (ret != CKR_OK) {
         return;
     }
 
     /* LOCKED SECTION ------------- */
-    if (obj->poolid >= pool->size || pool->objects[obj->poolid] != obj) {
-        ret = CKR_GENERAL_ERROR;
-        P11PROV_raise(pool->provctx, ret, "Objects pool in inconsistent state");
+    if (obj->poolid >= pool->size) {
+        errstr = "small pool";
+        goto done;
+    }
+    if (pool->objects[obj->poolid] != obj) {
+        errstr = "obj already removed";
         goto done;
     }
 
@@ -252,6 +264,12 @@ static void obj_rm_from_pool(P11PROV_OBJ *obj)
 done:
     (void)MUTEX_UNLOCK(pool);
     /* ------------- LOCKED SECTION */
+
+    if (errstr != NULL) {
+        P11PROV_raise(pool->provctx, ret,
+                      "Objects pool in inconsistent state - %s (obj=%p)",
+                      errstr, obj);
+    }
 }
 
 static CK_RV p11prov_obj_store_public_key(P11PROV_OBJ *key);
@@ -3366,6 +3384,10 @@ static CK_RV return_dup_key(P11PROV_OBJ *dst, P11PROV_OBJ *src)
 {
     CK_RV rv;
 
+    P11PROV_debug("duplicating obj key (dst=%p, src=%p, handle=%lu, "
+                  "slotid=%lu, raf=%d, numattrs=%d)",
+                  dst, src, src->handle, src->slotid, src->raf, src->numattrs);
+
     dst->slotid = src->slotid;
     dst->handle = src->handle;
     dst->class = src->class;
@@ -3434,6 +3456,8 @@ static CK_RV fix_ec_key_import(P11PROV_OBJ *key, int allocattrs)
     key->attrs[key->numattrs].ulValueLen = len;
     key->numattrs++;
 
+    P11PROV_debug("fixing EC key %p import", key);
+
     return CKR_OK;
 }
 
@@ -3459,6 +3483,7 @@ static CK_RV p11prov_obj_import_public_key(P11PROV_OBJ *key, CK_KEY_TYPE type,
 
     switch (type) {
     case CKK_RSA:
+        P11PROV_debug("obj import of RSA public key %p", key);
         rv = prep_rsa_find(ctx, params, &findctx);
         if (rv != CKR_OK) {
             goto done;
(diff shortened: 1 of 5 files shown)


src/objects.c
```diff
@@ -236,22 +243,39 @@ static void obj_rm_from_pool(P11PROV_OBJ *obj)
...
+            P11PROV_raise(pool->provctx, ret,
+                          "Objects pool in inconsistent state - small pool "
+                          "(idx=%d, size=%d)",
+                          idx, pool->size);
+        } else {
+            P11PROV_raise(pool->provctx, ret,
+                          "Objects pool in inconsistent state - obj already "
+                          "removed (idx=%d, pool_obj=%p, obj=%p)",
+                          idx, pool->objects[idx], obj);
+        }
```
simo5, line 259: perhaps it is time to split the whole error from line 247, instead of inferring that the second banch here is caused by `pool->objects[idx] != obj)`.


IE do:

```
if (idx >= pool->size) {
  ret = ..
  P11PROV_raise(..
  goto done;
}
if (pool->objects[idx] != obj) {
  ret = ...
  P11PROV_raise(..
  goto done;
}
```

simo5 (changes requested):
Minor nit, but otherwise LGTM

src/objects.c
bukka: yeah that looks better. Done

src/objects.c
```diff
@@ -154,6 +154,7 @@ static CK_RV obj_add_to_pool(P11PROV_OBJ *obj)
 {
     P11PROV_OBJ_POOL *pool;
     CK_RV ret;
+    int idx;
```
simo5, line 158: why this ?

src/objects.c
```diff
@@ -212,13 +213,20 @@ static CK_RV obj_add_to_pool(P11PROV_OBJ *obj)
     (void)MUTEX_UNLOCK(pool);
     /* ------------- LOCKED SECTION */
 
+    if (ret == CKR_OK) {
+        P11PROV_debug(
+            "Object added to pool (idx=%d, first_free=%d, obj=%p, num=%d)", idx,
```
simo5, line 218: nevermind, I saw only later idx is used here.

simo5: however you can simply use obj->poolid here, instead of extracting idx all they way out there, can you do that ?

src/objects.c
```diff
@@ -236,22 +244,41 @@ static void obj_rm_from_pool(P11PROV_OBJ *obj)
...
     }
     obj->poolid = 0;
 
 done:
     (void)MUTEX_UNLOCK(pool);
     /* ------------- LOCKED SECTION */
+
+    if (ret == CKR_OK) {
+        P11PROV_debug(
+            "Object removed from pool (idx=%d, first_free=%d, obj=%p)", idx,
```
simo5, line 279: Is it intentional here to put the poolid before it was zeroed ?
I guess the intention here is to show what the idx was, however perhaps it would be better to just add a debug line as soon as we enter the function announcing what the function is attempting to do, and then you can simplify also the new debug messages by just stating the error, and avoid having to introduce a idx variable at all.

bukka: The main reason for putting this at the end is not to add any logging to the locked section as it could slightly impact contentation. My client has significant traffic with lots of pool changes and we might need to enable debug logging in production to trouble shoot some issue. I just wanted to make sure that this won't impact the locked section that is currently really minimal so adding a write to file could be potentially noticable - I haven't verified if it can really have any impact though but though that adding it after is just a safer option.

src/objects.c
bukka: yeah it's probably better not to extract it. changed.

src/objects.c
simo5: Well that is why I said at the beginning of the function (before the locked section beings). 

Also if lock contention is a concern in debug mode then even more importantly you should avoid constructing error string in the lcoked section and instead setting a static error string based on the failure and defer P11PROV_raise() till after the locked section, as P11PROV_raise does a lot more than calling debug as it sets errors in the openssl error stack (and there is even a potential for deadlocking in there now that I think of it).


So I would do:
```
static void obj_rm_from_pool(P11PROV_OBJ *obj)
{
    P11PROV_OBJ_POOL *pool;
    const char *errstr = NULL;
    CK_RV ret;
    int idx;

    if (obj->slotid == CK_UNAVAILABLE_INFORMATION) {
        /* a mock object */
        return;
    }

    P11PROV_debug("Object to be removed from pool "
                  "(obj pool id=%d, first free=%d,"
                  " pool size=%d, obj=%p)",
                  obj->poolid, pool->first_free,
                  pool->size, obj);

   [...]

   if (idx >= pool->size) {
       ret = CKR_GENERAL_ERROR;
       errstr = "Objects pool in inconsistent state - small pool"
       goto done;
   }

   [...]

done:
   if (ret != CKR_OK) {
       P11PROV_raise(pool->provctx, ret,
                     "Failed to remove object: %s",
                     errstr ? errstr : "Unavailable errstr");
   [...]

```

simo5: edited the example to show where lock/unlock happens

bukka: Yeah this actually much better. Just pushed slightly simplified version of the above.

bukka: Thanks.

simo5 (approved):
LGTM

simo5:
Ugh coverity flagged all the debug statements accesses of the pool  (pool->num, pool->first_free, pool->size as unsafe access because we protect the pool struct with a lock.

And technically it is correct, it is possible a different thread is changing those fields under us while we are emitting the debug.

To make coverity happy we'd have to allocate temporary variables on the stack, save them in the locked session and use them in the debug statements.

For obj_add_to_pool() this is easy, as we debug after the locked session anyway.
For obj_rm_from_pool() this means moving logging of pool related info after the locked section de-referencing only object related data at the start of the function.

Can you please fix the code that way?
(also please) rebase on top of current main as I already merged a few of your PRs

simo5 (changes requested):
See comment for requested changes

bukka:
That makes sense, the pool details were not that important anyway - just idx and obj pointer were really useful for debugging so I stripped it a bit and put back the idx to `obj_rm_from_pool`. Hopefully it will be all good now.

src/objects.c
```diff
@@ -212,13 +212,20 @@ static CK_RV obj_add_to_pool(P11PROV_OBJ *obj)
...
+
     return ret;
 }
 
 static void obj_rm_from_pool(P11PROV_OBJ *obj)
 {
     P11PROV_OBJ_POOL *pool;
     CK_RV ret;
+    const char *errstr = NULL;
+    int idx;
```
simo5, line 228: It id not a problem to de-reference obj->poolid, so there is no need to reintroduce idx here.

Up can keep the debug message before the locked section given you removed all pool-> de-referencing.

simo5 (changes requested):
Sorry if I was not clear but the structure of the debug messages (with the main debug message before the locked session) was fine.

Only the de-referencing of the pool members was problematic, but those have been removed now.

I prefer to avoid the local variable idx generally.

(also please rebase on main if you can, but not big deal I can also force the update myself before running coverity again)

src/objects.c
bukka: Yeah of course. Sorry it's quite late in Europe... :)

bukka: Fixed

bukka:
@simo5 all rebased and that idx part also fixed. I moved it because I thought that I should save the slot value that were previously used but then removed those. It was a bit late to put it back... :) Anyway all should be addressed now.

simo5 (approved):
Great, thanks!

src/objects.c
simo5: Please never feel pressured to react immediately, I am ok waiting a bit to get things right.

PR #532, opened 2025-03-10
Debug logs time

bukka:
#### Description

This PR adds a configurable time entry to the debug log. The first commit adds it as disable by default and the second one enables it by default (option can still be used to disable it in that case). I find it more useful to have time there as my use case involved correlation with other logs. But I understand if you prefer not to have it enabled by default. In such case I will remove the second commit.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -7,7 +7,7 @@
 #include <string.h>
 
 int debug_level = -1;
-FILE *stddebug = NULL;
+static FILE *stddebug = NULL;
 
 /* this function relies on being called by P11PROV_debug, after
  * an __atomic_compare_exchange_n sets debug_lazy_init to -1,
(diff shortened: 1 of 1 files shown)


simo5:
@bukka Is there any good reason to make it optional ?

simo5:
If we want to make it optional let's enable it if debug_level is higher, instead of adding another option to the already fragile debug string syntax, otherwise I would just simply always add timestamps.

bukka:
I don't think there is much point to make it optional. I added that just in case you prefer it without the time. I removed the option and did a quick check. Loggs look like:

```
[2025-03-10 17:28:19.676] [../src/provider.c:1494] OSSL_provider_init(): Provided config params:
....

simo5:
It does make a log line much longer ... would it make sense to not only the time and not necessarily the date.
Or perhaps even frog the pretty print and log directly the unix-time integer?

I guess the question really is: do you need the actual date/time to correlate it with something else, or is relative difference between operations all that is needed here?

Not asing for further changes at this time, only trying to understand the goal.

bukka:
Yeah I need it mainly for correlation with other logs (nginx in my case). So using unix time integer would be a bit pain... But I can live with only keeping the time (although still prefer to keep milliseconds) if you prefer it but not sure if those 11 characters will make much difference. Some logs are already quite long anyway.

simo5:
If you need correlation with external log, might as well have the whole date in it, and in a convenient format.
This is just debugging so probably better to have too much info than too little, people can easily parse the log file through sed and remove the dates if they are upset about the verbosity :)

simo5:
Would you mind squashing the two commits?
Then I can merge.

bukka:
Thanks. Squashed.

simo5:
thanks!

PR #533, opened 2025-03-10
Prevent removing duplicated objects from pool

bukka:
#### Description

This PR prevents attempt to remove objects that have been created by duplication. Those objects have never been added to the pool but they are still attempted to be removed which results in an error.

#### Checklist

- [ ] Test suite updated with tests (still need to figure this out)

#### Reviewer's checklist:

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -32,6 +32,7 @@ struct p11prov_crt {
 struct p11prov_obj {
     P11PROV_CTX *ctx;
     bool raf; /* re-init after fork */
+    bool dup; /* duplicated */
 
     CK_SLOT_ID slotid;
     CK_OBJECT_HANDLE handle;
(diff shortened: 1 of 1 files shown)


src/objects.c
```diff
@@ -32,6 +32,7 @@ struct p11prov_crt {
 struct p11prov_obj {
     P11PROV_CTX *ctx;
     bool raf; /* re-init after fork */
+    bool dup; /* duplicated */
```
simo5, line 36: I do not think we need a new boolean here.
We have the poolid integer, if the object has not been added to a pooolid you can simply set poollid to a special value to indicate that. That value can be MAX(int) for example.

simo5 (changes requested):
Can you pleas eopen a bug that describe what you are trying to fix here as well?
Thanks.

bukka:
> Can you pleas eopen a bug that describe what you are trying to fix here as well?

Done in #540

simo5:
Closing this as I integrated a fix in #649

PR #535, opened 2025-03-10
Use PIN from refresh URI when taking login session

bukka:
#### Description

If multiple slots with different PINs are used and the pins are provided in uri using pin-value query param, then it doesn't work, then it might not get correctly the open session. With the current code it's more theoretical as there is not sessions refreshing needed but I saw it when I did session refreshing.

I will think about this more if it really makes sense actually.

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:


simo5:
This looks reasonable, but it also absolutely needs some tests if at all possible.

bukka:
This is superseded by #789

PR #536, opened 2025-03-10
Add pkcs11-module-default-slot-id cfg for default slot ID

bukka:
#### Description

There are use cases when one wants to select the default slot id (especially when the first slot ID is not convenient for this purpose) which is used for example for storing public keys.

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/docs/provider-pkcs11.7.md
+++ b/docs/provider-pkcs11.7.md
@@ -242,6 +242,23 @@ Example:
 
 ```pkcs11-module-assume-fips = true```
 
+## pkcs11-module-default-slot-id
+
+The default slot ID for new objects.
+
+The default slot is used for any objects that do not have an explicitly
+selected slot. Such objects include, for example, public keys.
+
+This option is useful when multiple slots with different PINs are used. In such
+cases, it allows you to ensure that pkcs11-module-token-pin points to the
+expected slot.
+
+If no value is specified, the first listed slot is used.
+
+Default: none
+
+```pkcs11-module-default-slot-id = 12345```
+
 
 ENVIRONMENT VARIABLES
 =====================
--- a/src/provider.c
+++ b/src/provider.c
@@ -33,6 +33,7 @@ struct p11prov_ctx {
     int cache_sessions;
     bool encode_pkey_as_pk11_uri;
     bool assume_fips;
+    CK_SLOT_ID default_slotid;
     /* TODO: ui_method */
     /* TODO: fork id */
 
@@ -461,6 +462,11 @@ P11PROV_INTERFACE *p11prov_ctx_get_interface(P11PROV_CTX *ctx)
         return NULL;
     }
     return p11prov_module_get_interface(ctx->module);
+}
+
+CK_SLOT_ID p11prov_ctx_get_default_slotid(P11PROV_CTX *ctx)
+{
+    return ctx->default_slotid;
 }
 
 P11PROV_SLOTS_CTX *p11prov_ctx_get_slots(P11PROV_CTX *ctx)
@@ -1778,6 +1784,7 @@ enum p11prov_cfg_enum {
     P11PROV_CFG_ENCODE_PROVIDER_URI_TO_PEM,
     P11PROV_CFG_BLOCK_OPS,
     P11PROV_CFG_ASSUME_FIPS,
+    P11PROV_CFG_DEFAULT_SLOT_ID,
     P11PROV_CFG_SIZE,
 };
 
@@ -1797,6 +1804,7 @@ static struct p11prov_cfg_names {
     { "pkcs11-module-encode-provider-uri-to-pem" },
     { "pkcs11-module-block-operations" },
     { "pkcs11-module-assume-fips" },
+    { "pkcs11-module-default-slot-id" },
 };
 
 int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
@@ -2088,6 +2096,24 @@ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
     if (ret != CKR_OK) {
         p11prov_ctx_free(ctx);
         return RET_OSSL_ERR;
+    }
+
+    if (cfg[P11PROV_CFG_DEFAULT_SLOT_ID] != NULL) {
+        char *end = NULL;
+        errno = 0;
+        ctx->default_slotid =
+            strtoul(cfg[P11PROV_CFG_DEFAULT_SLOT_ID], &end, 0);
+        if (errno != 0 || *end != '\0') {
+            P11PROV_raise(ctx, CKR_GENERAL_ERROR, "Invalid value for %s: (%s)",
+                          p11prov_cfg_names[P11PROV_CFG_DEFAULT_SLOT_ID].name,
+                          cfg[P11PROV_CFG_DEFAULT_SLOT_ID]);
+            p11prov_ctx_free(ctx);
+            return RET_OSSL_ERR;
+        }
+        P11PROV_debug("Default slot ID: %lu", ctx->default_slotid);
+    } else {
+        ctx->default_slotid = CK_UNAVAILABLE_INFORMATION;
+        P11PROV_debug("Default slot ID: None");
     }
 
     /* PAY ATTENTION: do this as the last thing */
--- a/src/slot.c
+++ b/src/slot.c
@@ -173,6 +173,10 @@ CK_RV p11prov_init_slots(P11PROV_CTX *ctx, P11PROV_SLOTS_CTX **slots)
     }
 
     sctx->default_slot = CK_UNAVAILABLE_INFORMATION;
+    CK_SLOT_ID prov_default_slotid =
+        p11prov_ctx_get_default_slotid(sctx->provctx);
+    if (prov_default_slotid != CK_UNAVAILABLE_INFORMATION)
+        sctx->default_slot = prov_default_slotid;
 
     for (size_t i = 0; i < num; i++) {
         P11PROV_SLOT *slot;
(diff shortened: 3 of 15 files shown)


src/provider.c
```diff
@@ -1733,6 +1741,24 @@ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
...
 
+    if (cfg[P11PROV_CFG_DEFAULT_SLOT_ID] != NULL) {
+        char *end = NULL;
+        errno = 0;
+        ctx->default_slotid =
+            strtoul(cfg[P11PROV_CFG_DEFAULT_SLOT_ID], &end, 0);
+        if (errno != 0 || *end != '\0') {
+            P11PROV_raise(ctx, CKR_GENERAL_ERROR, "Invalid value for %s: (%s)",
+                          p11prov_cfg_names[P11PROV_CFG_ALLOW_EXPORT].name,
+                          cfg[P11PROV_CFG_ALLOW_EXPORT]);
```
Jakuje, line 1752:
```suggestion
                          p11prov_cfg_names[P11PROV_CFG_DEFAULT_SLOT_ID].name,
                          cfg[P11PROV_CFG_DEFAULT_SLOT_ID]);
```

Jakuje (commented):
Do you plan to get back to this PR any time soon to finalize it? I found one copy paste error, but otherwise I think it is going in the right direction!

src/provider.c
bukka: Fixed, thanks

bukka:
> Do you plan to get back to this PR any time soon to finalize it? 

I just fixed that typo and added docs but tests will take me a bit more time. I had a look and it seems like it has currently only single slot setup so might need some extending to allow adding setup for multiple slots with different pins and then somehow pass through the slot id so it can be used (e.g. SoftHSMv2 will re-assign the value so it needs to be fetched - e.g. for the output message). Or maybe there is an easier way how to test it?

simo5:
> > Do you plan to get back to this PR any time soon to finalize it?
> 
> I just fixed that typo and added docs but tests will take me a bit more time. I had a look and it seems like it has currently only single slot setup so might need some extending to allow adding setup for multiple slots with different pins and then somehow pass through the slot id so it can be used (e.g. SoftHSMv2 will re-assign the value so it needs to be fetched - possibly from the output message). Or maybe there is an easier way how to test it?

Kryoptic can easily be configured to have as many slots as you want (each one with a different token), you just need to provide a config file that specifies them and initialize them.

Jakuje:
Here is a draft commit that adds multiple slots to the kryoptic setup (and initializes just one):

https://github.com/Jakuje/pkcs11-provider/commit/55272935f4f6764d800a04c003c94f592bf5584f

Feel free to reuse it or adjust it to crate a test case. I do not exactly know how the problem you are solving would demonstrate so if you will need some help to put things together, please let me know.

bukka:
I spent another chunk of time on this and getting this tested is actually a bit harder.

First I tried the new approach with just importing the pubkey. This can be done in a couple of ways and the simplest seemed to be doing that during the match. It means using `EVP_PKEY_eq`. However this didn't really get me too far because such imported key is just temporarily imported but no actual operation that requires login will run on it. It also doesn't get stored because storing happens only in `p11prov_obj_get_handle`.

I thought about it and what I really need first is to somehow get EVP_DigestVerifyInit executed on public key that will have keymgmt pointing to pkcs11 provider. That imports the key (through `evp_keymgmt_export -> ec_export / rsa_export -> evp_keymgmt_util_try_import`) and then does the verify (that should call `p11prov_sig_operate_init` which calls `p11prov_obj_store_public_key`). This is actually the path where I saw some other issues so it would be really good to get it tested as it would help for some other tests.

The part that seems a bit harder is to get the public key keymgmt pointing to pkcs11 provider. At least it doesn't seem to work in test when using store load as used by testing `load_key` even for pkcs11 URI (e.g. tried with `pkcs11:type=public;id=%02%01`) or pubkey PEM for pubkey present in hsm. The reason is that it just calls export from the provider and than cache the pubkey and uses default provider for further operations. I guess I need to find a different way how to load it because I can actually see it used when debugging nginx which happens during handshake. I will need to look more to the TLS code to see how it's loaded there for client cert pub key.

simo5:
Simply set propquery to "provider=pkcs11" in your tests and it should try to force operations to happen on the token.
However because the provider delays initialization you may need to change the token configuration to init early in order for openssl to find all the functionality the provider can offer.

tests/kryoptic.multislot-init.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash -ex
+# Copyright (C) 2024 Jakub Zelenka <jakub.openssl@gmail.com>
```
Jakuje:
```suggestion
# Copyright (C) 2024 - 2026 Jakub Zelenka <jakub.openssl@gmail.com>
```

docs/provider-pkcs11.7.md
```diff
@@ -242,6 +242,23 @@ Example:
 
 ```pkcs11-module-assume-fips = true```
 
+## pkcs11-module-default-slot-id
+
+The slot ID for the default slot.
```
Jakuje:
```suggestion
The default slot ID for new objects. 
```

Jakuje (dismissed):
just a nit for documentation, otherwise looks good.

Not sure whats up with the CI though

simo5:
@bukka can you squash all commits that deal with tests? I see a lot of back and forth there and would rather collapse all dead-ends away

bukka:
@simo5 I squashed to a single commit which probably makes most sense as that multi slot setup should be introduced with some test that uses it. It's also rebased.

src/slot.c
```diff
@@ -238,6 +240,8 @@ CK_RV p11prov_init_slots(P11PROV_CTX *ctx, P11PROV_SLOTS_CTX **slots)
          * softoken has a slot that can't be used to store session keys)
          * and the following query excludes it */
         if ((sctx->default_slot == CK_UNAVAILABLE_INFORMATION)
+            && (prov_default_slotid == CK_UNAVAILABLE_INFORMATION
+                || prov_default_slotid == slot->id)
```
simo5: Why did you clamp this check down here?
This code is meant to avoid using softokn first slot whch is usually a read-only slot and it also usually *not* what you'd mark as default.

tests/meson.build
```diff
@@ -122,7 +123,9 @@ foreach t, sources : test_programs
 endforeach
 
 setup_script=find_program('setup.sh')
-all_suites=['softokn', 'softhsm', 'kryoptic', 'kryoptic.nss']
+multi_suites=['kryoptic.multislot']
```
simo5: Doesn't look like you really need to change around these definitions, given kryoptic.multislot is used only in one test you can simply use it here, and leave "all_suites" as it is. It will make more clear that kryoptic.multislot is special and not one of the generic suites to run.
This will cause less churn as well.

simo5 (changes requested):
Overall we are close, just a few nitpicks from me on top of those from @Jakuje

src/slot.c
bukka: This just adds an extra condition to make sure that the specified provider default slot is set only if it's a valid slot id (meaning if there is any slot with such ID). The softokn conditions still applies here as it's AND. What am I missing?

tests/meson.build
bukka: Yeah sure I can change it if you prefer.

Just to be clear, all will not really mean all then. Also it might be useful to also add softhsm.multislot in the future so there could be more multislot ones.

But as I said I don't really mind to change it. This was just my thinking why I did it this way. Let me know if you still want me to change and I will do that.

bukka (commented):
Oh I replied 2 weeks ago but forgot to submit so here are my replies - sorry for the delay :)

src/slot.c
simo5: I would think that if a slot is set in configuration as the default it should be used regardless of other considerations ?

tests/meson.build
simo5: yes I prefer to keep the change minimal for now, and only add multislot to "all" if it really becomes more widely used.

docs/provider-pkcs11.7.md
```diff
@@ -242,6 +242,23 @@ Example:
...
+The default slot ID for new objects.
+
+The default slot is used for any objects that do not have an explicitly
+selected slot. Such objects include, for example, public keys.
+
+This option is useful when multiple slots with different PINs are used. In such
+cases, it allows you to ensure that pkcs11-module-token-pin points to the
+expected slot.
+
+If no value is specified, the first listed slot is used.
```
Jakuje, line 256: nit: The logic for the selection of the slot is more complicated -- the slot needs to be initialized, it needs to require a login and it can not be locked:
```
            && (slot->token.flags & CKF_LOGIN_REQUIRED)
            && (slot->token.flags & CKF_TOKEN_INITIALIZED)
            && (!(slot->token.flags & CKF_USER_PIN_LOCKED))) {
```
I think it would be better to be a bit more specific here, for example:

```suggestion
If no value is specified, the first slot listed by the token, that is initialized and does not have a PIN locked is used.
```
(I do not insist of this wording -- but I would prefer to have the documentation more specific)

src/slot.c
bukka: Ok changed it so it gets assigned unconditionally before the loop.

tests/meson.build
bukka: done

docs/provider-pkcs11.7.md
bukka: I changed the logic so it is set unconditionally so this no longer applies.

Jakuje (approved):

simo5 (approved):
LGTM

simo5:
Thanks for the contribution!

PR #539, opened 2025-03-10
Fix CS in p11prov_obj_pool_free

bukka:
#### Description

Minor formatting change so running `clang-format -i src/object.c` works fine.


simo5 (approved):
LGTM

PR #541, opened 2025-03-11
Fix CS in the generate code

bukka:
#### Description

This is just a minor update to reduce churn when proposing PR with generated code changes. It was probably created with the old formatter / old rules so this updates it.


simo5 (approved):
LGTM!
Thanks.

simo5:
Just style changes, coverity checks not needed.

PR #544, opened 2025-03-13
Correctly handle some aspects of key import

### LINKED ISSUE #543: import key doesn't always set key->class
**Describe the bug**
In cases like ECDH when creating new empty key and copying domain parameters from another key, key->class is not set correctly.

**To Reproduce**

EVP_PKEY *peer_pkey = EVP_PKEY_new();
// assume here that pkey is a valid, fully constructed EC key
EVP_PKEY_copy_parameters(peer_pkey, pkey)
size_t peer_pkey_bits = EVP_PKEY_bits(peer_pkey);
// peer_pkey_bits is 0 (zero).

**Expected behavior**
peer_pkey_bits above is expected to be 32.

**Additional context**
This issue seems to be introduced by commit 6cc41ff56bc67fa40332b3101ecfdd23e448be84 (Ensure we correctly deal with import selection).
In the above test scenario, key->class is not set correctly (key is peer_pkey's P11PROV_OBJ) by p11prov_obj_import_key() which causes EVP_PKEY_bits() to return zero later.

simo5:
#### Description

OpenSSL assumes you can create a new EC key by copy the domain parameters from a peer key first (to establish a compatible key type for operations like ECDH), and only later generates the private key material.
    
Better identify those keys by assigning the CKO_DOMAIN_PARAMETER class to them once parameters are set. We do not have a fully formed key at this point but we already have a bunch of parameter set so this also allows to make decisions on what should or should not be changed anymore at this point. (for example this now will prevent re-importing other parameters over the key).
    
Fixes #543

#### Checklist

- [ ] ~Documentation updated~

#### Reviewer's checklist:

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -574,6 +574,7 @@ CK_KEY_TYPE p11prov_obj_get_key_type(P11PROV_OBJ *obj)
         switch (obj->class) {
         case CKO_PRIVATE_KEY:
         case CKO_PUBLIC_KEY:
+        case CKO_DOMAIN_PARAMETERS:
             return obj->data.key.type;
         }
     }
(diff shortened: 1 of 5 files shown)


Jakuje (approved):
lgtm

PR #545, opened 2025-03-13
FIX: Compare attributes of type bignum as native-endian bignum

bashmachnikov:
#### Description
# **Fix: Correct Comparison for Numeric `CK_ATTRIBUTE_TYPE` Values** 

## **Problem Description**  
1. **Inconsistent BIGNUM Representations:**  
   - OpenSSL exports the RSA public exponent (`e = 65537`) into provider  `object->attrs` using a **minimal byte representation** (`3` bytes: `0x01 00 01`).  
   - AWS CloudHSM returns the same exponent value (`65537`), but it is stored as **8 bytes** (`0x01 00 01 00 00 00 00 00`).  

2. **Memory Comparison Issue in `cmp_attr`:**  
   - The function `cmp_attr(pub_key1, pub_key2, CKA_PUBLIC_EXPONENT)` **fails** because it performs a **raw memory comparison**.  
   - Due to different encodings (**3 bytes vs. 8 bytes**), the function incorrectly treats identical values as different.  

## **Proposed Solution**  
For attributes representing numeric values (e.g., `CKA_PUBLIC_EXPONENT`, `CKA_MODULUS`):  
Compare values using **`BN_cmp()`** instead of a raw memory comparison.  

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -2477,6 +2477,39 @@ CK_ATTRIBUTE *p11prov_obj_get_ec_public_raw(P11PROV_OBJ *key)
     return pub_key;
 }
 
+static int cmp_bn_attr(P11PROV_OBJ *key1, P11PROV_OBJ *key2,
+                       CK_ATTRIBUTE_TYPE attr)
+{
+    BIGNUM *bx1;
+    BIGNUM *bx2;
+    CK_ATTRIBUTE *x1, *x2;
+    int rc = RET_OSSL_ERR;
+
+    /* is BN ?*/
+    if (attr != CKA_MODULUS && attr != CKA_PUBLIC_EXPONENT) {
+        return rc;
+    }
+
+    x1 = p11prov_obj_get_attr(key1, attr);
+    x2 = p11prov_obj_get_attr(key2, attr);
+
+    if (!x1 || !x2) {
+        return rc;
+    }
+
+    bx1 = BN_native2bn(x1->pValue, x1->ulValueLen, NULL);
+    bx2 = BN_native2bn(x2->pValue, x2->ulValueLen, NULL);
+
+    if (BN_cmp(bx1, bx2) == 0) {
+        rc = RET_OSSL_OK;
+    }
+
+    BN_free(bx1);
+    BN_free(bx2);
+
+    return rc;
+}
+
 static int cmp_attr(P11PROV_OBJ *key1, P11PROV_OBJ *key2,
                     CK_ATTRIBUTE_TYPE attr)
 {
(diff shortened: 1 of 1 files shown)


simo5:
This is what the spec PKCS#11 spec says:
```
  Big integer       a string of CK_BYTEs representing an unsigned integer of arbitrary size, most-significant byte first (e.g., the integer 32768 is represented as the 2-byte string 0x80 0x00)
```

Unfortunately it doesn't say anything about leading zeros, but the example does not show leading zero, which does leave a bit of uncertainty but strongly indicates Amazon's pkcs#11 driver is buggy (all other software tokens seem to be doing the right thing).

I think the correct process would be to open a bug to amazon and an issue here, and link them, but at this point I see slim chances of them changing the driver, because it would probably cause compatibility issues with custom clients that now expect this broken representation ... sigh...

src/objects.c
```diff
@@ -2483,6 +2513,9 @@ static int cmp_attr(P11PROV_OBJ *key1, P11PROV_OBJ *key2,
     x2 = p11prov_obj_get_attr(key2, attr);
     if (!x1 || !x2) {
         return RET_OSSL_ERR;
+    }    
+    if(is_BN(attr)){
```
simo5, line 2517: I do not think we need to play magic here, instead of trying to autodetect random cases please drop is_BN and add a cmp_bn_attr() function, then fix the callers to call the expected function based on the attribute they want to compare.

simo5 (changes requested):
Please run `make style-check` / `make style-check-fix` to fix the style issues.

src/objects.c
bashmachnikov: ok

bashmachnikov:
> Unfortunately it doesn't say anything about leading zeros, but the example does not show leading zero, which does leave a bit of uncertainty but strongly indicates Amazon's pkcs#11 driver is buggy (all other software tokens seem to be doing the right thing).

We have similar behavior with AWS cloudHSM and lunaHSM. (this change fix the problem for both drivers)

src/objects.c
```diff
@@ -2474,6 +2474,29 @@ CK_ATTRIBUTE *p11prov_obj_get_ec_public_raw(P11PROV_OBJ *key)
     return pub_key;
 }
 
+static int cmp_bn_attr(CK_ATTRIBUTE *x1, CK_ATTRIBUTE *x2)
```
simo5, line 2477: can you pass P11PROV_OBJ *key as argument here, so the:
        x1 = p11prov_obj_get_attr(pub_key1, CKA_MODULUS);
        x2 = p11prov_obj_get_attr(pub_key2, CKA_MODULUS);

invocation are done within this function?

I want to keep the signatures of cmp_attr and comp_bn_attr consistent

Otherwise LGTM

simo5 (changes requested):

simo5:
Also please feel free to rebase on latest main and squash all commits into one.

simo5 (approved):
LGTM

simo5:
thank you!

Issue #547, opened 2025-03-14
Don't login to tokens when reading public keys


The-Mule:
**Describe the bug**

Commit 16b491c revealed the problem that the provider attempts to log in when reading public keys. As a consequence of this, a token asks for a PIN even though we are only trying to extract public key from it:

```
## Test fetching public keys without PIN in config files
openssl pkey -in $PUBURI -pubin -pubout -out ${TMPPDIR}/rsa.pub.nopin.pem
Enter PIN for PKCS#11 Token (Slot 2 - NSS User Private Key and Certificate Services):
```

**To Reproduce**

This is currently not reproducible in CI since there is no tty and that prevents PIN prompt. You can easily reproduce it running provider self-test locally - in tbasic test there will be a timeout because of PIN prompt.

**Expected behavior**

When extracting public key no PIN is requested.

**Operating environment (please complete the following information):**

RHEL-10, Fedora-41, CentOS-Stream 10

**Token and application used (please complete the following information):**

All current SW tokens used in the self-test.

**Additional context**

We can perhaps tun nopin tests in tbasic via expect to ensure the problem is gone (that will simulate having TTY I think).

simo5:
@Jakuje can you take a look at how to deal with this one given you are familiar with the code that introduced it?

PR #549, opened 2025-03-18
Prevent login prompt when reading public RSA keys

Jakuje:
#### Description

The #547 describes an issue that since 16b491c  the reading of public RSA keys prompts for a PIN. This is due to search for allowed mechanisms on associated private key, which is done in any case the public key does not have allowed_mechanisms attribute.

This PR has a reproducer for #547 and a partial fix for tokens not supporting allowed mechanisms.

For the remaining, it will likely need some more digging or workaround as we should login if the pin is provided by some means, but we should not if it is not. I did not manage to express this yet in the code.

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -1289,7 +1289,7 @@ P11PROV_OBJ *p11prov_obj_find_associated(P11PROV_OBJ *obj,
     slotid = p11prov_obj_get_slotid(obj);
 
     ret = p11prov_get_session(obj->ctx, &slotid, NULL, NULL,
-                              CK_UNAVAILABLE_INFORMATION, NULL, NULL, true,
+                              CK_UNAVAILABLE_INFORMATION, NULL, NULL, false,
                               false, &session);
     if (ret != CKR_OK) {
         goto done;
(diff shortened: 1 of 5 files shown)


Jakuje:
The last attempt works everywhere but on Ubuntu with OpenSSL 3.0.13, which looks like missing some code path to convert the private key to public key in the x509 command.

src/objects.c
```diff
@@ -1274,7 +1289,7 @@ P11PROV_OBJ *p11prov_obj_find_associated(P11PROV_OBJ *obj,
     slotid = p11prov_obj_get_slotid(obj);
 
     ret = p11prov_get_session(obj->ctx, &slotid, NULL, NULL,
-                              CK_UNAVAILABLE_INFORMATION, NULL, NULL, true,
+                              CK_UNAVAILABLE_INFORMATION, NULL, NULL, false,
```
simo5, line 1292: So the "fix" boils down to not logging in?
This may work in some cases if you are already logged in for some reason, but will fail in cases where no previous login was attempted.
I think we need a test cases that carefully manages how login is done and tries a public key + find associated and shows how gracefully (or not) this fails.

simo5 (dismissed):
I have no objection to any individual commit, but taken in aggregate I think this begs for a test that carfully exercises find_associated w/o a previous login.

src/objects.c
simo5: On second thought this might be sufficient.

Because most of the cases of finding associated keys go from private to public or certificate to public anyway, therefore either you are already logged in or probably do not need to be.

Probably ok for now.

PR #550, opened 2025-03-18
CI: Add testing with OpenSSL master branch

The-Mule:
#### Description

This tests provider build and it tests (for softokn and softhsm tokens) when openssl is built from its git master branch. It caches openssl installation in /opt directory and the cache is valid for one day.  

#### Checklist

- [ ] Code modified for feature (N/A)
- [x] Test suite updated with functionality tests (N/A)
- [ ] Test suite updated with negative tests (N/A)
- [ ] Documentation updated (N/A)

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/openssl.yml
+++ b/.github/workflows/openssl.yml
@@ -1,0 +1,83 @@
+---
+name: OpenSSL master
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+
+jobs:
+  build:
+    name: CI with openssl master 
+    runs-on: ubuntu-22.04
+    container: fedora:latest
+    steps:
+      - name: Get Date for DNF cache entry
+        id: get-date-dnf
+        run: |
+          echo "date=$(/bin/date -u "+%Y%V")" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: DNF cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            /var/cache/libdnf5
+          key: ${{ runner.os }}-dnf-${{ steps.get-date-dnf.outputs.date }}
+
+      - name: Install Dependencies
+        run: |
+            dnf -y install perl-FindBin perl-IPC-Cmd perl-File-Compare \
+              perl-File-Copy perl-Pod-Html clang git meson cargo expect \
+              pkgconf-pkg-config opensc p11-kit-devel gcc g++ sqlite-devel \
+              python3-six which cmake nss-softokn nss-tools nss-softokn-devel \
+              nss-devel softhsm
+
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Get Date for /opt cache entry
+        id: get-date-opt
+        run: |
+          echo "date=$(/bin/date -u "+%Y%m%d")" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: /opt cache
+        id: cache-opt
+        uses: actions/cache@v4
+        with:
+          path: /opt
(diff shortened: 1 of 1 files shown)


.github/workflows/openssl.yml
```diff
@@ -0,0 +1,82 @@
...
+        run: |
+          echo "date=$(/bin/date -u "+%Y%V")" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: DNF cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            /var/cache/libdnf5
+          key: ${{ runner.os }}-dnf-${{ steps.get-date.outputs.date }}
```
Jakuje, line 27:
```suggestion
          key: ${{ runner.os }}-dnf-${{ steps.get-date-dnf.outputs.date }}
```

.github/workflows/openssl.yml
```diff
@@ -0,0 +1,82 @@
...
+        run: |
+          git clone https://github.com/openssl/openssl.git
+
+      - name: Config, build and install OpenSSL
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: |
+          cd openssl
+          ./config --prefix=/opt && make && make install_sw
+
+      - name: Setup and build
```
Jakuje, line 63:
```suggestion
      - name: Setup and build pkcs11-provider
```

.github/workflows/openssl.yml
Jakuje: The pipeline looks like it works to restore the cache, but it fails to store the cache in `Post DNF cache` stage::
```
Cache saved with key: Linux-dnf-
```

.github/workflows/openssl.yml
The-Mule: Yes, this looks better, fixed.

The-Mule:
There is still one thing to be resolved - to have a cached /opt available in PRs we need to run action in main branch somehow.

The-Mule:
There's one more issue in the action causing that /opt cache is not used, I'll fix that quickly.

The-Mule:
Fixed, now it uses /opt cache correctly (runtime drops from 8 minutes to 1 minute with the cache).

Jakuje:
> There is still one thing to be resolved - to have a cached /opt available in PRs we need to run action in main branch somehow.

The caching should work also in PRs. The only difference is that if you create cache from PR, it will be available only for that PR, while the caches from `main` should get inherited to the PRs.

.github/workflows/openssl.yml
```diff
@@ -0,0 +1,83 @@
...
+            /var/cache/libdnf5
+          key: ${{ runner.os }}-dnf-${{ steps.get-date-dnf.outputs.date }}
+
+      - name: Install Dependencies
+        run: |
+            dnf -y install perl-FindBin perl-IPC-Cmd perl-File-Compare \
+              perl-File-Copy perl-Pod-Html clang git meson cargo expect \
+              pkgconf-pkg-config opensc p11-kit-devel gcc g++ sqlite-devel \
+              python3-six which cmake nss-softokn nss-tools nss-softokn-devel \
+              nss-devel softhsm
```
Jakuje, line 35: I think this cache should have different key than the one in kryoptic as it caches different DNF packages.

The-Mule: Right, they are not the same, fixed in [bc2f565](https://github.com/latchset/pkcs11-provider/pull/550/commits/bc2f5657ede5d81553b14479192e36a93e21cabd).

Jakuje (approved):
lgtm. I do not think coverity is needed for this

simo5:
Thanks!

PR #553, opened 2025-03-27
Add support for EVP_SKEY with AES cipher

simo5:
#### Description

Initial code to add support for the new EVP_SKEY facility in OpenSSL 3.5 which allows to use opaque symmetric keys.

This code was used in aiding development of the OpenSSL 3.5 feature.

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:


simo5:
I was finally able to make the openssl-master build use kryoptic and test skey ... and there is a segfault :)
So this will need some more work apparently

src/skeymgmt.c
```diff
@@ -0,0 +1,290 @@
...
+        { CKA_VALUE_LEN, &keylen, sizeof(keylen) },
+        { 0 },
+    };
+    size_t tmax = sizeof(tmpl) / sizeof(CK_ATTRIBUTE);
+    size_t tsize = 1;
+    P11PROV_URI *uri = NULL;
+    int ephemeral = 1; /* by default we create session keys */
+    /* all flags by default */
+    CK_FLAGS usage = CKF_ENCRYPT | CKF_DECRYPT | CKF_SIGN | CKF_VERIFY
+                     | CKF_WRAP | CKF_UNWRAP | CKF_DECRYPT;
```
Jakuje, line 132:
```suggestion
    CK_FLAGS usage = CKF_ENCRYPT | CKF_DECRYPT | CKF_SIGN | CKF_VERIFY
                     | CKF_WRAP | CKF_UNWRAP | CKF_DERIVE;
```

src/skeymgmt.c
```diff
@@ -0,0 +1,290 @@
...
+    p = OSSL_PARAM_locate_const(params, OSSL_SKEY_PARAM_KEY_LENGTH);
+    if (p) {
+        size_t len;
+        ret = OSSL_PARAM_get_size_t(p, &len);
+        if (ret != RET_OSSL_OK) {
+            P11PROV_raise(ctx, rv, "Invalid length parameter type");
+            goto done;
+        }
+        if (len != 16 && len != 24 && len != 32) {
+            P11PROV_raise(ctx, rv, "Invalid length parameter value");
```
Jakuje, line 179: I would be more specific in the error messages. Using just `length` is quite generic here:
```suggestion
            P11PROV_raise(ctx, rv, "Invalid key length parameter type");
            goto done;
        }
        if (len != 16 && len != 24 && len != 32) {
            P11PROV_raise(ctx, rv, "Invalid key length parameter value");
```

src/skeymgmt.c
```diff
@@ -0,0 +1,290 @@
...
+            P11PROV_raise(ctx, rv, "Invalid ephemeral parameter value");
+            goto done;
+        }
+    }
+
+    if (ephemeral) {
+        p11prov_set_attr_bool(&tmpl[tsize++], CKA_TOKEN, false);
+    } else {
+        p11prov_set_attr_bool(&tmpl[tsize++], CKA_TOKEN, true);
+    }
```
Jakuje, line 203: simplify here? Or just push that into the tmpl at the definition type, same way as the key length?
```suggestion
    p11prov_set_attr_bool(&tmpl[tsize++], CKA_TOKEN, !ephemeral);
```

src/store.c
```diff
@@ -343,6 +348,9 @@ static int p11prov_store_load(void *pctx, OSSL_CALLBACK *object_cb,
                 }
             }
             break;
+        default:
+            P11PROV_debug("Usnsupported object of class %x, skipping!", class);
```
Jakuje, line 352:
```suggestion
            P11PROV_debug("Unsupported object of class %x, skipping!", class);
```

tests/meson.build
```diff
@@ -158,6 +158,7 @@ tests = {
   'genkey': {'suites': all_suites},
   'pkey': {'suites': all_suites},
   'session': {'suites': all_suites},
+  'skey': {'suites': ['kryoptic', 'kryoptic.nss']},
```
Jakuje, line 161: the softhsm/sofotkn do not work here? It would be good to add a notes here when not all suites are supported why it is.

tests/setup.sh
```diff
@@ -24,6 +24,7 @@ OPENSC_VERSION=$(opensc-tool -i | grep OpenSC | sed -e "s/OpenSC 0\.\([0-9]*\).*
 if [[ "$OPENSC_VERSION" -le "25" ]]; then
     SUPPORT_ED448=0
 fi
+SUPPORT_ED448=0
```
Jakuje, line 27: some testing artifact that should not made it to the final PR?

src/cipher.c
```diff
@@ -0,0 +1,689 @@
...
+        }
+        param_as_iv = true;
+        break;
+
+    case CKM_AES_CBC_PAD:
+        if (!ctx->pad) {
+            ctx->mech.mechanism = CKM_AES_CBC;
+        }
+        param_as_iv = true;
+        break;
```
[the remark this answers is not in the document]
simo5: I prefer to maintain them as distinct branches.

src/interface.c
```diff
@@ -142,12 +147,17 @@ static void populate_interface(P11PROV_INTERFACE *intf, CK_INTERFACE *ck_intf)
...
     ASSIGN_FN(GenerateKeyPair);
     ASSIGN_FN(DeriveKey);
     ASSIGN_FN(SeedRandom);
     ASSIGN_FN(GenerateRandom);
     if (intf->version.major == 3) {
         ASSIGN_FN_3_0(GetInterface);
+        ASSIGN_FN_3_0(EncryptUpdate);
+        ASSIGN_FN_3_0(EncryptFinal);
+        ASSIGN_FN_3_0(DecryptUpdate);
+        ASSIGN_FN_3_0(DecryptFinal);
```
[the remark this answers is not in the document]
simo5: I do not know that I thought of it more than 3 seconds :)
Will re-check.

src/provider.c
```diff
@@ -849,6 +857,12 @@ static CK_RV alg_set_op(OSSL_ALGORITHM **op, int idx, OSSL_ALGORITHM *alg)
         CKM_ECDSA_SHA384, CKM_ECDSA_SHA512, CKM_ECDSA_SHA3_224, \
         CKM_ECDSA_SHA3_256, CKM_ECDSA_SHA3_384, CKM_ECDSA_SHA3_512
 
+#if SKEY_SUPPORT == 1
+#define AES_MECHS \
+    CKM_AES_ECB, CKM_AES_CBC, CKM_AES_OFB, CKM_AES_CFB128, CKM_AES_CFB1, \
```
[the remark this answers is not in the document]
simo5: probably

tests/setup.sh
simo5: good catch

src/skeymgmt.c
```diff
@@ -0,0 +1,290 @@
...
+static const OSSL_PARAM *p11prov_aes_imp_settable_params(void *provctx)
+{
+    return aes_import_params;
+}
+
+static const OSSL_PARAM aes_generate_params[] = {
+    OSSL_PARAM_size_t(OSSL_SKEY_PARAM_KEY_LENGTH, NULL),
+    OSSL_PARAM_utf8_string(P11PROV_PARAM_URI, NULL, 0),
+    OSSL_PARAM_octet_string(P11PROV_PARAM_KEY_USAGE, NULL, 0),
+    OSSL_PARAM_int(P11PROV_PARAM_EPHEMERAL, NULL), OSSL_PARAM_END
```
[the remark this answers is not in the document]
simo5: You have to complain to rustfmt not me :)

src/provider.h
```diff
@@ -167,6 +239,11 @@ int p11prov_pop_error_to_mark(P11PROV_CTX *ctx);
 #include "random.h"
 #include "pk11_uri.h"
 
+#if SKEY_SUPPORT == 1
+#include "cipher.h"
```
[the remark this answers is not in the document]
simo5: only in the if clause, good catch

tests/meson.build
simo5: ok softhsm deadlocks in openssl enc
and softoken fails key generation with CKR_USER_TYPE_INVALID ...

src/skeymgmt.c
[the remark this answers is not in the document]
simo5: lol yeah been working cross the two languages too much :-)
but same thing, it is the formatter that decides, and as we agreed way back then, no exceptions! :-)

simo5:
Covscan revealed a single issue with a printf argument type (usual int vs long), fixed and pushed, marking covscan as ok.

tests/meson.build
Jakuje: The sotfokn failure sounds weird and I think would deserve some investigation. But it can be done later.

Jakuje (approved):

simo5:
Before I merge, a question pops up in my mind, softhsm will now break badly whenever it is attempted to be used on an openssl-3.5 based build if a configuration is present that prefers using the pkcs11 provider for all operations.

Should we consider adding a quirk that will disable symmetric key operations ?

PR #554, opened 2025-03-27
CI: split openssl action into two jobs

The-Mule:
#### Description

Cache actions only caches when job succeeds and since we can expect CI tests to fails during the development work OpenSSL built in /opt might not get cached during multiple executions within the same PR.

Separating openssl build into its own job make it always succeed and cache is stored even if the tests from the second job fails.

#### Checklist

- [ ] Test suite updated with functionality tests (N/A)
- [ ] Test suite updated with negative tests (N/A)
- [ ] Documentation updated (N/A)

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/openssl.yml
+++ b/.github/workflows/openssl.yml
@@ -8,8 +8,8 @@ on:
     branches: ["main"]
 
 jobs:
-  build:
-    name: CI with openssl master 
+  setup:
+    name: Build OpenSSL from master branch 
     runs-on: ubuntu-22.04
     container: fedora:latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/openssl.yml
```diff
@@ -61,6 +58,51 @@ jobs:
...
+      - name: Get Date for /opt cache entry
+        id: get-date-opt
+        run: |
+          echo "date=$(/bin/date -u "+%Y%m%d")" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: /opt cache
+        id: cache-opt
+        uses: actions/cache@v4
+        with:
```
simo5, line 98: I think you should add the option `fail-on-cache-miss: true` here so that we do not get confused about why the build fails if the openssl cache somehow failed to be retrieved

I would also name the steos something explicitly like "Fetch openssl build from cache"

simo5 (changes requested):
Just a minor nit, otherwise looks good.

.github/workflows/openssl.yml
The-Mule: It's there - just a little below (line 102) but fail-on-cache-miss will be a better way, I'll fix that (and name as well).

The-Mule: Fixed.

simo5 (approved):

Issue #562, opened 2025-04-25
CMS decryption with explicit EC is broken


manison:
**Describe the bug**
After upgrading pkcs11-provider from version 0.3 to 1.0 I can no longer decrypt CMS encrypted with explicit EC.

**To Reproduce**

1. Run

```
openssl cms -decrypt -recip "pkcs11:token=T8406af1f9a2d43bb9cc03b926632c13;id=%2500%2507?pin-value=12345678" -in cms_encrypted.pem -inform pem -out cms_decrypted.txt
```
where _cms_encrypted.pem_ has been encrypted with explicit EC.

2. See error
 
```
Error decrypting CMS using private key
54890000:error:40800067:pkcs11:import_ec_params:reason(103):D:\Shared\openssl\pkcs11-provider-build-win\pkcs11-provider\src\objects.c:4102:Unknown curve
54890000:error:170000BC:CMS routines:ecdh_cms_decrypt:peer key error:crypto\cms\cms_ec.c:238:
```

**Expected behavior**
CMS is correctly decrypted.

**Operating environment (please complete the following information):**
 - OS: Debian
 - Version: 12

**Token and application used (please complete the following information):**
 - Device: SoftHSM2

**Additional context**

It seems that the bug is caused by simple inversion of equality condition in _objects.c_:

https://github.com/latchset/pkcs11-provider/blob/8f6b94409d4872265076df310492da1e5f6abdf7/src/objects.c#L4099-L4106

Simple fix

```diff
-    if (curve_nid == NID_undef) {
+    if (curve_nid != NID_undef) {
```

seems to solve the issue.

I tried to add a test to demonstrate the issue (it would be copy/paste from _tcms_ test with URIs replaced from EC to ECX), however I'm lost in the current matrix of platforms, tokens and enabled suites. It seems to me that there is currently no combination (platform supporting explicit EC, ie. not RHEL flavor, token supporting explicit EC and such combination being enabled for tests).

simo5:
Uhmm turns out SoftHSM CMS tests are disabled because SoftHSM seem to allow only CKD_NULL as the ECDH key derivation KDF, while CMS is trying to use a Digest derivation (CKD_DIGEST) ... investigating if there is a way to convince openssl to not use a KDF ...

simo5:
Apparently no way, openssl's CMD set SHA1 if no digest is selected ...

manison:
[...]

PR #563, opened 2025-04-25
Ensure CMS with Explicit EC works

simo5:
#### Description

Fix EC params import for Explicit EC Keys, which broke CMS operations.
Add tests to avoid future regressions, enable ECDH tests for SoftHSM2 where possible.
Changes how some tests are run in CI to test more cases even when some tests are known fails on some platforms (eg TLS tests on Debian with SoftHSM2).

#562 

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
       matrix:
         name: [fedora, debian, centos9, centos10, ubuntu, almalinux8]
         compiler: [gcc, clang]
-        token: [softokn, softhsm]
+        token: [softokn, softhsm, kryoptic]
         include:
           - name: fedora
             container: fedora:latest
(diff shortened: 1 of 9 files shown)


simo5:
I dropped the test, because it turns out no backend is compatible with it anyway :/

manison:
@simo5 this looks good. Is there anything else that prevents merging? Do we know why Ubuntu tests are failing?

simo5:
> @simo5 this looks good. Is there anything else that prevents merging? Do we know why Ubuntu tests are failing?

No, seem like Ubuntu's softhsm can't deal with ths new test, but then debian actually skips (in a bad way) softhsm tests.

Will push some test commits to see if I can narrow down things a little

simo5:
Looks like we have no distro where this test will actually run correctly.

The only distro that enable explicit ECs are Debian and Ubuntu, and it fails on both for some reason. On the other platforms the tets is just skipped...

simo5:
Interestingly this test passes if I run it with regular EC keys on softokn, but fails with kryoptic, so perhaps I'll find out what's up this way

manison:
I think they don't support KDF nor shared data (yet, see https://github.com/softhsm/SoftHSMv2/pull/599).

simo5:
Ok turned out to be the keys we were using with Explicit parameters were too small to generate a 256bit output ... doh!
Once I reduced the derivedkey size requested to 16 bytes it worked ... of course we'd figure this out much sooner if SoftHSM2 had returned the proper error message for this case (CKR_DATA_LEN_RANGE or CKR_KEY_SIZE_RANGE) ... sigh.

PR #564, opened 2025-04-28
Run all pkcs11 tests from libssh

Jakuje:
#### Description

Some time ago, I wrote the tests for libssh to use Ed25519 keys from pkcs11-provider:

https://gitlab.com/libssh/libssh-mirror/-/merge_requests/558

But it looks like the integration tests explicitly select just the tree tests that were here before so the new one was never picked up. This adjusts the regex to match all the pkcs11 uri tests in libssh testsuite

In libssh tests this one test is failing though (unexpectedly prompting for pin interactively)  so I am trying to figure out what is going on there and if the libssh or pkcs11-provider needs a fix. We need to extend this coverage anyway though.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/distcheck.yml
+++ b/.github/workflows/distcheck.yml
@@ -12,19 +12,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        name: [fedora, debian, centos]
+        name: [fedora, debian, centos9, centos10]
         include:
           - name: fedora
             container: fedora:latest
           - name: debian
             container: debian:sid
-          - name: centos
+          - name: centos9
             container: quay.io/centos/centos:stream9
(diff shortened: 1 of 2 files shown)


Jakuje:
Sounds like the libssh integration tests work just ok with current pkcs11-provider so the problem is likely the old pkcs11-provider in Fedora 41 images we have in libssh upstream. Trying to build the new one based on Fedora 42 now (which should have more recent pkcs11-provider).

Jakuje:
(the centos10 change was something I probably wanted to submit some time ago but never got to that so squeezing it here)

The-Mule (approved):
LGTM

PR #567, opened 2025-05-06
Add support for HMAC

ilie-halip-nxp:
#### Description

In the context of TLS 1.3, HMAC is used to calculate the `verify_data` with the Finished key. Right now, OpenSSL falls back to it's own internal HMAC.

Add an HMAC implementation that calls into the PKCS11 library to do the HMAC calculation. TLS1.3 does not truncate the HMAC length, so use `CKM_XXX_HMAC` mechanisms without an explicit length.

Add tests that compare the output from PKCS11 to the output of OpenSSL. Since PKCS11 libs may not have implementations for all the mechanisms in the specification, define a subset of "mandatory" HMACs that should be supported by all the test suites.

In the process, I had to add 2 new attributes for the generic secret key:
- `CKA_SIGN`: needed to be able to call `C_SignInit` with the key
- `CKA_PRIVATE`: `softhsm` seems to require this attribute for everything to work fine - I guess this could be removed, but the tests would have to be disabled for softhsm

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/mac.c
+++ b/src/mac.c
@@ -1,0 +1,356 @@
+// SPDX-License-Identifier: Apache-2.0
+/*
+ * Copyright 2025 NXP
+ */
+
+#include "provider.h"
+#include <string.h>
+
+struct p11prov_mac_ctx {
+    P11PROV_CTX *provctx;
+
+    P11PROV_OBJ *key;
+    P11PROV_SESSION *session;
+
+    CK_MECHANISM_TYPE mech;
+    CK_MECHANISM_TYPE hash_mech;
+};
+
+typedef struct p11prov_mac_ctx P11PROV_MAC_CTX;
+
+#define HMAC_MECHANISM(hash_mech) \
+    { \
+        .hash = CKM_##hash_mech, .mech = CKM_##hash_mech##_HMAC \
+    }
+
+static const struct {
+    CK_MECHANISM_TYPE mech;
+    CK_MECHANISM_TYPE hash;
+} hmac_mechanisms[] = { HMAC_MECHANISM(SHA_1),
+                        HMAC_MECHANISM(SHA224),
+                        HMAC_MECHANISM(SHA256),
+                        HMAC_MECHANISM(SHA384),
+                        HMAC_MECHANISM(SHA512),
+
+                        HMAC_MECHANISM(BLAKE2B_160),
+                        HMAC_MECHANISM(BLAKE2B_256),
+                        HMAC_MECHANISM(BLAKE2B_384),
+                        HMAC_MECHANISM(BLAKE2B_512),
+
+                        HMAC_MECHANISM(MD2),
+                        HMAC_MECHANISM(MD5),
+
+                        HMAC_MECHANISM(RIPEMD128),
+                        HMAC_MECHANISM(RIPEMD160),
+
+                        HMAC_MECHANISM(SHA3_224),
+                        HMAC_MECHANISM(SHA3_256),
+                        HMAC_MECHANISM(SHA3_384),
+                        HMAC_MECHANISM(SHA3_512),
+
+                        HMAC_MECHANISM(SHA512_224),
+                        HMAC_MECHANISM(SHA512_256),
+                        HMAC_MECHANISM(SHA512_T),
+
+                        { .hash = CK_UNAVAILABLE_INFORMATION,
+                          .mech = CK_UNAVAILABLE_INFORMATION } };
+
+static void *p11prov_mac_newctx(void *provctx)
+{
+    P11PROV_CTX *ctx = (P11PROV_CTX *)provctx;
+    P11PROV_MAC_CTX *macctx;
+
+    P11PROV_debug("mac newctx");
+
+    macctx = OPENSSL_zalloc(sizeof(*macctx));
+    if (macctx == NULL) {
+        return NULL;
+    }
+
+    macctx->provctx = ctx;
+
+    /* default mechanism */
+    macctx->mech = CKM_SHA_1_HMAC;
+    macctx->hash_mech = CKM_SHA_1;
+
+    return macctx;
+}
+
+static void p11prov_mac_freectx(void *ctx)
+{
+    P11PROV_MAC_CTX *macctx = (P11PROV_MAC_CTX *)ctx;
+
+    P11PROV_debug("mac freectx (ctx:%p)", ctx);
+
+    p11prov_obj_free(macctx->key);
+
+    if (macctx->session) p11prov_return_session(macctx->session);
+
+    OPENSSL_clear_free(macctx, sizeof(*macctx));
+}
+
+static int p11prov_hmac_get_ctx_params(void *ctx, OSSL_PARAM params[])
+{
+    P11PROV_MAC_CTX *macctx = (P11PROV_MAC_CTX *)ctx;
+    OSSL_PARAM *p;
+    CK_RV ret;
+    size_t digest_size = 0;
+    size_t block_size = 0;
+
+    P11PROV_debug("mac get ctx params (ctx=%p, params=%p)", macctx, params);
+
+    if (params == NULL) return RET_OSSL_OK;
+
+    p = OSSL_PARAM_locate(params, OSSL_MAC_PARAM_SIZE);
+    if (p) {
+        ret = p11prov_digest_get_digest_size(macctx->hash_mech, &digest_size);
+        if (ret != CKR_OK || digest_size == 0) {
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST);
+            return RET_OSSL_ERR;
+        }
+
+        OSSL_PARAM_set_size_t(p, digest_size);
+    }
+
+    p = OSSL_PARAM_locate(params, OSSL_MAC_PARAM_BLOCK_SIZE);
+    if (p) {
+        ret = p11prov_digest_get_block_size(macctx->hash_mech, &block_size);
+        if (ret != CKR_OK || block_size == 0) {
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST);
+            return RET_OSSL_ERR;
+        }
+
+        OSSL_PARAM_set_size_t(p, block_size);
+    }
+
+    return RET_OSSL_OK;
+}
+
+static const OSSL_PARAM *p11prov_hmac_gettable_ctx_params(void *ctx,
+                                                          void *provctx)
+{
+    static const OSSL_PARAM known_gettable_ctx_params[] = {
+        OSSL_PARAM_size_t(OSSL_MAC_PARAM_SIZE, NULL),
+        OSSL_PARAM_size_t(OSSL_MAC_PARAM_BLOCK_SIZE, NULL), OSSL_PARAM_END
+    };
+    return known_gettable_ctx_params;
+}
+
+static CK_RV inner_mac_key(P11PROV_MAC_CTX *macctx, const uint8_t *key,
+                           size_t keylen, P11PROV_OBJ **keyobj)
+{
+    CK_SLOT_ID slotid = CK_UNAVAILABLE_INFORMATION;
+    CK_RV ret;
+
+    if (macctx->session == NULL) {
+        ret = p11prov_get_session(macctx->provctx, &slotid, NULL, NULL,
+                                  macctx->mech, NULL, NULL, false, false,
+                                  &macctx->session);
+        if (ret != CKR_OK) {
+            return ret;
+        }
+    }
+    if (macctx->session == NULL) {
+        return CKR_SESSION_HANDLE_INVALID;
+    }
+
+    *keyobj = p11prov_create_secret_key(macctx->provctx, macctx->session, true,
+                                        (unsigned char *)key, keylen);
+    if (*keyobj == NULL) {
+        return CKR_KEY_HANDLE_INVALID;
+    }
+    return CKR_OK;
+}
+
+static int p11prov_hmac_set_ctx_params(void *ctx, const OSSL_PARAM params[])
+{
+    P11PROV_MAC_CTX *macctx = (P11PROV_MAC_CTX *)ctx;
+    const OSSL_PARAM *p;
+    int ret;
+
+    P11PROV_debug("mac set ctx params (ctx=%p, params=%p)", macctx, params);
+
+    p = OSSL_PARAM_locate_const(params, OSSL_MAC_PARAM_TLS_DATA_SIZE);
+    if (p) {
+        /*
+         * TLS_DATA_SIZE parameter should activate some special handling
+         * for TLS records that are mac-then-encrypt and have variable length
+         * padding (i.e. CBC ciphersuites).
+         *
+         * For now, just raise "not supported" error.
+         */
+        P11PROV_raise(ctx, CKR_ARGUMENTS_BAD,
+                      "TLS_DATA_SIZE is not supported!");
+        return RET_OSSL_ERR;
+    }
+
+    p = OSSL_PARAM_locate_const(params, OSSL_MAC_PARAM_DIGEST);
+    if (p) {
+        const char *digest = NULL;
+        CK_RV rv;
+
+        ret = OSSL_PARAM_get_utf8_string_ptr(p, &digest);
+        if (ret != RET_OSSL_OK) {
+            return ret;
+        }
+
+        rv = p11prov_digest_get_by_name(digest, &macctx->hash_mech);
+        if (rv != CKR_OK) {
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST);
+            return RET_OSSL_ERR;
+        }
+        P11PROV_debug("set digest to %lu", macctx->hash_mech);
+
+        macctx->mech = CK_UNAVAILABLE_INFORMATION;
+        for (int i = 0; hmac_mechanisms[i].hash != CK_UNAVAILABLE_INFORMATION;
+             i++) {
+            if (macctx->hash_mech == hmac_mechanisms[i].hash) {
+                macctx->mech = hmac_mechanisms[i].mech;
+                break;
+            }
+        }
+
+        if (macctx->mech == CK_UNAVAILABLE_INFORMATION) {
+            P11PROV_debug("No associated HMAC mechanism");
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST);
+            return RET_OSSL_ERR;
+        }
+
+        P11PROV_debug("set MAC to %lu", macctx->mech);
+    }
+
+    p = OSSL_PARAM_locate_const(params, OSSL_MAC_PARAM_KEY);
+    if (p) {
+        const void *key = NULL;
+        size_t key_len;
+        CK_RV rv;
+
+        ret = OSSL_PARAM_get_octet_string_ptr(p, &key, &key_len);
+        if (ret != RET_OSSL_OK) {
+            return ret;
+        }
+
+        /* Create Session and key from key material */
+        p11prov_obj_free(macctx->key);
+        rv = inner_mac_key(macctx, key, key_len, &macctx->key);
+        if (rv != CKR_OK) {
+            return RET_OSSL_ERR;
+        }
+        P11PROV_debug("set key (len:%lu)", key_len);
+    }
+
+    return RET_OSSL_OK;
+}
+
+static const OSSL_PARAM *p11prov_hmac_settable_ctx_params(void *ctx,
+                                                          void *provctx)
+{
+    static const OSSL_PARAM known_settable_ctx_params[] = {
+        OSSL_PARAM_utf8_string(OSSL_MAC_PARAM_DIGEST, NULL, 0),
+        OSSL_PARAM_utf8_string(OSSL_MAC_PARAM_PROPERTIES, NULL, 0),
+        OSSL_PARAM_octet_string(OSSL_MAC_PARAM_KEY, NULL, 0),
+        OSSL_PARAM_size_t(OSSL_MAC_PARAM_TLS_DATA_SIZE, NULL), OSSL_PARAM_END
+    };
+    return known_settable_ctx_params;
+}
+
+static int p11prov_hmac_init(void *ctx, const unsigned char *key, size_t keylen,
+                             const OSSL_PARAM params[])
+{
+    P11PROV_MAC_CTX *macctx = (P11PROV_MAC_CTX *)ctx;
+    CK_RV ret = CKR_OK;
+    CK_OBJECT_HANDLE pkey_handle = CK_INVALID_HANDLE;
+    CK_MECHANISM mech = { 0 };
+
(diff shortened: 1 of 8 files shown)


src/mac.c
```diff
@@ -0,0 +1,356 @@
...
+#include <string.h>
+
+struct p11prov_mac_ctx {
+    P11PROV_CTX *provctx;
+
+    P11PROV_OBJ *key;
+    P11PROV_SESSION *session;
+
+    CK_MECHANISM_TYPE mech;
+    CK_MECHANISM_TYPE hash_mech;
```
simo5, line 16: Sounds like hash_mech is only used to query the digst_map() when obtaining context params.

Instead of storing this type provide a helper function (or even just a macro) called hash_mech() that is just a case switch that returns the corresponding hash mechanism given an hmac mechanism.

Wherever you currently use provctx->hash_mech just call hash_mech(provctx->hmac_mech)

src/mac.c
```diff
@@ -0,0 +1,356 @@
...
+                        HMAC_MECHANISM(SHA256),
+                        HMAC_MECHANISM(SHA384),
+                        HMAC_MECHANISM(SHA512),
+
+                        HMAC_MECHANISM(BLAKE2B_160),
+                        HMAC_MECHANISM(BLAKE2B_256),
+                        HMAC_MECHANISM(BLAKE2B_384),
+                        HMAC_MECHANISM(BLAKE2B_512),
+
+                        HMAC_MECHANISM(MD2),
```
simo5, line 40: drop md2 and md5 as well as ripmed, we do not want to support obsolete and insecure hash functions really, and they are not needed for tls

src/objects.c
```diff
@@ -1526,12 +1526,15 @@ P11PROV_OBJ *p11prov_create_secret_key(P11PROV_CTX *provctx,
...
+    CK_BBOOL val_false = CK_FALSE;
     CK_BBOOL val_token = session_key ? CK_FALSE : CK_TRUE;
-    CK_ATTRIBUTE key_template[5] = {
+    CK_ATTRIBUTE key_template[7] = {
         { CKA_CLASS, &key_class, sizeof(key_class) },
         { CKA_KEY_TYPE, &key_type, sizeof(key_type) },
         { CKA_TOKEN, &val_token, sizeof(val_token) },
         { CKA_DERIVE, &val_true, sizeof(val_true) },
+        { CKA_SIGN, &val_true, sizeof(val_true) },
+        { CKA_PRIVATE, &val_false, sizeof(val_false) },
```
simo5, line 1537: Instead of adding these values to p11prov_create_secret_key() you should probably add a p11prov_create_mac_key function.

We do not want all of the secret keys to be enabled for signing unconditionally.
(and yes we should probably rename p11prov_create_secret_key -> p11prov_create_derive_key)

src/provider.c
```diff
@@ -1133,6 +1144,49 @@ static CK_RV operations_init(P11PROV_CTX *ctx)
...
 #endif
+            case CKM_BLAKE2B_160_HMAC:
+            case CKM_BLAKE2B_160_HMAC_GENERAL:
+            case CKM_BLAKE2B_256_HMAC:
+            case CKM_BLAKE2B_256_HMAC_GENERAL:
+            case CKM_BLAKE2B_384_HMAC:
+            case CKM_BLAKE2B_384_HMAC_GENERAL:
+            case CKM_BLAKE2B_512_HMAC:
+            case CKM_BLAKE2B_512_HMAC_GENERAL:
+            case CKM_MD2_HMAC:
```
simo5, line 1155: we probably do not want to support obsolete MAc mechanisms (see above)

simo5 (changes requested):
This looks like good work, and I have mostly nits, however I got to ask why do you care to execute the HMAC function on the token, given you are not using really a token generated key, but rather importing a key that openssl has access to already.

ilie-halip-nxp:
Hey Simo, thanks for the quick review!

> why do you care to execute the HMAC function on the token, given you are not using really a token generated key, but rather importing a key that openssl has access to already.

This is just to allow delegating the operation to a token that is a hardware crypto accelerator. But, I have some WIP changes to do TLS related operations entirely on the token, so this would be a prerequisite for those to work. Do you think such changes could be accepted, in future PRs?

simo5:
> Hey Simo, thanks for the quick review!
> 
> > why do you care to execute the HMAC function on the token, given you are not using really a token generated key, but rather importing a key that openssl has access to already.
> 
> This is just to allow delegating the operation to a token that is a hardware crypto accelerator. But, I have some WIP changes to do TLS related operations entirely on the token, so this would be a prerequisite for those to work. Do you think such changes could be accepted, in future PRs?

We are interested in that kind of functionality, yes, although it may require new functionality in OpenSSL as well. But we can collaborate in getting what's needed there, as we are also expanding on the EVP_SKEY stuff (which MAC code should really be able use once we add EVP_SKEY to MAC operations in OpenSSL)

src/mac.c
```diff
@@ -0,0 +1,431 @@
...
+                             (CK_BYTE_PTR)in, inl);
+    if (ret != CKR_OK) {
+        return RET_OSSL_ERR;
+    }
+
+    return RET_OSSL_OK;
+}
+
+static int p11prov_hmac_final(void *ctx, unsigned char *out, size_t *outl,
+                              size_t outsize)
```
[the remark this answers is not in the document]
ilie-halip-nxp: openssl provides a buffer that is large enough to hold the hmac output (`EVP_MAX_MD_SIZE * 2`): https://github.com/openssl/openssl/blob/8109618a1ce0103045329ed44f9d7a4ded8654be/ssl/tls13_enc.c#L309

src/mac.c
```diff
@@ -0,0 +1,431 @@
...
+
+    macctx = OPENSSL_zalloc(sizeof(*macctx));
+    if (macctx == NULL) {
+        return NULL;
+    }
+
+    macctx->provctx = ctx;
+
+    /* default mechanism */
+    macctx->hmac = CKM_SHA_1_HMAC;
```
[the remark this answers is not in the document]
ilie-halip-nxp: setting it to `CK_UNAVAILABLE_INFORMATION` seems fine, I don't think there's any scenario in which openssl does not set the digest

src/mac.c
```diff
@@ -0,0 +1,431 @@
...
+
+    CK_MECHANISM_TYPE hmac;
+};
+
+typedef struct p11prov_mac_ctx P11PROV_MAC_CTX;
+
+#define HMAC_MECHANISM(digest) \
+    { .hash = CKM_##digest, .hmac = CKM_##digest##_HMAC }, \
+    { \
+        .hash = CKM_##digest, .hmac = CKM_##digest##_HMAC_GENERAL \
```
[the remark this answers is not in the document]
ilie-halip-nxp: > doesn't look like the CKM_XX_HMAC_GENERAL mechanisms are ever returned

Yeah, I noticed the same thing right after I pushed that commit. I'll change this up.

src/objects.c
```diff
@@ -1589,6 +1589,81 @@ P11PROV_OBJ *p11prov_create_secret_key(P11PROV_CTX *provctx,
...
+                                    unsigned char *secret, size_t secretlen)
+{
+    CK_SESSION_HANDLE sess = CK_INVALID_HANDLE;
+    CK_SESSION_INFO session_info;
+    CK_OBJECT_CLASS key_class = CKO_SECRET_KEY;
+    CK_KEY_TYPE key_type = CKK_GENERIC_SECRET;
+    CK_BBOOL val_true = CK_TRUE;
+    CK_BBOOL val_false = CK_FALSE;
+    CK_BBOOL val_token = session_key ? CK_FALSE : CK_TRUE;
+    CK_ATTRIBUTE key_template[6] = {
```
Jakuje, line 1604: Please, use constants instead of the magic number 6 here and on line 1632

src/provider.c
```diff
@@ -916,9 +925,9 @@ static CK_RV operations_init(P11PROV_CTX *ctx)
                              DIGEST_MECHS,
                              CKM_EDDSA,
 #if SKEY_SUPPORT == 1
-                             AES_MECHS
+                             AES_MECHS,
 #endif
-    };
+                             HMAC_MECHS };
```
Jakuje, line 930: nit: I think if you add the trailing comma here, the clang will land the closing brace on the newline, which will make diffs of adding more elements easier to read.

src/mac.c
```diff
@@ -0,0 +1,448 @@
...
+    CK_MECHANISM_TYPE hmac;
+};
+
+typedef struct p11prov_mac_ctx P11PROV_MAC_CTX;
+
+static CK_ULONG digest_to_hmac(CK_ULONG digest)
+{
+    switch (digest) {
+    case CKM_SHA_1:
+        return CKM_SHA_1_HMAC;
```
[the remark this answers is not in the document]
ilie-halip-nxp: I would rather have a mapping array as well (maybe not as part of `src/digests.c`, that code should deal with digests only) and these functions should do a simple search very much like `p11prov_digest_get_name()`. I had such a mapping in the initial commit but removed it in the mean time; @simo5 what do you think?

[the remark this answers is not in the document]
simo5: I am not opposed to consolidate the mapping with digests in a single place as long as adding the hmac mechanisms to the map doe snot break anything else.

ilie-halip-nxp:
Hey, I believe I went through all the comments.

I see there are a couple of CI failures, maybe these are sporadic failures but I'm taking a look. Any other things which would I need to address before merging?

Jakuje:
> I see there are a couple of CI failures, maybe these are sporadic failures but I'm taking a look. Any other things which would I need to address before merging?

I think the failures are related to to openssl or something. Please, rebase your changes on top of current `main`.

src/objects.c
```diff
@@ -1589,6 +1589,82 @@ P11PROV_OBJ *p11prov_create_secret_key(P11PROV_CTX *provctx,
...
+    CK_OBJECT_CLASS key_class = CKO_SECRET_KEY;
+    CK_KEY_TYPE key_type = CKK_GENERIC_SECRET;
+    CK_BBOOL val_true = CK_TRUE;
+    CK_BBOOL val_false = CK_FALSE;
+    CK_BBOOL val_token = session_key ? CK_FALSE : CK_TRUE;
+    CK_ATTRIBUTE key_template[] = {
+        { CKA_CLASS, &key_class, sizeof(key_class) },
+        { CKA_KEY_TYPE, &key_type, sizeof(key_type) },
+        { CKA_TOKEN, &val_token, sizeof(val_token) },
+        { CKA_SIGN, &val_true, sizeof(val_true) },
```
simo5, line 1786: Should we also add CKA_VERIFY here?

I know we currently do not use C_Verify in this code, but seem prudent to mark the key as good for both operations, as eventually we may want to use it that way.

src/objects.c
```diff
@@ -1589,6 +1589,82 @@ P11PROV_OBJ *p11prov_create_secret_key(P11PROV_CTX *provctx,
...
+        { CKA_PRIVATE, &val_false, sizeof(val_false) },
```
simo5: Why is CKA_PRIVATE set to false here?

tests/thmac
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash -e
+# Copyright 2025 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+source "${TESTSSRCDIR}/helpers.sh"
+
+# Need to configure early loading, otherwise MACs might not be available
+sed "s/#pkcs11-module-load-behavior/pkcs11-module-load-behavior = early/" \
+    "${OPENSSL_CONF}" > "${OPENSSL_CONF}.early_load"
+OPENSSL_CONF=${OPENSSL_CONF}.early_load
```
simo5: Early loading seem odd to need, can you instead force all operations on the token? Otherwise there is no guarantee thes operations are done on the token during tests.
See other tests for the properties to set to force all ops on a token.

simo5 (commented):
Very close, just some minor nits.

src/objects.c
ilie-halip-nxp: Hm, that was a workaround for SoftHSM needing that value to be set explicitly (otherwise C_CreateObject would return CKR_USER_NOT_LOGGED_IN). Maybe doing the actual login would be more suitable.

src/objects.c
ilie-halip-nxp: Sure thing.

src/objects.c
simo5: if it needs to be stored in the token (your code allows nut does not exercise that code patch when session = false) you will need to login with most tokens anyway, so just login, yes.

tests/thmac
ilie-halip-nxp: Forcing all ops on the token leads to a hang because of a seemingly non-recursive mutex used by SoftHSM to lock the token. Here's a backtrace:
```
  * frame #0: 0x00007ffff7298f70 libc.so.6`__GI___lll_lock_wait [inlined] futex_wait(private=0, expected=2, futex_word=0x00005555556e89e0) at futex-internal.h:146:13
    frame #1: 0x00007ffff7298f5a libc.so.6`__GI___lll_lock_wait(futex=0x00005555556e89e0, private=0) at lowlevellock.c:49:7
    frame #2: 0x00007ffff72a0101 libc.so.6`___pthread_mutex_lock [inlined] lll_mutex_lock_optimized(mutex=0x00005555556e89e0) at pthread_mutex_lock.c:48:5
    frame #3: 0x00007ffff72a00f8 libc.so.6`___pthread_mutex_lock(mutex=0x00005555556e89e0) at pthread_mutex_lock.c:93:7
    frame #4: 0x00007ffff7571d59 libsofthsm2.so`OSLockMutex(mutex=0x00005555556e89e0) at osmutex.cpp:109:30
    frame #5: 0x00007ffff75719dd libsofthsm2.so`MutexFactory::LockMutex(this=0x00005555556b6970, mutex=0x00005555556e89e0) at MutexFactory.cpp:183:26
    frame #6: 0x00007ffff75715d1 libsofthsm2.so`Mutex::lock(this=0x00005555556e8710) at MutexFactory.cpp:61:50
    frame #7: 0x00007ffff7571669 libsofthsm2.so`MutexLocker::MutexLocker(this=0x00007fffffffc620, inMutex=0x00005555556e8710) at MutexFactory.cpp:82:32
    frame #8: 0x00007ffff75bf5b9 libsofthsm2.so`Token::isSOLoggedIn(this=0x00005555556e5300) at Token.cpp:95:29
    frame #9: 0x00007ffff75b9c28 libsofthsm2.so`SessionManager::openSession(this=0x00005555556e8320, slot=0x00005555556e4d50, flags=4, pApplication=0x00005555556e7df0, notify=(pkcs11.so`token_session_callback at session.c:44:1), phSession=0x00005555556e7e08) at SessionManager.cpp:83:58
    frame #10: 0x00007ffff7535af0 libsofthsm2.so`SoftHSM::C_OpenSession(this=0x00005555556d0c10, slotID=1146617904, flags=4, pApplication=0x00005555556e7df0, notify=(pkcs11.so`token_session_callback at session.c:44:1), phSession=0x00005555556e7e08) at SoftHSM.cpp:1414:40
    frame #11: 0x00007ffff750e07a libsofthsm2.so`C_OpenSession(slotID=1146617904, flags=4, pApplication=0x00005555556e7df0, notify=(pkcs11.so`token_session_callback at session.c:44:1), phSession=0x00005555556e7e08) at main.cpp:317:37
    frame #12: 0x00007ffff7dc2d69 pkcs11.so`p11prov_OpenSession(ctx=0x00005555556baa10, slotID=1146617904, flags=4, pApplication=0x00005555556e7df0, Notify=(pkcs11.so`token_session_callback at session.c:44:1), phSession=0x00005555556e7e08) at interface.gen.c:304:11
    frame #13: 0x00007ffff7de78e3 pkcs11.so`token_session_open(session=0x00005555556e7df0, flags=4) at session.c:76:19
    frame #14: 0x00007ffff7dea631 pkcs11.so`p11prov_get_session(provctx=0x00005555556baa10, slotid=0x00007fffffffc958, next_slotid=0x0000000000000000, uri=0x0000000000000000, mechtype=592, pw_cb=0x0000000000000000, pw_cbarg=0x0000000000000000, reqlogin=false, rw=false, _session=0x0000555555732470) at session.c:1023:19
    frame #15: 0x00007ffff7db41ca pkcs11.so`p11prov_digest_init(ctx=0x0000555555732460, params=0x0000000000000000) at digests.c:315:9
    frame #16: 0x00007ffff7846c9b libcrypto.so.3`evp_md_init_internal(ctx=0x0000555555728df0, type=0x0000555555731f00, params=0x0000000000000000, impl=0x0000000000000000) at digest.c:302:12
    frame #17: 0x00007ffff7846f8f libcrypto.so.3`EVP_DigestInit_ex(ctx=0x0000555555728df0, type=0x00007ffff7c8f8a0, impl=0x0000000000000000) at digest.c:382:12
    frame #18: 0x00007ffff75894ac libsofthsm2.so`OSSLEVPHashAlgorithm::hashInit(this=0x00005555556e8f10) at OSSLEVPHashAlgorithm.cpp:61:24
    frame #19: 0x00007ffff7596e48 libsofthsm2.so`RFC4880::PBEDeriveKey(password=0x00007fffffffcda0, salt=0x00007fffffffcb70, ppKey=0x00007fffffffcb58) at RFC4880.cpp:74:21
    frame #20: 0x00007ffff7597f7a libsofthsm2.so`SecureDataManager::login(this=0x00005555556e50e0, passphrase=0x00007fffffffcda0, encryptedKey=0x00005555556e50e8) at SecureDataManager.cpp:264:28
    frame #21: 0x00007ffff75983f5 libsofthsm2.so`SecureDataManager::loginUser(this=0x00005555556e50e0, userPIN=0x00007fffffffcda0) at SecureDataManager.cpp:317:30
    frame #22: 0x00007ffff75bfa28 libsofthsm2.so`Token::loginUser(this=0x00005555556e5300, pin=0x00007fffffffcda0) at Token.cpp:176:21
    frame #23: 0x00007ffff7535f53 libsofthsm2.so`SoftHSM::C_Login(this=0x00005555556d0c10, hSession=1, userType=1, pPin="12345678", ulPinLen=8) at SoftHSM.cpp:1538:25
    frame #24: 0x00007ffff750e328 libsofthsm2.so`C_Login(hSession=1, userType=1, pPin="12345678", ulPinLen=8) at main.cpp:407:31
    frame #25: 0x00007ffff7dc39ce pkcs11.so`p11prov_Login(ctx=0x00005555556baa10, hSession=1, userType=1, pPin="12345678", ulPinLen=8) at interface.gen.c:451:11
    frame #26: 0x00007ffff7de930a pkcs11.so`token_login(session=0x00005555556e7bb0, uri=0x0000000000000000, pw_cb=0x0000000000000000, pw_cbarg=0x0000000000000000, slot=0x0000555555712ec0, user_type=1) at session.c:565:11
    frame #27: 0x00007ffff7de9f3b pkcs11.so`slot_login(slot=0x0000555555712ec0, uri=0x0000000000000000, pw_cb=0x0000000000000000, pw_cbarg=0x0000000000000000, reqlogin=true, _session=0x0000000000000000) at session.c:834:15
    frame #28: 0x00007ffff7dea481 pkcs11.so`p11prov_get_session(provctx=0x00005555556baa10, slotid=0x00007fffffffd128, next_slotid=0x0000000000000000, uri=0x0000000000000000, mechtype=545, pw_cb=0x0000000000000000, pw_cbarg=0x0000000000000000, reqlogin=true, rw=false, _session=0x00005555556e6c90) at session.c:968:23
    frame #29: 0x00007ffff7dc07cf pkcs11.so`import_mac_key(macctx=0x00005555556e6c80, key="00112233445566778899aabbccddeeff", keylen=32, keyobj=0x00005555556e6c88) at mac.c:122:15
    frame #30: 0x00007ffff7dc0c3f pkcs11.so`p11prov_hmac_set_ctx_params(ctx=0x00005555556e6c80, params=0x00005555556e7d40) at mac.c:204:14
    frame #31: 0x00007ffff787c0ff libcrypto.so.3`EVP_MAC_CTX_set_params(ctx=0x00005555556ce2a0, params=0x00005555556e7d40) at mac_lib.c:220:16
    frame #32: 0x00005555555d7457 openssl`mac_main(argc=12, argv=0x00007fffffffd528) at mac.c:162:14
    frame #33: 0x00005555555dbe4a openssl`do_cmd(prog=0x00005555556b8860, argc=12, argv=0x00007fffffffd4d0) at openssl.c:428:16
    frame #34: 0x00005555555db9d2 openssl`main(argc=12, argv=0x00007fffffffd4d0) at openssl.c:309:19
    frame #35: 0x00007ffff722a1ca libc.so.6`__libc_start_call_main(main=(openssl`main at openssl.c:238:1), argc=13, argv=0x00007fffffffd4c8) at libc_start_call_main.h:58:16
    frame #36: 0x00007ffff722a28b libc.so.6`__libc_start_main_impl(main=(openssl`main at openssl.c:238:1), argc=13, argv=0x00007fffffffd4c8, init=<unavailable>, fini=<unavailable>, rtld_fini=<unavailable>, stack_end=0x00007fffffffd4b8) at libc-start.c:360:3
    frame #37: 0x00005555555a21a5 openssl`_start + 37
```

The mutex is locked during [Token::loginUser()](https://github.com/softhsm/SoftHSMv2/blob/develop/src/lib/slot_mgr/Token.cpp#L155) (frame #22) and is still locked when `Token::isSOLoggedIn()` is called (frame #8).

I got the test working without early loading or forcing all operations on the token, it was just a matter of initializing pkcs11 a bit earlier and adding HMAC to `static_operations_init()`.

simo5: softhsm is hopelessly broken like that in many places, feel free to exclude the test on softhsm

simo5: Note that it is likely you will never be able to use softhsm with this feature, if that is your target token you may want to consider switching to use kryoptic as token instead (there is a utility to export existing softhsm databases to its format too).

ilie-halip-nxp: My main target is NXP's SMW library (https://github.com/nxp-imx/imx-smw) which exposes a PKCS11 interface to a secure enclave that can do TLS derivation without exporting the key material, thus the need for the HMAC operation as in this PR.

If I do all operations on the token in the test as you mentioned, wouldn't both `openssl mac` commands do exactly the same thing?

simo5: When you say "both" what do you mean?

ilie-halip-nxp: The test script, `thmac`, runs `openssl mac` 1) with and 2) without `-propquuery ?provider=pkcs11`, then compares the outputs.

simo5: the results should always be the same, certainly

ilie-halip-nxp: Yes, and I saw in other tests that running all operations on the token implies setting
`default_properties = ?provider=pkcs11`
in openssl.cnf.

In this case, both `openssl mac` command would invoke the pkcs11 provider implementation. Is this right?

Jakuje:
Can you rebase on top of the current master? Rebuild should fix the centos10 issues. Then we are left with the softhsm issues.

tests/thmac
ilie-halip-nxp: > Early loading seem odd to need, can you instead force all operations on the token?

I may have misunderstood what you meant by this, could you take a look at the last commit(s) in this series of fixups? I'm not sure I went the right way with my changes. Many thanks!

PR #573, opened 2025-05-14
skeymgmt: add support to import skeyopt pkcs11_uri

mgrzeschik:
When using skeymgmt it is currently not possible to set skeyopt pkcs11_uri when importing a key. This is necessary to use a key that is already stored in the token. This patch implements a solution by searching for skey objects using the parameterized pkcs11_uri.

### DIFF
--- a/src/skeymgmt.c
+++ b/src/skeymgmt.c
@@ -22,6 +22,19 @@ static void p11prov_aes_free(void *key)
     p11prov_obj_free((P11PROV_OBJ *)key);
 }
 
+struct skey_cb_ctx {
+    P11PROV_OBJ *obj;
+};
+
+static CK_RV skey_ctx_obj(void *pctx, P11PROV_OBJ *obj)
+{
+    struct skey_cb_ctx *skey_ctx = (struct skey_cb_ctx *)pctx;
+
+    skey_ctx->obj = obj;
+
+    return CKR_OK;
+}
+
 static void *p11prov_aes_import(void *provctx, int selection,
                                 const OSSL_PARAM params[])
 {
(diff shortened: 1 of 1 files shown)


src/skeymgmt.c
```diff
@@ -22,25 +22,73 @@ static void p11prov_aes_free(void *key)
...
     P11PROV_debug("aes import");
 
     if (!ctx) {
         return NULL;
     }
 
-    if (!(selection & OSSL_SKEYMGMT_SELECT_SECRET_KEY)) {
-        /* TODO: check for hack import uri */
-        return NULL;
+    p = OSSL_PARAM_locate_const(params, P11PROV_PARAM_URI);
```
simo5, line 54: This parameter import needs to be under the conditional you moved below.

src/skeymgmt.c
```diff
@@ -22,25 +22,73 @@ static void p11prov_aes_free(void *key)
...
+           rv = p11prov_obj_find(ctx, session, slotid, uri, skey_ctx_obj, &skey_ctx);
+           if (rv != CKR_OK) {
+               P11PROV_raise(ctx, rv, "Failed to run obj_find");
+               return NULL;
+           }
+
+           return skey_ctx.obj;
     }
 
+    if (!(selection & OSSL_SKEYMGMT_SELECT_SECRET_KEY))
```
simo5, line 89: Se above, here you should conflict with any other parameter, ie this "hack" must be available as a single parameter import, because we plan to use the pkcs11_uri to indicate properties of the key (CKA_ID, CKA_LABEL, etc...) that are not normally available as OpenSSL properties in normal imports, so the only case where we should interpret the URI as a fake import to surce material from the token is in the case where the uri is the only parameter provided and the import is explicitly for a secret key.

simo5 (changes requested):
See comments, and we also need a test to show how this works.

simo5:
@mgrzeschik are you still interested in completeing this PR ?

simo5:
closing for now, feel free to reopen if you find time to work on it.

PR #577, opened 2025-05-29
Add proper version and build info to provider

### LINKED ISSUE #576: Report real provider version
**Describe the feature**
Provider should report its real version, currently it reports version of OpenSSL it is built against. The comment in code even mentions:

```
/* temporarily return the OpenSSL build version */
```

**Expected behavior**
Provider reports its own version (e.g. 1.0) instead of OpenSSL version (e.g. 3.0.8).

**Additional context**
It would be fine if we could specify custom version suffix for custom builds of pkcs11-provider. Something like `BUILD_METADATA` in OpenSSL, so we can custom build pkcs11-provider `1.0+mybuild`.

simo5:
Fixes #576

#### Description

Sets C macros for P11PROV_VERSION taken from meson's project version and P11PROV_BUILDINFO which is set from the meson option named "build_info". If this option is empty the build info string is set to
"Built with OpenSSL version <ver>" where <ver> is the version of OpenSSL used for headers at built time during compilation.

#### Checklist

- [x] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -83,8 +83,16 @@ jobs:
           NUMBER: ${{ github.event.number }}
         run: gh pr edit "$NUMBER" --remove-label "covscan"
 
+  on-covscan-not-needed:
+    if: ${{ contains(github.event.*.labels.*.name, 'covscan-not-needed')) }}
+    name: Coverity Scan on PR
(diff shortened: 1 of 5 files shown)


simo5:
I do not think we need a specific test for this, it is just informational, worst case you have an incorrect string and can be easily fixed

Issue #583, opened 2025-06-05
AES ciphers do not support setting OSSL_CIPHER_PARAM_PADDING


simo5:
**Describe the bug**
There are parameter that are allowed to be set after the cipher initialization but our code does not support any of them, including OSSL_CIPHER_PARAM_PADDING whic is useful in some situations (see PR #574 for the origin of this bug)

PR #585, opened 2025-06-05
AES: better handling of ctx params

simo5:
#### Description

Return the gettable and settable ctx params we support.

Move setting padding and CTS mode to the set_ctx_params function so that those can be set by a direct set_ctx_param() as well.

Fixes: #583

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/cipher.c
+++ b/src/cipher.c
@@ -234,47 +234,24 @@ static int set_iv(struct p11prov_cipher_ctx *ctx, const unsigned char *iv,
     return CKR_OK;
 }
 
+static int p11prov_aes_set_ctx_params(void *vctx, const OSSL_PARAM params[]);
+
 static CK_RV p11prov_cipher_prep_mech(struct p11prov_cipher_ctx *ctx,
                                       const unsigned char *iv, size_t ivlen,
                                       const OSSL_PARAM params[])
 {
-    const OSSL_PARAM *p;
     bool param_as_iv = false;
     CK_RV rv = CKR_OK;
-
-    p = OSSL_PARAM_locate_const(params, OSSL_CIPHER_PARAM_PADDING);
-    if (p) {
-        unsigned int pad;
-        int ret = OSSL_PARAM_get_uint(p, &pad);
-        if (ret != RET_OSSL_OK) {
-            rv = CKR_MECHANISM_PARAM_INVALID;
-            P11PROV_raise(ctx->provctx, rv, "Invalid padding parameter");
-            return rv;
-        }
-        ctx->pad = pad == 1;
-    }
+    int ret;
 
     switch (ctx->mech.mechanism) {
     case CKM_AES_ECB:
-        if (ctx->pad) {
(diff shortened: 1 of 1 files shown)


src/cipher.c
```diff
@@ -641,20 +612,149 @@ static int p11prov_aes_get_ctx_params(void *ctx, OSSL_PARAM params[])
...
+    case CKM_AES_ECB:
+    case CKM_AES_CBC_PAD:
+    case CKM_AES_OFB:
+    case CKM_AES_CFB128:
+    case CKM_AES_CFB1:
+    case CKM_AES_CFB8:
+    case CKM_AES_CTR:
+        return p11prov_aes_generic_settable_ctx_params;
+    case CKM_AES_CTS:
+        return p11prov_aes_cts_settable_ctx_params;
```
Jakuje, line 764: I see in the skey test only the AES-CBC. Should we add tests for the other modes we support (might need some allow/block-listing for pkcs11 backends though.

src/cipher.c
```diff
@@ -641,20 +612,149 @@ static int p11prov_aes_get_ctx_params(void *ctx, OSSL_PARAM params[])
...
+    OSSL_PARAM_END
+};
+
+static const OSSL_PARAM *p11prov_aes_gettable_ctx_params(void *vctx,
                                                          void *provctx)
 {
+    struct p11prov_cipher_ctx *ctx = (struct p11prov_cipher_ctx *)vctx;
+
+    if (!ctx) {
+        return p11prov_aes_generic_gettable_ctx_params;
```
Jakuje, line 710: Should not this be an error? If not, I think it deserves a comment why it is as I would  expect this call should be invalid.

simo5: I had to go through segfaults in openssl to learn that on loading algorithms they query both gettable_params and gettable_ctx_params, the latter with null context, but the native openssl providers do not care about context so they respond.

Note that they query a single special parameter we do not support at all and probably do not care for at this time, so we just want to avoid a failure here.

Eventually we may have to also ignore the context completely and instead register a different shim call for each different cipher (this is what openssl does via a maze of absolutely bonker macros).

For now we just respond that way, if not context we return the generic params valid for all ciphers.

src/cipher.c
simo5: We still have a long way to go with skey support, could you file a ticket for this?

src/cipher.c
Jakuje: So lets put a comment here describing why it is here for a future reader to not be confused as me :)

src/cipher.c
Jakuje: Sure. Filled #586.

Jakuje (approved):
with the comment about the odd case, ack.

Jakuje (approved):
thought I approved already. Retry.

simo5:
All good, thanks!

PR #598, opened 2025-07-21
Add login-behaviour to p11prov_rand_generate session

### LINKED ISSUE #597: Make p11prov_rand_generate with auth session
**Describe the feature**
There are HSM that require an authenticated session to call C_GenerateRandom.

**Expected behavior**
To use this HSM this call should follow the login-behaviour for the authentication.

sandevins:
#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

Closes #597


simo5 (changes requested):
p11prov_ctx_login_behavior() does not return a boolean so this change is not acceptable as is, you need to pull the value and then check if the integer returned is PUBKEY_LOGIN_ALWAYS, and only in that cases for the reqlogin boolean to true.

As coded this will result in forcing login both when it is "always" wanted and when it is "never" wanted.

And thinking about this it sounds like we maywant to move and check if login_behavior is set to PUBKEY_LOGIN_ALWAYS inside p11prov_get_seesion() and really always attempt to get a login session in that case regardless of the actual operation is trying to obtain a session.

sandevins:
Now it should be ready.

simo5 (approved):
I think ultimately we should make a helper function that collapses all checks within a commented list, but for now this fits the bill.

simo5:
Can you please squash the commits?
Also please drop the "feat:" from the commit message.

simo5:
Thanks for the PR!
Merging.

Issue #610, opened 2025-08-10
Associated private key search needs login


bukka:
**Describe the bug**

Searching for associated private key should require session with login as this is usually needed to get private keys. This is however not the case as the session does not require login: https://github.com/latchset/pkcs11-provider/blob/4b3c201d3afa7f1a8826e676c4330e3ead76c169/src/objects.c#L1455-L1457 . This results in an error because such key is not found (unless logged in already open session is used)/

**To Reproduce**
Load RSA using the util function and check the debug logs. It should result in this sort of error: `p11prov_obj_find_associated(): Error: 0x00000000; Error in C_FindObjects (count=0)`

**Expected behavior**
Session should be created with login

**Operating environment (please complete the following information):**
 - OS: Ubuntu
 - Version 24.04

**Token and application used (please complete the following information):**
 - Device: Kryotic
 - PKCS11 Driver version: later version (does not matter really)
 - Application: Test

**Additional context**

I fuond this when working on test for #536 and checking the debug logs:
```
[2025-08-10 22:15:11.177] [../src/util.c:68] p11prov_fetch_attributes(): Attribute| type:0x40000600 value:(nil), len:0
[2025-08-10 22:15:11.177] [../src/objects.c:1411] p11prov_obj_find(): Find objects: found 1 objects; Returning 0
[2025-08-10 22:15:11.177] [../src/objects.c:1430] p11prov_obj_find_associated(): Find associated object
[2025-08-10 22:15:11.177] [../src/session.c:907] p11prov_get_session(): Get session on slot 52, reqlogin=false, rw=false
[2025-08-10 22:15:11.177] [../src/session.c:916] p11prov_get_session(): single-shot request for slot 52
[2025-08-10 22:15:11.177] [../src/session.c:634] check_slot(): Checking Slot id=52, uri=(nil), mechtype=ffffffffffffffff, rw=false)
[2025-08-10 22:15:11.177] [../src/provider.c:621] p11prov_ctx_login_behavior(): login_behavior = 0
[2025-08-10 22:15:11.177] [../src/session.c:273] session_new(): Creating new P11PROV_SESSION session on pool 0x617fa05002a0
[2025-08-10 22:15:11.177] [../src/session.c:307] session_new(): Total sessions: 2
[2025-08-10 22:15:11.177] [../src/session.c:331] session_check(): Checked session 0 handle invalid
[...]

PR #611, opened 2025-08-10
Require login for session when getting private assoc key

bukka:
#### Description

This is a fix for #610 to require login if search associated object class is `CKO_PRIVATE_KEY`

I was looking into implementing a test but it seems quite tricky to create a key pair where only private key has the mechanisms attribute set but public key doesn't - I'm not sure if that's even possible using pkcs11-tool?

Alternativelly I could just do a grep from the debug log to see if there is that reported error which should be relatively simple.

Or if you are fine to get this merged without test, then it's ready as it is.

#### Checklist

#### Reviewer's checklist:


bukka:
Hmm, it's actually failing the tests as some of them don't have default PIN set and this is being checked. It might need a bit more considering as it could potentially cause issues.

simo5:
I think this has been superseeded by other PRs, so closing it for now.

PR #618, opened 2025-08-21
PKCS11 provider returns a P11PROV_OBJ *

beldmit:
...and should be able to take it back, fill with attributes and return back

Related: https://github.com/openssl/openssl/pull/28278

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -1145,6 +1145,7 @@ static CK_RV fetch_secret_key(P11PROV_CTX *ctx, P11PROV_SESSION *session,
     struct fetch_attrs attrs[BASE_KEY_ATTRS_NUM];
     int num;
     CK_RV ret;
+    CK_ATTRIBUTE *size = NULL;
 
     key->attrs = OPENSSL_zalloc(BASE_KEY_ATTRS_NUM * sizeof(CK_ATTRIBUTE));
     if (key->attrs == NULL) {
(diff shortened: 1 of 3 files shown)


src/objects.c
```diff
@@ -1166,6 +1168,17 @@ static CK_RV fetch_secret_key(P11PROV_CTX *ctx, P11PROV_SESSION *session,
     key->numattrs = 0;
     p11prov_move_alloc_attrs(attrs, num, key->attrs, &key->numattrs);
 
+    size = p11prov_obj_get_attr(key, CKA_VALUE_LEN);
+    if (!size) {
+        P11PROV_raise(key->ctx, CKR_GENERAL_ERROR, "Missing Key Size");
+        return CKR_GENERAL_ERROR;
```
simo5: Not all tokens set CKA_VALUE_LEN so this should not be a fatal error.

src/objects.c
```diff
@@ -1166,6 +1168,17 @@ static CK_RV fetch_secret_key(P11PROV_CTX *ctx, P11PROV_SESSION *session,
...
+    }
+
+    if (size->ulValueLen == sizeof(CK_ULONG)) {
```
simo5, line 1177: if CKA_VALUE_LEN is not a CK_ULONG this should be a fatal error

simo5:
We already support loading from key reference for public and private key, and loading an skey should be no different.
So if upstream code handles SKEY by reference differently from POKEY by reference that's a bug in upstream code and needs to be fixed there.

simo5 (changes requested):

beldmit:
> We already support loading from key reference for public and private key, and loading an skey should be no different. So if upstream code handles SKEY by reference differently from PKEY by reference that's a bug in upstream code and needs to be fixed there.

First, upstream is me as it's a draft PR.
Second, `try_key_ref` function (https://github.com/openssl/openssl/blob/master/crypto/store/store_result.c#L203-L274) is EVP_PKEY specific and has nothing to do with EVP_SKEY. So what I do in my PR is basically reimplementing it with limitations of EVP_SKEY.

src/objects.c
beldmit: Fixed

src/objects.c
beldmit: Fixed, thanks!

PR #624, opened 2025-09-04
tests: disable ML-DSA tesing in kryoptic on Fedora < 42

The-Mule:
#### Description

Kryoptic 1.2 is needed for ML-DSA tests and it is not available in Fedora until version 43 (FYI, kryoptic token is currently only available on Fedora).

#### Checklist

N/A

#### Reviewer's checklist:


tests/kryoptic-init.sh
```diff
@@ -51,4 +51,8 @@ export TOKENCONFIGVARS="export KRYOPTIC_CONF=$TOKDIR/kryoptic.conf"
 export TESTPORT="34000"
 
 export SUPPORT_ALLOWED_MECHANISMS=1
-export SUPPORT_ML_DSA=1
+
+# ML-DSA support is available since version 1.2 (in Fedora > 42) 
+if [ -e /etc/fedora-release ] && [ $(rpm -q --qf "%{V}" fedora-release) -gt 42 ]; then
```
simo5: I think you should test the kryoptic rpm version rather than fedora's version.
Also we need a way to let me test while developing on an arbitrary machine, so this check will get in the way.

I would rather have SUPPORT_ML_DSA settable by the CI scripts and only overridden if not already set, and then set it in the github CI env vars for the test

The-Mule: Right, that's better IMO. I did fedora version check because of kryoptic.yml GH action where we are building kryoptic from upstream. But not touching SUPPORT_ML_DSA once it is already set solves that problem. Let me do that then.

tests/kryoptic-init.sh
```diff
@@ -51,4 +51,13 @@ export TOKENCONFIGVARS="export KRYOPTIC_CONF=$TOKDIR/kryoptic.conf"
...
 
 export SUPPORT_ALLOWED_MECHANISMS=1
-export SUPPORT_ML_DSA=1
+
+# Set SUPPORT_ML_DSA only if is not set already
+if [ -z "$SUPPORT_ML_DSA" ]; then
+
+    # ML-DSA support is available since version 1.2
+    SUPPORT_ML_DSA=0
+    if rpm -q kryoptic &>/dev/null && [[ "$(rpm -q --qf '%{V}' kryoptic)" =~ 1\.[01]\. ]]; then
```
simo5: Sorry I was not clear, but I do not want this check in this file, as it will cause unnecessary errors on non RPM systems.

This file should only support the ability to override the  SUPPORT_ML_DSA from an external job and default to 1 otherwise.

The CI jobs can then test kryoptic's version based on whatever packaging system the build is running on and set the override there.

The-Mule: no worries, got it, I hate the check as well and I'll gladly drop it, so we want to default to SUPPORT_ML_DSA=1, that would require a temporary update in openssl CI as well (until F43 gets out). So here's what I am gonna do:

1. for kryoptic I'll use the default (1) unless it's set already.
2. for the other tokens (including kryoptic-nss at the moment) I'll forcibly disable it by setting 0 (regarless of whether it's set of not).

This won't require any change in out GH workflows but at the same it  allows me to set it in openssl CI temporarily via Dima's PR.

The-Mule: The current version should reflect this.

simo5 (approved):
LGTM

The-Mule:
Added forgotten export to setup.sh (`export SUPPORT_ED448="${SUPPORT_ED448}"`) and replaced `test -z` with `if [ -z` in kryoptic-init.sh (to avoid setup fail when SUPPORT_ML_DSA is already defined). Verified carefully that the mldsa test runs with kryoptic and that it can be disabled by setting `SUPPORT_ML_DSA=0`.

Now it's finally good to go.

Issue #622, opened 2025-09-02
SKEY derivation for EVP_PKEY and KDFs, PKCS#11 counterpart


beldmit:
**Describe the feature**
https://github.com/openssl/openssl/pull/28369 is merged to OpenSSL. It allows SKEY derivation for KDFs and key exchange, we need to implement the PKCS#11 counterpart and extend the openssl tests

**Expected behavior**
It should be possible to derive EVP_SKEY objects having a private key on a token and external peer, and use the derived key for ciphers and macs.

It should be possible to derive EVP_SKEY objects using token KDF implementations, and use the derived key for ciphers and macs.

PR #631, opened 2025-09-12
Implement SKEY derive for HKDF

simo5:
#### Description

Adapt the HKDF derive function to also provide the new SKEY interfaces allowing to use and return only key objects from the token.

This PR also adds skey management for Generic Secrets as well as the ability to export Generic Secrets if the token allows export.

Add test program to check HKDF works as expected.

This PR has been partially aided by Gemini to create the test program and some of the infrastructure needed by the tests code.

Implements the KDF part of #622 

#### Checklist

- [ ] ~Documentation updated~

#### Reviewer's checklist:

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -162,7 +162,7 @@ tests = {
   'genkey': {'suites': all_suites},
   'pkey': {'suites': all_suites},
   'session': {'suites': all_suites},
-  'skey': {'suites': all_suites},
+  'skey': {'suites': ['softokn', 'kryoptic', 'kryoptic.nss']},
   'rand': {'suites': all_suites},
   'readkeys': {'suites': all_suites},
   'tls': {'suites': all_suites, 'is_parallel': false, 'timeout': 60},


src/kdf.c
```diff
@@ -189,46 +212,66 @@ static int inner_derive_key(P11PROV_CTX *ctx, P11PROV_OBJ *key,
...
+        params->bExtract = CK_TRUE;
+    } else {
+        params->bExtract = CK_FALSE;
+    }
+    if (hkdfctx->mode == EVP_KDF_HKDF_MODE_EXTRACT_AND_EXPAND
+        || hkdfctx->mode == EVP_KDF_HKDF_MODE_EXPAND_ONLY) {
+        params->bExpand = CK_TRUE;
+    } else {
+        params->bExpand = CK_FALSE;
+    }
```
Jakuje, line 232: wondering if the switch/case would not make it more readable. But not sure:
```suggestion
    switch (hkdfctx->mode) {
    case EVP_KDF_HKDF_MODE_EXTRACT_AND_EXPAND:
        params->bExtract = CK_TRUE;
        params->bExpand = CK_TRUE;
        break;
    case EVP_KDF_HKDF_MODE_EXTRACT_ONLY:
        params->bExtract = CK_TRUE;
        params->bExpand = CK_FALSE;
        break;
    case EVP_KDF_HKDF_MODE_EXPAND_ONLY:
        params->bExtract = CK_FALSE;
        params->bExpand = CK_TRUE;
        break;
    }
```

src/skeymgmt.c
```diff
@@ -302,4 +302,259 @@ const OSSL_DISPATCH p11prov_aes_skeymgmt_functions[] = {
...
+DISPATCH_SKEYMGMT_FN(generic_secret, gen_settable_params);
+DISPATCH_SKEYMGMT_FN(generic_secret, imp_settable_params);
+
+static void p11prov_generic_secret_free(void *key)
+{
+    P11PROV_debug("generic_secret free %p", key);
+    p11prov_obj_free((P11PROV_OBJ *)key);
+}
+
+static void *p11prov_generic_secret_import(void *provctx, int selection,
```
beldmit, line 319: Can't we somehow combine this function with the similar AES one?

src/skeymgmt.c
```diff
@@ -302,4 +302,259 @@ const OSSL_DISPATCH p11prov_aes_skeymgmt_functions[] = {
...
+{
+    P11PROV_OBJ *key = (P11PROV_OBJ *)keydata;
+
+    P11PROV_raise(p11prov_obj_get_prov_ctx(key), CKR_KEY_FUNCTION_NOT_PERMITTED,
+                  "Not exportable");
+
+    return RET_OSSL_ERR;
+}
+
+static void *p11prov_generic_secret_generate(void *provctx,
```
beldmit, line 491: And this

beldmit (approved):
Probably you should separate the common code for GENERIC/AES SKEYMGMTs to reduce the overall volume of the files. 

I'm also not sure about the default key purposes so leaving it up to you.

All this stuff, however, can be aligned later. LGTM.

src/skeymgmt.c
```diff
@@ -302,4 +302,367 @@ const OSSL_DISPATCH p11prov_aes_skeymgmt_functions[] = {
...
+        if (ret != RET_OSSL_OK) {
+            P11PROV_raise(ctx, CKR_KEY_INDIGESTIBLE, "Invalid data");
+            return NULL;
+        }
+        return p11prov_obj_from_typed_reference(reference, reference_sz,
+                                                CKK_GENERIC_SECRET);
+    }
+
+    /* Not a digestible secret key */
+    P11PROV_raise(ctx, CKR_KEY_INDIGESTIBLE, "Raw Bytes param unavailable");
```
Jakuje:
```suggestion
    P11PROV_raise(ctx, CKR_KEY_INDIGESTIBLE, "Raw Bytes nor reference param unavailable");
```

src/skeymgmt.c
```diff
@@ -302,4 +302,367 @@ const OSSL_DISPATCH p11prov_aes_skeymgmt_functions[] = {
...
+        if (param_cb(params, cbarg)) {
+            ret = RET_OSSL_OK;
+        }
+
+        /* Note: we MUST cache the attribute here, because the callback
+         * OpenSSL use expect the value to valid for the life of the key
+         * object and will just store the provided pointer. Therefore we need
+         * to keep value_attr.pValue alive as it will be used after this
+         * function returns. This mechanism works also as cache to avoid
+         * refecthing from the token multiple times. As multiple import/export
```
Jakuje:
```suggestion
         * re-fetching from the token multiple times. As multiple import/export
```

src/skeymgmt.c
```diff
@@ -302,4 +302,367 @@ const OSSL_DISPATCH p11prov_aes_skeymgmt_functions[] = {
...
+         * cycles may happen when a mize of legacy and SKEY functions are used
```
Jakuje: ?
```suggestion
         * cycles may happen when a mix of legacy and SKEY functions are used
```

src/skeymgmt.c
simo5: We could, but i prefer to keep them distinct until we have some more experinece on what are the common behaviors we want and what we don't want.
For example it is intentional that Generic Secrets can be exported while AES keys cannot (yet) right now.

src/kdf.c
simo5: This code just moved over form same allocations on the stack to this helper function, I do not want to change what it looks like in this patch set.

PR #641, opened 2025-09-25
Improve session handling

### LINKED ISSUE #636: Review session usage in EVP_SKEY interfaces
Some interfaces now create ephemeral session keys that are supposed to outlive the original operation.
For these operation we need to ensure that the key handle is stored on the login session so that it is not inadvertently recycled away (and the session key destroyed) while the key is still in use.

Ideally a test case is added that ensure key availability after ephemeral EVP_SKEYs are generated (Derive/Exchange operations).

simo5:
#### Description

This patchset comes out of a review oh how sessions are used after the EVP_SKEY facilities introduction.

One of the things I notices was that we had overly complicated code around handling two items on the session pool, login_session and open_sessions. Manipulating these attributes required, potentially costly, locking of pools every time a session is taken.

Additionally I noticed that in repeat operations session could be passed around which is not optimal on tokens that cache or otherwise internally tie ephemeral keys to sessions.

By adding a simple reference counter that prevents session from being closed and saving a pointer to those sessions on objects that have been created, it was possible to remove a bunch of locking and simplify other calls.

All this also ensures that ephemeral keys are not inadvertently discarded by closing the session they were created on.

Fixes #636 

#### Checklist

- [x] Test suite already covers the code
- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/address-sanitizer.yml
+++ b/.github/workflows/address-sanitizer.yml
@@ -18,7 +18,7 @@ jobs:
         # , debian]
         include:
           - name: fedora
-            container: fedora:latest
+            container: quay.io/fedora/fedora:latest
         #  - name: debian
         #    container: debian:sid
     container: ${{ matrix.container }}
--- a/src/asymmetric_cipher.c
+++ b/src/asymmetric_cipher.c
@@ -151,7 +151,6 @@ static int p11prov_rsaenc_encrypt(void *ctx, unsigned char *out, size_t *outlen,
     CK_MECHANISM mechanism;
     P11PROV_SESSION *session;
     CK_SESSION_HANDLE sess;
-    CK_SLOT_ID slotid;
     CK_OBJECT_HANDLE handle;
     CK_ULONG out_size = *outlen;
     int result = RET_OSSL_ERR;
@@ -169,12 +168,6 @@ static int p11prov_rsaenc_encrypt(void *ctx, unsigned char *out, size_t *outlen,
         return RET_OSSL_OK;
     }
 
-    slotid = p11prov_obj_get_slotid(encctx->key);
-    if (slotid == CK_UNAVAILABLE_INFORMATION) {
-        P11PROV_raise(encctx->provctx, CKR_SLOT_ID_INVALID,
-                      "Provided key has invalid slot");
-        return RET_OSSL_ERR;
-    }
     handle = p11prov_obj_get_handle(encctx->key);
     if (handle == CK_INVALID_HANDLE) {
         P11PROV_raise(encctx->provctx, CKR_KEY_HANDLE_INVALID,
@@ -187,12 +180,10 @@ static int p11prov_rsaenc_encrypt(void *ctx, unsigned char *out, size_t *outlen,
         return RET_OSSL_ERR;
     }
 
-    ret = p11prov_get_session(encctx->provctx, &slotid, NULL, NULL,
-                              mechanism.mechanism, NULL, NULL, false, false,
-                              &session);
+    ret = p11prov_try_session_ref(encctx->key, mechanism.mechanism, false,
+                                  false, &session);
     if (ret != CKR_OK) {
-        P11PROV_raise(encctx->provctx, ret,
-                      "Failed to open session on slot %lu", slotid);
+        P11PROV_raise(encctx->provctx, ret, "Failed to acquire session");
         return RET_OSSL_ERR;
     }
     sess = p11prov_session_handle(session);
@@ -301,7 +292,6 @@ static int p11prov_rsaenc_decrypt(void *ctx, unsigned char *out, size_t *outlen,
     CK_MECHANISM mechanism;
     P11PROV_SESSION *session = NULL;
     CK_SESSION_HANDLE sess;
-    CK_SLOT_ID slotid;
     CK_OBJECT_HANDLE handle;
     CK_ULONG key_size = CK_UNAVAILABLE_INFORMATION;
     unsigned char *tmpbuf = NULL;
@@ -350,13 +340,6 @@ static int p11prov_rsaenc_decrypt(void *ctx, unsigned char *out, size_t *outlen,
         out_size = key_size;
     }
 
-    slotid = p11prov_obj_get_slotid(encctx->key);
-    if (slotid == CK_UNAVAILABLE_INFORMATION) {
-        P11PROV_raise(encctx->provctx, CKR_SLOT_ID_INVALID,
-                      "Provided key has invalid slot");
-        result = RET_OSSL_ERR;
-        goto done;
-    }
     handle = p11prov_obj_get_handle(encctx->key);
     if (handle == CK_INVALID_HANDLE) {
         P11PROV_raise(encctx->provctx, CKR_KEY_HANDLE_INVALID,
@@ -371,12 +354,10 @@ static int p11prov_rsaenc_decrypt(void *ctx, unsigned char *out, size_t *outlen,
         goto done;
     }
 
-    ret = p11prov_get_session(encctx->provctx, &slotid, NULL, NULL,
-                              mechanism.mechanism, NULL, NULL, true, false,
-                              &session);
+    ret = p11prov_try_session_ref(encctx->key, mechanism.mechanism, true, false,
+                                  &session);
     if (ret != CKR_OK) {
-        P11PROV_raise(encctx->provctx, ret,
-                      "Failed to open session on slot %lu", slotid);
+        P11PROV_raise(encctx->provctx, ret, "Failed to acquire session");
         result = RET_OSSL_ERR;
         goto done;
     }
--- a/src/cipher.c
+++ b/src/cipher.c
@@ -333,15 +333,7 @@ static CK_RV p11prov_cipher_op_init(void *ctx, void *keydata, CK_FLAGS op,
 
 static CK_RV p11prov_cipher_session_init(struct p11prov_cipher_ctx *cctx)
 {
-    CK_SLOT_ID slotid;
     CK_RV rv;
-
-    slotid = p11prov_obj_get_slotid(cctx->key);
-    if (slotid == CK_UNAVAILABLE_INFORMATION) {
-        P11PROV_raise(cctx->provctx, CKR_SLOT_ID_INVALID,
-                      "Provided key has invalid slot");
-        return CKR_SLOT_ID_INVALID;
-    }
 
     if (cctx->tlsver != 0 && cctx->mech.mechanism == CKM_AES_CBC_PAD) {
         /* In the special TLS mode we handle de-padding and mac extraction
@@ -349,9 +341,8 @@ static CK_RV p11prov_cipher_session_init(struct p11prov_cipher_ctx *cctx)
         cctx->mech.mechanism = CKM_AES_CBC;
     }
 
-    rv = p11prov_get_session(cctx->provctx, &slotid, NULL, NULL,
-                             cctx->mech.mechanism, NULL, NULL, true, false,
-                             &cctx->session);
+    rv = p11prov_try_session_ref(cctx->key, cctx->mech.mechanism, true, false,
+                                 &cctx->session);
     if (rv != CKR_OK) {
         return rv;
     }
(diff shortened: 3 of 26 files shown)


src/asymmetric_cipher.c
```diff
@@ -187,12 +180,10 @@ static int p11prov_rsaenc_encrypt(void *ctx, unsigned char *out, size_t *outlen,
...
 
-    ret = p11prov_get_session(encctx->provctx, &slotid, NULL, NULL,
-                              mechanism.mechanism, NULL, NULL, false, false,
-                              &session);
+    ret = p11prov_try_session_ref(encctx->key, mechanism.mechanism, false,
+                                  false, &session);
     if (ret != CKR_OK) {
-        P11PROV_raise(encctx->provctx, ret,
-                      "Failed to open session on slot %lu", slotid);
+        P11PROV_raise(encctx->provctx, ret, "Failed to open session");
```
Jakuje: given that this does not open the session always, I would adjust the error message:
```suggestion
        P11PROV_raise(encctx->provctx, ret, "Failed to acquire session");
```

src/asymmetric_cipher.c
```diff
@@ -371,12 +354,10 @@ static int p11prov_rsaenc_decrypt(void *ctx, unsigned char *out, size_t *outlen,
...
 
-    ret = p11prov_get_session(encctx->provctx, &slotid, NULL, NULL,
-                              mechanism.mechanism, NULL, NULL, true, false,
-                              &session);
+    ret = p11prov_try_session_ref(encctx->key, mechanism.mechanism, true, false,
+                                  &session);
     if (ret != CKR_OK) {
-        P11PROV_raise(encctx->provctx, ret,
-                      "Failed to open session on slot %lu", slotid);
+        P11PROV_raise(encctx->provctx, ret, "Failed to open session");
```
Jakuje: same here and likely in  other cases following this patttern.

src/objects.c
```diff
@@ -798,6 +804,22 @@ void *p11prov_obj_from_typed_reference(const void *reference,
     return key;
 }
 
+P11PROV_SESSION *p11prov_obj_get_session_ref(P11PROV_OBJ *obj)
+{
+    return obj->ref_session;
+}
+
+void p11prov_obj_ref_session(P11PROV_OBJ *obj, P11PROV_SESSION *session)
```
Jakuje: I think the function names like this will make it very confusing (especially when they do not have any attacked comments or doc text). On the first sight, I have no idea if I would like to use the first or the second and what is their actual difference (before diving into the source code). Should I do `ref_session` or `get_session_ref`?

src/objects.c
```diff
@@ -1768,46 +1781,54 @@ P11PROV_OBJ *p11prov_create_secret_key(P11PROV_CTX *provctx,
...
+                         P11PROV_SESSION **_session, CK_OBJECT_HANDLE *dkey)
 {
+    P11PROV_CTX *ctx = p11prov_obj_get_prov_ctx(key);
+    CK_OBJECT_HANDLE handle = CK_INVALID_HANDLE;
+    P11PROV_SESSION *session = *_session;
     bool first_pass = true;
-    P11PROV_SESSION *s = *session;
     CK_RV ret;
 
+    /* do this first as it may cause a refresh o the object that will
```
Jakuje:
```suggestion
    /* do this first as it may cause a refresh of the object that will
```

src/session.c
```diff
@@ -1052,6 +1076,76 @@ CK_RV p11prov_get_session(P11PROV_CTX *provctx, CK_SLOT_ID *slotid,
     return ret;
 }
 
+CK_RV p11prov_try_session_ref(P11PROV_OBJ *obj, CK_MECHANISM_TYPE mechtype,
+                              bool reqlogin, bool rw,
+                              P11PROV_SESSION **_session)
+{
+    P11PROV_CTX *ctx = p11prov_obj_get_prov_ctx(obj);
+    CK_SLOT_ID slotid = p11prov_obj_get_slotid(obj);
```
Jakuje, line 1119: check for UNAVAILABLE_INFORMATION here?

simo5: It was unclear to me that it was invalid to not have a slotid, some objects don't, until a session and a handle is found.

src/objects.c
simo5: added doc string and changed name to p11prov_obj_set_session_ref

simo5:
@Jakuje this should be ready for review now that I found all the various failures tlsfuzzer tests surfaced

.github/workflows/address-sanitizer.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         # , debian]
         include:
           - name: fedora
-            container: fedora:latest
+            container: quay.io/fedora/fedora:latest
         #  - name: debian
         #    container: debian:sid
```
Jakuje, line 23: a bit off-topic, but are the debian tests something to re-check if it is working already? Or is it going to work with trixie as you changed other pipelines to use it?

src/cipher.c
```diff
@@ -209,8 +214,41 @@ static void p11prov_cipher_freectx(void *ctx)
...
+                                          CK_INVALID_HANDLE);
+            } else {
+                ret = p11prov_DecryptInit(cctx->provctx, sess, NULL,
+                                          CK_INVALID_HANDLE);
+            }
+            if (ret != CKR_OK) {
+                /* NSS softokn has a broken interface and is incapable of
+                 * dropping operations on sessions returning a generic
+                 * CKR_MECHANISM_PARAM_INVALID when the mechanism is set to
+                 * NULL. Attempt to foce cancellation via C_SessionCancel. */
```
Jakuje:
```suggestion
                 * NULL. Attempt to force cancellation via C_SessionCancel. */
```

src/encoder.gen.c
```diff
@@ -34,8 +37,10 @@ void P11PROV_RSA_PUBKEY_free(P11PROV_RSA_PUBKEY *a)
 }
 
 static const ASN1_TEMPLATE P11PROV_RSA_PUBKEY_seq_tt[] = {
+
     { (0), (0), __builtin_offsetof(P11PROV_RSA_PUBKEY, n), "n",
       (ASN1_INTEGER_it) },
+
```
Jakuje, line 43: this is some gemini artifact or the new enforced format after update of clang-format?

src/objects.c
```diff
@@ -1768,46 +1784,54 @@ P11PROV_OBJ *p11prov_create_secret_key(P11PROV_CTX *provctx,
...
     switch (ret) {
     case CKR_OK:
-        *session = s;
+        *_session = session;
         return CKR_OK;
     case CKR_SESSION_CLOSED:
     case CKR_SESSION_HANDLE_INVALID:
         if (first_pass) {
             first_pass = false;
             /* TODO: Explicitly mark handle invalid */
```
Jakuje: should this set the session state to broken/invalid as said in the todo? Or is it happening somewhere inside?

src/objects.c
```diff
@@ -4106,6 +4132,11 @@ static CK_RV p11prov_store_rsa_public_key(P11PROV_OBJ *key)
     rv = CKR_OK;
 
 done:
+    if (rv == CKR_OK) {
+        /* we just created an ephemeral key on this session, ensure the
+        * session is not closed until the key goes away */
```
Jakuje:
```suggestion
        /* we just created an ephemeral key on this session, ensure the
         * session is not closed until the key goes away */
```

src/objects.c
```diff
@@ -4174,6 +4207,11 @@ static CK_RV p11prov_store_ec_public_key(P11PROV_OBJ *key)
...
```
Jakuje:
```suggestion
         * session is not closed until the key goes away */
```
same in all the following comments

src/objects.c
```diff
@@ -4090,7 +4114,9 @@ static CK_RV p11prov_store_rsa_public_key(P11PROV_OBJ *key)
         goto done;
     }
 
-    rv = p11prov_take_login_session(key->ctx, slot, &session);
+    rv = p11prov_get_session(key->ctx, &slot, NULL, key->refresh_uri,
+                             CK_UNAVAILABLE_INFORMATION, NULL, NULL, true, true,
+                             &session);
```
Jakuje: Why the RSA kase requires login, while the ec case below does not (and mldsa does need again)? Is there some logic behind this inconsistency?

src/session.c
```diff
@@ -1052,6 +1111,81 @@ CK_RV p11prov_get_session(P11PROV_CTX *provctx, CK_SLOT_ID *slotid,
...
+    if (mechtype != CK_UNAVAILABLE_INFORMATION) {
+        ret = p11prov_check_mechanism(ctx, slotid, mechtype);
+        if (ret != CKR_OK) {
+            return ret;
+        }
+    }
+
+    if (session) {
+        P11PROV_debug("Try referenced session: %p, handle: %lu", session,
+                      p11prov_session_handle(session));
```
Jakuje: could we use just `session->session` instead of calling into `p11prov_session_handle` which basically just retunrs the reference?
```suggestion
                      session->session);
```
likely not significant performance, but sounds needless.

src/sig/internal.h
```diff
@@ -29,6 +29,7 @@ struct p11prov_sig_ctx {
 
     CK_FLAGS operation;
     P11PROV_SESSION *session;
+    enum { SESS_UNUSED = 0, SESS_INITIALIZED, SESS_FINALIZED } session_state;
```
Jakuje, line 32: should we make this named enum instead of defining it twice in different headers inline?

src/sig/signature.c
```diff
@@ -186,12 +191,45 @@ void p11prov_sig_freectx(void *ctx)
...
+                                       CK_INVALID_HANDLE);
+            } else {
+                ret = p11prov_VerifyInit(sigctx->provctx, sess, NULL,
+                                         CK_INVALID_HANDLE);
+            }
+            if (ret != CKR_OK) {
+                /* NSS softokn has a broken interface and is incapable of
+                 * dropping operations on sessions returning a generic
+                 * CKR_MECHANISM_PARAM_INVALID when the mechanism is set to
+                 * NULL. Attempt to foce cancellation via C_SessionCancel. */
```
Jakuje:
```suggestion
                 * NULL. Attempt to force cancellation via C_SessionCancel. */
```

.github/workflows/address-sanitizer.yml
simo5: maybe in another PR

src/encoder.gen.c
simo5: check-style-fix did it :man_shrugging:

simo5: it is an auto-generated file at any rate, so not going to touch it

src/objects.c
simo5: Not technically needed as in p11prov_get_session() we always check the session is valid by querying the session info, so it gets discarded as not good there.

However it could be a mini-optimization to reset the handle, let me do that.

simo5: Actually looking at this it is not clear to me why we'd go through this retry loop only in this place. If the session was closed we'd get a failure to source the session in the first place. And any other error is unlikely to be resolved by just grabbing a new session.

Would you mind if Iturn this into a regular failure and avoid the retry ?

src/objects.c
simo5: no I think I made some mistake here, which generally does not matter because we are logged in by the time we call any of these.

However I think we only really need a rw session, but not login session given we are storing these keys as ephemeral in the session.

src/sig/internal.h
simo5: It would have to be public (declaration in header, type-defing, etc...) for no much gain, I originally did it that way and in the end reverted to this single liner which was more compact.

src/session.c
simo5: The function transparently handles the case where session is NULL, otherwise I would have to guard for it in the debug statement, easier and more compact this way.

Jakuje: you are in the `if (session)` block here so session can not be null here.

src/objects.c
Jakuje: I do not really have a strong opinion. I would go with whatever sounds safer.

src/session.c
simo5: doh!

simo5:
@Jakuje thanks for the review!

PR #643, opened 2025-10-13
Add ML-KEM support

### LINKED ISSUE #558: Add support for ML-KEM and ML-DSA
**Describe the feature**
We should provide access to tokens that support ML-KEM and ML-DSA at least,
this will require adding support for the PKCS#11 3.2 API.

**Expected behavior**
New PQC algoriuthms are available

**Additional context**
See PKCS#11 3.2 draft documents.

simo5:
#### Description

Implement all the plumbing needed to expose the ML-KEM Key Encapsulation mechanisms

Resolves #558

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:


src/encoder.c
```diff
@@ -1375,3 +1375,286 @@ const OSSL_DISPATCH p11prov_mldsa_encoder_text_functions[] = {
...
+    int ret;
+
+    P11PROV_debug("mlkem Text Encoder");
+
+    type = p11prov_obj_get_key_type(key);
+    if (type != CKK_ML_KEM) {
+        P11PROV_raise(ctx->provctx, CKR_GENERAL_ERROR, "Invalid Key Type");
+        return RET_OSSL_ERR;
+    }
+    class = p11prov_obj_get_class(key);
```
Jakuje: could we move getting class close to its usage, likely on line 1610? It looks misplaced here.

src/kem.c
```diff
@@ -0,0 +1,393 @@
...
+    return p11prov_mlkem_set_ctx_params(vctx, params);
+}
+
+static int p11prov_mlkem_encapsulate_init(void *vctx, void *vkey,
+                                          const OSSL_PARAM params[])
+{
+    P11PROV_OBJ *key = vkey;
+
+    if (p11prov_obj_get_class(key) != CKO_PUBLIC_KEY) {
+        ERR_raise(ERR_LIB_PROV, PROV_R_MISSING_KEY);
```
Jakuje: Would a `PROV_R_NOT_A_PUBLIC_KEY` be a better error here? Same in the function below, `PROV_R_NOT_A_PRIVATE_KEY`?

src/kem.c
```diff
@@ -0,0 +1,393 @@
...
+}
+
+static CK_RV p11prov_mlkem_get_lengths(P11PROV_OBJ *key, size_t *ct_len,
+                                       size_t *ss_len)
+{
+    CK_ULONG key_size;
+
+    key_size = p11prov_obj_get_key_size(key);
+
+    *ss_len = 32;
```
Jakuje: should we set this in the end of the function, rather than at the beginning to make sure we check the size matches?

src/kem.c
```diff
@@ -0,0 +1,393 @@
...
+    CK_RV rv;
+    CK_SLOT_ID slot_id;
+
+    if (ctx == NULL || ctx->key == NULL) {
+        ERR_raise(ERR_LIB_PROV, PROV_R_MISSING_KEY);
+        return RET_OSSL_ERR;
+    }
+
+    rv =
+        p11prov_mlkem_get_lengths(ctx->key, &expected_ct_len, &expected_ss_len);
```
Jakuje: nit: I hate this format. Could we cut one or two characters somewhere to keep it on one line, please? :)

src/keymgmt.c
```diff
@@ -2616,6 +2621,475 @@ const OSSL_DISPATCH p11prov_mldsa87_keymgmt_functions[] = {
...
+    if (!key) {
+        return RET_OSSL_ERR;
+    }
+
+    if (selection & OSSL_KEYMGMT_SELECT_PRIVATE_KEY) {
+        class = CKO_PRIVATE_KEY;
+    } else if (selection & OSSL_KEYMGMT_SELECT_PUBLIC_KEY) {
+        class = CKO_PUBLIC_KEY;
+    } else if (selection & OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS) {
+        class = CKO_DOMAIN_PARAMETERS;
```
Jakuje: I do not think there is a support to import domain parameters object for ML-KEM, is it?

src/keymgmt.c
```diff
@@ -2616,6 +2621,475 @@ const OSSL_DISPATCH p11prov_mldsa87_keymgmt_functions[] = {
...
+        case CKP_ML_KEM_768:
+            secbits = 192;
+            break;
+        case CKP_ML_KEM_1024:
+            secbits = 256;
+            break;
+        }
+        if (secbits == 0) {
+            return RET_OSSL_ERR;
+        }
```
Jakuje, line 2955: would this be more readable to keep the return in the switch?
```suggestion
        switch (param_set) {
        case CKP_ML_KEM_512:
            secbits = 128;
            break;
        case CKP_ML_KEM_768:
            secbits = 192;
            break;
        case CKP_ML_KEM_1024:
            secbits = 256;
            break;
        default:
            return RET_OSSL_ERR;
        }
```

src/objects.c
```diff
@@ -3203,6 +3320,23 @@ int p11prov_obj_key_cmp(P11PROV_OBJ *key1, P11PROV_OBJ *key2, CK_KEY_TYPE type,
         }
         break;
 
+    case CKK_ML_KEM:
+        /* The parameter set maps to the specific key bit length that was
+         * compared already. If the bit size matches, the param set matches. */
+        if (cmp_type & OBJ_CMP_KEY_PRIVATE) {
+            /* unfortunately we can't really read private attributes
+             * and there is no comparison function int he PKCS11 API.
```
Jakuje:
```suggestion
             * and there is no comparison function in the PKCS11 API.
```

src/objects.c
```diff
@@ -3840,6 +3974,99 @@ static CK_RV prep_mldsa_find(P11PROV_CTX *ctx, const OSSL_PARAM params[],
     return CKR_OK;
 }
 
+static CK_RV prep_mlkem_find(P11PROV_CTX *ctx, const OSSL_PARAM params[],
+                             struct pool_find_ctx *findctx)
+{
+    data_buffer digest_data[4];
```
Jakuje, line 3980: We use just 3 members here. Is it intentional for alignment or overlook?

tests/kryoptic.nss-init.sh
```diff
@@ -31,3 +31,5 @@ export TESTPORT="29000"
 export SUPPORT_ALLOWED_MECHANISMS=0
 # not yet
 export SUPPORT_ML_DSA=0
+# not yet
+export SUPPORT_ML_KEM=0
```
Jakuje, line 35: Why the NSS DB should not support these?

src/encoder.c
simo5: ok

src/kem.c
simo5: sure

src/kem.c
simo5: not sure it make an important difference, but I'll do.

src/kem.c
simo5: lemme see

src/keymgmt.c
simo5: There isn't, I'll drop it

src/keymgmt.c
simo5: yeah makes sense

src/objects.c
simo5: It is a convention throughout the code.
I'd like to avoid touching this here because I am already working on a follow-up PR so reduce code duplication and refactor objects.c

tests/kryoptic.nss-init.sh
simo5: because kryoptic was broken, we just merged https://github.com/latchset/kryoptic/pull/350 to unbreak it but until we get it into fedora it won't work.

Jakuje: Don't we run the kryoptic build from main in the CI here? https://github.com/latchset/pkcs11-provider/blob/main/.github/workflows/kryoptic.yml#L64

src/kem.c
```diff
@@ -0,0 +1,392 @@
...
```
Jakuje: I pointed the one use, but this has the same oddness so once changed in the one case, it would be good to change it too.

src/keymgmt.c
```diff
@@ -2616,6 +2621,471 @@ const OSSL_DISPATCH p11prov_mldsa87_keymgmt_functions[] = {
...
+    if (selection == OSSL_KEYMGMT_SELECT_PUBLIC_KEY) {
+        return p11prov_mlkem_exp_key_types;
+    }
+    return NULL;
+}
+
+/* See FIPS-203, 4. Parameter Sets */
+#define ML_KEM_512_MAX_SIZE 768
+#define ML_KEM_768_MAX_SIZE 1088
+#define ML_KEM_1024_MAX_SIZE 1568
```
Jakuje: these match the `ML_KEM_*_CIPHERTEXT_BYTES` defines from the header file. Do we need to define them separately?

src/objects.c
```diff
@@ -4275,6 +4512,79 @@ static CK_RV p11prov_store_mldsa_public_key(P11PROV_OBJ *key)
...
+    }
+
+    key->slotid = slot;
+
+    rv = CKR_OK;
+
+done:
+    if (rv == CKR_OK) {
+        /* we just created an ephemeral key on this session, ensure the
+        * session is not closed until the key goes away */
```
Jakuje, line 4581:
```suggestion
         * session is not closed until the key goes away */
```

src/objects.c
```diff
@@ -4701,6 +5014,97 @@ static CK_RV p11prov_store_mldsa_private_key(P11PROV_OBJ *key,
...
+            goto done;
+        }
+        key->numattrs++;
+    }
+    rv = CKR_OK;
+
+done:
+    if (rv == CKR_OK) {
+        /* we just created an ephemeral key on this session, ensure the
+        * session is not closed until the key goes away */
```
Jakuje, line 5101:
```suggestion
         * session is not closed until the key goes away */
```

Jakuje (dismissed):
few more random stuff, but mostly nits. And I think we should enable this on kryoptic-nss as we build it in CI.
Some of the comments pointed out before (CKO_DOMAIN_PARAMETERS, PROV_R_MISSING_KEY) are also applicable to ML-DSA or other key types, but this could be squashed later (or never).

src/kem.c
simo5: missed this

src/keymgmt.c
simo5: To be honest I am not sure these values are actually correct now that I think of it, but they need to be moved in the header, yes, I'll modify this as needed

src/objects.c
simo5: I am changing all this in a followup PR, so it will go away later anyway

PR #649, opened 2025-10-17
Refactor object management and split into multiple files

### LINKED ISSUE #540: Objects pool in inconsistent state error for duplicated objects
**Describe the bug**

There is an with trying to remove objects from pool that are not present in it. This happens for duplicated object and results in errors `Objects pool in inconsistent state error` even though it is not necessary to trigger those errors.

**To Reproduce**
1. Setup nginx with pkcs11-provider and server key using pkcs11 key and client cert using normal file cert
2. Run nginx with OPENSSL_CONF loading the provider using SoftHSMv2
3. Send bunch of requests
4. Check logs and search for "Objects pool in inconsistent state error"

**Expected behavior**
No error

 - OS: Ubuntu
 - Version 20.04 

 - Device: SoftHSMv2
 - PKCS11 Driver version: latest develop
 - Application nginx
 - Version 1.27.4

simo5:
#### Description

Being fed up with the humongous size objects.c had grown to, I started refactoring the code to split it in smaller files.

Along the way I also added isolated commits to simplify or de-duplicate functions.
No behavior change is expected, this is just pure refactoring.

Also fixes #540

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/obj/keymgmt.c
+++ b/src/obj/keymgmt.c
@@ -543,8 +543,6 @@ static int cmp_public_key_values(P11PROV_OBJ *pub_key1, P11PROV_OBJ *pub_key2)
         ret = cmp_attr(pub_key1, pub_key2, CKA_P11PROV_PUB_KEY);
         break;
     case CKK_ML_DSA:
-        ret = cmp_attr(pub_key1, pub_key2, CKA_VALUE);
-        break;
     case CKK_ML_KEM:
         ret = cmp_attr(pub_key1, pub_key2, CKA_VALUE);
         break;
@@ -749,18 +747,8 @@ int p11prov_obj_key_cmp(P11PROV_OBJ *key1, P11PROV_OBJ *key2, CK_KEY_TYPE type,
 
     switch (key1->data.key.type) {
     case CKK_RSA:
-        if (cmp_type & OBJ_CMP_KEY_PRIVATE) {
-            /* unfortunately we can't really read private attributes
-             * and there is no comparison function in the PKCS11 API.
-             * Generally you do not have 2 identical keys stored in to two
-             * separate objects so the initial shortcircuit that matches if
-             * slotid/handle are identical will often cover this. When that
-             * fails we have no option but to fail for now. */
-            P11PROV_debug("We can't really match private keys");
-            /* OTOH if modulus and exponent match either this is a broken key
-             * or the private key must also match */
-            cmp_type = OBJ_CMP_KEY_PUBLIC;
-        }
+    case CKK_ML_DSA:
+    case CKK_ML_KEM:
         break;
 
     case CKK_EC:
@@ -819,19 +807,8 @@ int p11prov_obj_key_cmp(P11PROV_OBJ *key1, P11PROV_OBJ *key2, CK_KEY_TYPE type,
                 return ret;
             }
         }
-        if (cmp_type & OBJ_CMP_KEY_PRIVATE) {
-            /* unfortunately we can't really read private attributes
-             * and there is no comparison function in the PKCS11 API.
-             * Generally you do not have 2 identical keys stored in to two
-             * separate objects so the initial shortcircuit that matches if
-             * slotid/handle are identical will often cover this. When that
-             * fails we have no option but to fail for now. */
-            P11PROV_debug("We can't really match private keys");
-            /* OTOH if group and pub point match either this is a broken key
-             * or the private key must also match */
-            cmp_type = OBJ_CMP_KEY_PUBLIC;
-        }
         break;
+
     case CKK_EC_EDWARDS:
         /* The EdDSA params can be encoded as printable string, which is
          * not recognized by OpenSSL and does not have respective EC_GROUP */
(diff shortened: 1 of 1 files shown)


src/obj/create.c
```diff
@@ -0,0 +1,117 @@
...
+                                       bool session_key, unsigned char *secret,
+                                       size_t secretlen)
+{
+    CK_SESSION_HANDLE sess = CK_INVALID_HANDLE;
+    CK_SESSION_INFO session_info;
+    CK_OBJECT_CLASS key_class = CKO_SECRET_KEY;
+    CK_KEY_TYPE key_type = CKK_GENERIC_SECRET;
+    CK_BBOOL val_true = CK_TRUE;
+    CK_BBOOL val_token = session_key ? CK_FALSE : CK_TRUE;
+    CK_ATTRIBUTE key_template[5] = {
```
Jakuje: once we touch it, I would prefer to get rid of the magic number 5 here and use either constant or dynamic size calculation such as following used elsewhere in the code:
```
size_t key_tmpl_size = sizeof(key_template) / sizeof(CK_ATTRIBUTE);
```

src/obj/export.c
```diff
@@ -0,0 +1,653 @@
...
+                params[attrnum] = OSSL_PARAM_construct_octet_string(
+                    OSSL_PKEY_PARAM_PUB_KEY, NULL, 0);
+                attrnum++;
+                break;
+            }
+        }
+    } else {
+        rv = CKR_OBJECT_HANDLE_INVALID;
+        goto done;
+    }
```
Jakuje, line 96: Out of scope of this PR, but here we should handle the ML-DSA and SLH-DSA certificates here when ready. What tests are using this code path? I though we have already the ML-DSA support for certificates and export already.

src/obj/fetch.c
```diff
@@ -0,0 +1,745 @@
...
+        CKO_PRIVATE_KEY, CKK_##name, name##_private_attrs, \
+            sizeof(name##_private_attrs) / sizeof(struct fetch_attrs), alloc \
+    }
+
+#define FILL_SECRET(name, alloc) \
+    { CKO_SECRET_KEY, CKK_##name, secret_key_attrs, \
+      sizeof(secret_key_attrs) / sizeof(struct fetch_attrs), alloc }
+#define FILL_KNOWN_SECRETS(alloc) \
+    FILL_SECRET(GENERIC_SECRET, alloc), FILL_SECRET(AES, alloc), \
+        FILL_SECRET(SHA_1_HMAC, alloc), FILL_SECRET(SHA256_HMAC, alloc), \
```
Jakuje: missing SHA224_HMAC here

src/obj/fetch.c
```diff
@@ -0,0 +1,745 @@
...
+        /* CKO_SECRET_KEY types */
+        a = p11prov_obj_get_attr(key, CKA_VALUE_LEN);
+        if (!a) {
+            /* Not all tokens support this attributes, it's ok */
+            P11PROV_debug("Failed to query key attribute CKA_VALUE_LEN");
+            rv = CKR_OK;
+        } else {
+            if (a->ulValueLen == sizeof(CK_ULONG)) {
+                key->data.key.size = *(CK_ULONG *)a->pValue;
+                key->data.key.bit_size = key->data.key.size * 8;
```
Jakuje: we do this in each of the branches of the switch. Should we move it out to avoid code duplication? Is there an use case where the bit size would not be a byte size * 8?

src/obj/fetch.c
```diff
@@ -0,0 +1,745 @@
...
+                rv = CKR_OK;
+            } else {
+                P11PROV_raise(key->ctx, CKR_GENERAL_ERROR,
+                              "Unsupported Key Size format");
+                rv = CKR_GENERAL_ERROR;
```
Jakuje: return here directly as in other cases in this function, rather than setting the `rv`?

src/obj/keymgmt.c
```diff
@@ -0,0 +1,928 @@
...
+    val = ec_params->pValue;
+    group = d2i_ECPKParameters(NULL, &val, ec_params->ulValueLen);
+    if (!group) {
+        ret = RET_OSSL_ERR;
+        goto done;
+    }
+
+    x = BN_CTX_get(bnctx);
+    y = BN_CTX_get(bnctx);
+    if (!y) {
```
Jakuje:
```suggestion
    if (!x || !y) {
```

src/obj/keymgmt.c
```diff
@@ -0,0 +1,928 @@
...
+}
+
+static int cmp_public_key_values(P11PROV_OBJ *pub_key1, P11PROV_OBJ *pub_key2)
+{
+    int ret;
+
+    switch (pub_key1->data.key.type) {
+    case CKK_RSA:
+        /* pub_key1 pub_key2 could be CKO_PRIVATE_KEY here but
+         *  nevertheless contain these two attributes
```
Jakuje: nit
```suggestion
         * nevertheless contain these two attributes
```

src/obj/keymgmt.c
```diff
@@ -0,0 +1,928 @@
...
+    case CKK_EC:
+    case CKK_EC_EDWARDS:
+        ret = cmp_attr(pub_key1, pub_key2, CKA_P11PROV_PUB_KEY);
+        break;
+    case CKK_ML_DSA:
+        ret = cmp_attr(pub_key1, pub_key2, CKA_VALUE);
+        break;
+    case CKK_ML_KEM:
+        ret = cmp_attr(pub_key1, pub_key2, CKA_VALUE);
+        break;
```
Jakuje, line 548: this is really the same and does not make sense to separate:
```suggestion
    case CKK_ML_DSA:
    case CKK_ML_KEM:
        ret = cmp_attr(pub_key1, pub_key2, CKA_VALUE);
        break;
```

src/obj/keymgmt.c
```diff
@@ -0,0 +1,928 @@
...
+    ret = get_attrs_from_cert(cert, attrs, num);
+    if (ret != CKR_OK) {
+        P11PROV_raise(priv_key->ctx, ret,
+                      "Failed to get public attrs from cert");
+        ret = RET_OSSL_ERR;
+        goto done;
+    }
+
+    switch (pub_key->data.key.type) {
+    case CKK_RSA:
```
Jakuje, line 593: TODO ML-DSA certificate comparsion missing

src/obj/keymgmt.c
```diff
@@ -0,0 +1,928 @@
...
+             * Generally you do not have 2 identical keys stored in to two
+             * separate objects so the initial shortcircuit that matches if
+             * slotid/handle are identical will often cover this. When that
+             * fails we have no option but to fail for now. */
+            P11PROV_debug("We can't really match private keys");
+            /* OTOH if param set and pub value match either this is a broken key
+             * or the private key must also match */
+            cmp_type = OBJ_CMP_KEY_PUBLIC;
+        }
+        break;
```
Jakuje: same blocks -- we should merge
```suggestion
    case CKK_ML_DSA:
    case CKK_ML_KEM:
        /* The parameter set maps to the specific key bit length that was
         * compared already. If the bit size matches, the param set matches. */
        if (cmp_type & OBJ_CMP_KEY_PRIVATE) {
            /* unfortunately we can't really read private attributes
             * and there is no comparison function in the PKCS11 API.
             * Generally you do not have 2 identical keys stored in to two
             * separate objects so the initial shortcircuit that matches if
             * slotid/handle are identical will often cover this. When that
             * fails we have no option but to fail for now. */
            P11PROV_debug("We can't really match private keys");
            /* OTOH if param set and pub value match either this is a broken key
             * or the private key must also match */
            cmp_type = OBJ_CMP_KEY_PUBLIC;
        }
        break;
```

src/obj/create.c
simo5: makes sense!

src/obj/export.c
simo5: not sure, can you open an issue about this?

src/obj/fetch.c
simo5: ugh ...

src/obj/fetch.c
simo5: there is one case where this is reversed (CKK_EC computes bit_size and derives key_size from it), so I ultimately decided to leave it there, it didn't make a lot of sense to have special case handling for this.

src/obj/fetch.c
simo5: ok

src/obj/keymgmt.c
simo5: ack, although I think this was pre-existing...

src/obj/export.c
Jakuje: #651. Done.

src/obj/keymgmt.c
simo5: this was in the original file and I did not touch it, I will leave as is for now

src/obj/keymgmt.c
simo5: same as bug above we need a ML-DSA cert and some tests to make sure this works.

src/obj/keymgmt.c
simo5: Ok I guess I'll make a single commit where some of these duplications are merged

simo5: oooh as I looked at this I see we have a big problem here. ML-DSA and ML-KEM private keys do not carry the public values, so this can't be done it will end up trying to source CKA_VALUE from the private key (which is not possible, and is not the public key CKA_VALUE anyway). I'll handle this appropriately

simo5: ah, fale alarm, the logic while a bit twisted actually works correctly, as long as at least one of the two keys is a public key, the sraight comparison will simply fail to fetch the CKA_VALUE so no incorrect comparison is made and then the match_public_keys() code will try to find a public key associated with the private, so I can just squash a bounch of these conditions with a comment.

PR #650, opened 2025-10-20
Split kmgmt.c in smaller files

simo5:
#### Description

Make operating on keymgmt provider a little easier by splitting into a file per each key type.

Depends on #649 
 
#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/kmgmt/common.c
+++ b/src/kmgmt/common.c
@@ -306,6 +306,9 @@ int p11prov_kmgmt_match(const void *keydata1, const void *keydata2,
     P11PROV_OBJ *key2 = (P11PROV_OBJ *)keydata2;
     int cmp_type = OBJ_CMP_KEY_TYPE;
 
+    P11PROV_debug("key (type: %ld) match %p %p %d", type, keydata1, keydata2,
+                  selection);
+
     if (key1 == key2) {
         return RET_OSSL_OK;
     }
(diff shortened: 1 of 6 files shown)


src/kmgmt/common.c
```diff
@@ -0,0 +1,513 @@
...
+        case CKA_SIGN_RECOVER:
+            set_bool_val(&template[i], sig);
+            break;
+        case CKA_DERIVE:
+            set_bool_val(&template[i], der);
+            break;
+        case CKA_WRAP:
+        case CKA_UNWRAP:
+            set_bool_val(&template[i], wrap);
+            break;
```
Jakuje: Is there the Key encapsulation usage to set too?

src/kmgmt/common.c
```diff
@@ -0,0 +1,513 @@
...
+        return RET_OSSL_ERR;
+    }
+
+    if (class == CKO_DOMAIN_PARAMETERS && type != CKK_EC) {
+        P11PROV_debug("import of domain parameters is only for EC keys");
+        /* There are no Domain parameters associated with an
+         * ECX or RSA, etc.. key in OpenSSL, so there is nothing really
+         * we can import */
+        return RET_OSSL_ERR;
+    }
```
Jakuje, line 437: thinking if merging this into the previous condition would make it easier to read?
```suggestion
    } else if (selection & OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS) {
        P11PROV_debug("import of domain parameters is only for EC keys");
        /* There are no Domain parameters associated with an
         * ECX or RSA, etc.. key in OpenSSL, so there is nothing really
         * we can import */
        if (type != CKK_EC) {
            return RET_OSSL_ERR;
        }
        class = CKO_DOMAIN_PARAMETERS;
    } else {
        return RET_OSSL_ERR;
    }
```

src/kmgmt/common.c
```diff
@@ -0,0 +1,513 @@
...
+                         void *cb_arg)
+{
+
+    P11PROV_OBJ *key = (P11PROV_OBJ *)keydata;
+    P11PROV_CTX *ctx = p11prov_obj_get_prov_ctx(key);
+    CK_OBJECT_CLASS class = p11prov_obj_get_class(key);
+    CK_KEY_TYPE type = p11prov_obj_get_key_type(key);
+    bool params_only = false;
+
+    P11PROV_debug("key %p import (type: %ld, selection: %d)", key, type,
```
Jakuje:
```suggestion
    P11PROV_debug("key %p export (type: %ld, selection: %d)", key, type,
```

src/kmgmt/ec.c
```diff
@@ -0,0 +1,924 @@
...
+                              "Unknown Curve %*s", (int)p->data_size,
+                              (char *)p->data);
+                return RET_OSSL_ERR;
+            }
+            ctx->data.ec.ec_params = ec_name_to_params[i].ec_param;
+            ctx->data.ec.ec_params_size = ec_name_to_params[i].ec_param_size;
+        }
+        break;
+    case CKK_EC_EDWARDS:
+        p = OSSL_PARAM_locate_const(params, "p11prov_edname");
```
Jakuje: Should we make some constant for this string, like `P11PROV_EDNAME` to avoid keeping constant strings all over the place?

src/kmgmt/ec.c
```diff
@@ -0,0 +1,924 @@
...
+                                                        sizeof(ED25519)),
+                                 OSSL_PARAM_END };
+    int ret;
+
+    P11PROV_debug("ed25519 gen_init %p", provctx);
+
+    /* we need to allow to initialize a generation of just domain parameters,
+     * as this is used by OpenSSL for ECDH, to set the expected parameters
+     * first and then import the received public peer key */
+    if ((selection & OSSL_KEYMGMT_SELECT_ALL) == 0) {
```
Jakuje, line 617: I think this comment applies only for the ECDSA and key derivation, which is not available in EdDSA, right?
(yes, it was in the original code).
The condition though is still relevant, I think.

src/kmgmt/ec.c
```diff
@@ -0,0 +1,924 @@
...
+    ctx = p11prov_kmgmt_gen_init(provctx, CKK_EC_EDWARDS,
+                                 CKM_EC_EDWARDS_KEY_PAIR_GEN);
+    if (!ctx) {
+        return NULL;
+    }
+
+    /* is this a parameter generator request ? */
+    if ((selection & OSSL_KEYMGMT_SELECT_KEYPAIR) == 0) {
+        ctx->data.ec.paramgen = true;
+    }
```
Jakuje, line 677: This is not used for anything in the eddsa keygen? lets drop that.

src/kmgmt/ec.c
```diff
@@ -0,0 +1,924 @@
...
+
```
Jakuje, line 632: Not used in eddsa. Leftover from the ECDSA, I think.

src/kmgmt/mldsa.c
```diff
@@ -0,0 +1,362 @@
...
+}
+
+static const OSSL_PARAM *p11prov_mldsa_gettable_params(void *provctx)
+{
+    static const OSSL_PARAM params[] = {
+        OSSL_PARAM_octet_string(OSSL_PKEY_PARAM_PUB_KEY, NULL, 0),
+        OSSL_PARAM_int(OSSL_PKEY_PARAM_BITS, NULL),
+        OSSL_PARAM_int(OSSL_PKEY_PARAM_SECURITY_BITS, NULL),
+        OSSL_PARAM_int(OSSL_PKEY_PARAM_MAX_SIZE, NULL),
+        OSSL_PARAM_utf8_string(OSSL_PKEY_PARAM_MANDATORY_DIGEST, NULL, 0),
```
Jakuje, line 300: you list this as a gettable parameter, but never handle it in the above `p11prov_mldsa_get_params()`.

src/kmgmt/rsa.c
```diff
@@ -0,0 +1,666 @@
...
+        return 152;
+    case 6144:
+        return 176;
+    case 7680:
+        return 192;
+    case 8192:
+        return 200;
+    case 15360:
+        return 256;
+    }
```
Jakuje: Is this part needed when the following block would return exactly the same values for these specific values?

src/kmgmt/common.c
simo5: I guess I can do this

src/kmgmt/ec.c
simo5: ok!

src/kmgmt/ec.c
simo5: I think it applies to ECDH using Edwards curves too.

When I look at EVP_PKEY_copy_parameters() in OpenSSL it ends up calling `evp_keymgmt_util_copy(to, (EVP_PKEY *)from, SELECT_PARAMETERS);` when both `from` and `to` keys are from the same provider.
With:
`# define SELECT_PARAMETERS OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS`

And this function is called for many key types, including RSA, DSA, DH and ML-KEM, as well as EC, so it may need to actually be applied more broadly, even if effectively it ends up doing almost nothing (it will set things like param_set for ML_DSA/ML-KEM keys or sec_bits to diffeentiate between Ed25519 and Ed448)

So I think I will leave this as is for now.

src/kmgmt/ec.c
simo5: Au contraire, p11prov_ec_gen() uses it and this function is used by both EC and ED providers

src/kmgmt/ec.c
simo5: it is used

src/kmgmt/mldsa.c
simo5: well spotted, we need to handle it.

src/kmgmt/rsa.c
simo5: I forget what the reasoning ever was here, so I'll simplify this, good catch

simo5:
Thanks for the review!!

PR #652, opened 2025-10-22
add further block cipher modes and extend tests

ep69:
#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:


src/cipher.c
```diff
@@ -276,11 +276,26 @@ static int set_iv(struct p11prov_cipher_ctx *ctx, const unsigned char *iv,
...
-        if (!ctx->mech.pParameter) {
-            return CKR_HOST_MEMORY;
+        if (ctx->mech.mechanism == CKM_AES_CTR) {
+            struct CK_AES_CTR_PARAMS *ctr_params = \
+                OPENSSL_malloc(sizeof(struct CK_AES_CTR_PARAMS));
+            if (!ctr_params) {
+                return CKR_HOST_MEMORY;
+            }
+            if (ivlen > 16) {
+                return CKR_GENERAL_ERROR; // TODO what error is approrpiate?
```
Jakuje: probably `CKR_MECHANISM_PARAM_INVALID`.

Jakuje: Please, move the check before the allocation to avoid a memory leak on error case

src/cipher.c
```diff
@@ -276,11 +276,26 @@ static int set_iv(struct p11prov_cipher_ctx *ctx, const unsigned char *iv,
...
      * and the mechanism requires it */
     if (iv != NULL && ivlen != 0) {
-        ctx->mech.pParameter = OPENSSL_memdup(iv, ivlen);
-        if (!ctx->mech.pParameter) {
-            return CKR_HOST_MEMORY;
+        if (ctx->mech.mechanism == CKM_AES_CTR) {
+            if (ivlen > 16) {
+                return CKR_MECHANISM_PARAM_INVALID;
+            }
+            struct CK_AES_CTR_PARAMS *ctr_params = \
```
Jakuje: The trailing backslash is not needed in c:
```suggestion
            struct CK_AES_CTR_PARAMS *ctr_params =
```

tests/tskey
```diff
@@ -30,37 +28,107 @@ else
...
+    local mode_name="$1" # mode, e.g., "CBC"
+    local cipher="$2" # ossl cipher without mode, e.g., "aes-128"
+    local input_len="$3" # plaintext length to test
+    local expected_outlen="$4" # expected output length
+    local key_mng="$5" # key management, "legacy" or "skey"
+
+    # generate plaintext of length $input_len
+    printf 'Plaintext%.0s' {1..100} |head -c "$input_len" >"${TMPPDIR}"/plaintext
+
+    mode_lower=$(echo "$mode_name" |tr '[:upper:]' '[:lower:]')
```
Jakuje, line 54: I would put here a comment why this is done this way rather than the obvious bash shortcut to prevent people trying to simplify that.

src/cipher.c
```diff
@@ -326,6 +342,9 @@ static CK_RV p11prov_cipher_prep_mech(struct p11prov_cipher_ctx *ctx,
 
     ret = p11prov_aes_set_ctx_params(ctx, params);
     if (ret != RET_OSSL_OK) {
+        P11PROV_debug("invalid mechanism param (ctx=%p, iv=%p, "
+                      "ivlen=%p, params=%p)",
+                      ctx, iv, ivlen, params);
```
Jakuje, line 347: coverity complains about the format string here (and on in the other instance of the same log):

>>>     Argument "ivlen" to format specifier "%p" was expected to have pointer type but has type "unsigned long".

PR #654, opened 2025-10-28
util: ensure that exact 32 character PIN works

numinit:
From the manpage of BIO_gets:

> BIO_gets() performs the BIOs "gets" operation and places the data in
> buf.  Usually this operation will attempt to read a line of data from
> the BIO of maximum length size - 1. [...]
>
> The returned string is always NUL-terminated and the '\n' is preserved
> if present in the input data.

It follows that we should pass the max pin length + 1 to BIO_gets, so the last byte isn't replaced by a NUL.

(This was reproducible in the tests, so update the tests).

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -274,7 +274,7 @@ done:
 static int get_pin_file(P11PROV_CTX *ctx, const char *str, size_t len,
                         void **output)
 {
-    char pin[MAX_PIN_LENGTH + 1];
+    char pin[MAX_PIN_LENGTH + 1] = { 0 };
     char *pinfile;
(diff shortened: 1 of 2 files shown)


simo5:
Wow, nice find!
Although I wonder if we should have such a tight limit on PINs?
Right now there is a variable holing MAX_PIN_LENGHT but in truth the largest length is defined by the token's info ulMaxPinLen ... so it seems this 32 byte limitation is quite artificial and may unduly restrict access to some tokens ...

PR #657, opened 2025-11-07
Implement support for key exchange with montgomery curves (x25519/x448)

### LINKED ISSUE #482: Add support for EDCH with x25519 and x448 (CKK_EC_MONTGOMERY)
**Describe the feature**
Kryoptic implements CKK_EC_MONTGOMERY, so we should add support for Key Exchange with x25519 and x448 curves, as we can test it.

simo5:
#### Description

Add key management support for curve25519 and curve448 keys, and key exchange (x25519/x448) providers for CKK_EC_MONTGOMERY keys.

Tests are provided with the Kryoptic token

NOTE: Most PKCS#11 implementation only accept OID for Edwards and Montgomery curve parameters instead of curveName, and Kryoptic accepts either. So this PR also changes the default encoding for Edwards and Montgomery curves CKA_EC_PARAMS to be OID based. 

Fixes #482

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -4,6 +4,21 @@
 #include "provider.h"
 #include <openssl/prov_ssl.h>
 
+/* Montgomery Curves */
+unsigned int x25519_group_id = 0x001d;
+unsigned int x25519_secbits = 128;
+int x25519_mintls = TLS1_2_VERSION;
+int x25519_maxtls = 0;
+int x25519_mindtls = DTLS1_2_VERSION;
+int x25519_maxdtls = 0;
+
+unsigned int x448_group_id = 0x001e;
+unsigned int x448_secbits = 224;
+int x448_mintls = TLS1_2_VERSION;
+int x448_maxtls = 0;
+int x448_mindtls = DTLS1_2_VERSION;
+int x448_maxdtls = 0;
+
 /* NIST EC */
 unsigned int p224_group_id = 0x0015;
 unsigned int p224_secbits = 112;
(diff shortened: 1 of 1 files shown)


out
Jakuje, line 1: should not be committed. Should we make sure its in gitignore?

src/provider.c
```diff
@@ -1216,6 +1219,9 @@ static CK_RV operations_init(P11PROV_CTX *ctx)
                              p11prov_ed448ph_signature_functions);
 #endif
                 break;
+            case CKM_EC_MONTGOMERY_KEY_PAIR_GEN:
+                ADD_ALGO(X25519, x25519, exchange, prop);
+                ADD_ALGO(X448, x448, exchange, prop);
```
Jakuje, line 1224: I think these should be enabled also for the following derive mechanisms:
```
            case CKM_ECDH1_DERIVE:
            case CKM_ECDH1_COFACTOR_DERIVE:
```

simo5: can't add more than once to the list

and they are only available if the token actually support montgomery curves, and if it does then CKM_ECDH1_DERIVE is ensured as well, as there is no point in doing montgomery curves if not for CKM_ECDH1_DERIVE.

out
simo5: I just need to make sure my autocommit scripts don't pick up random files :-D

Jakuje (approved):
good to go. No concerns noticed. Thank you for digging all the issues!

PR #665, opened 2025-11-18
Use CKA_PUBLIC_KEY_INFO when available

### LINKED ISSUE #659: Use  CKA_PUBLIC_KEY_INFO in encoders if available
Instead of always recomputing the subjectPublicKeyInfo, optimize the encoders code to use directly the contents of the CKA_PUBLIC_KEY_INFO if available on the key object.

This requires attempting to pull out the attribute when the key is fetched from the token.

simo5:
#### Description

Use CKA_PUBLIC_KEY_INFO when we have a private key but public key values are requested.
This can avoid searches on the token in some case, and can avoid errors in some cases where a private key is available but a corresponding public key is not stored in the token.

Fixes #659 

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/encoder.c
+++ b/src/encoder.c
@@ -555,7 +555,6 @@ static int p11prov_common_encoder_encode_text(void *inctx, OSSL_CORE_BIO *cbio,
     }
 
     if (selection & OSSL_KEYMGMT_SELECT_PUBLIC_KEY) {
-        CK_KEY_TYPE type = p11prov_obj_get_key_type(key);
         CK_ATTRIBUTE *pkeyinfo = NULL;
         P11PROV_OBJ *public = NULL;
         bool free_public = false;
@@ -597,7 +596,7 @@ static int p11prov_common_encoder_encode_text(void *inctx, OSSL_CORE_BIO *cbio,
             }
         }
         if (public) {
-            ret = p11prov_obj_export_public_key(public, type, true,
+            ret = p11prov_obj_export_public_key(public,
                                                 p11prov_print_public_key, out);
             if (ret != RET_OSSL_OK) {
                 BIO_printf(out, "[Error: Failed to decode public key data]\n");
@@ -693,8 +692,8 @@ static P11PROV_RSA_PUBKEY *p11prov_rsa_pubkey_to_asn1(P11PROV_OBJ *key)
         return NULL;
     }
 
-    ret = p11prov_obj_export_public_key(key, CKK_RSA, true,
-                                        p11prov_rsa_set_asn1key_data, asn1key);
+    ret = p11prov_obj_export_public_key(key, p11prov_rsa_set_asn1key_data,
+                                        asn1key);
 
     if (ret != RET_OSSL_OK) {
         P11PROV_RSA_PUBKEY_free(asn1key);
@@ -1095,9 +1094,8 @@ int p11prov_ec_pubkey_to_x509(X509_PUBKEY *pubkey, P11PROV_OBJ *key)
     struct ecdsa_key_point keypoint = { 0 };
     int ret;
 
-    ret =
-        p11prov_obj_export_public_key(key, CK_UNAVAILABLE_INFORMATION, false,
-                                      p11prov_ec_set_keypoint_data, &keypoint);
+    ret = p11prov_obj_export_public_key(key, p11prov_ec_set_keypoint_data,
+                                        &keypoint);
     if (ret != RET_OSSL_OK) {
         ecdsa_key_point_free(&keypoint);
         return ret;
@@ -1325,8 +1323,8 @@ int p11prov_mldsa_pubkey_to_x509(X509_PUBKEY *pubkey, P11PROV_OBJ *key)
         goto done;
     }
 
-    ret = p11prov_obj_export_public_key(
-        key, CKK_ML_DSA, true, p11prov_mldsa_set_keypoint_data, &keypoint);
+    ret = p11prov_obj_export_public_key(key, p11prov_mldsa_set_keypoint_data,
+                                        &keypoint);
     if (ret != RET_OSSL_OK) {
         goto done;
     }
@@ -1439,8 +1437,8 @@ int p11prov_mlkem_pubkey_to_x509(X509_PUBKEY *pubkey, P11PROV_OBJ *key)
         goto done;
     }
 
-    ret = p11prov_obj_export_public_key(
-        key, CKK_ML_KEM, true, p11prov_mlkem_set_keypoint_data, &keypoint);
+    ret = p11prov_obj_export_public_key(key, p11prov_mlkem_set_keypoint_data,
+                                        &keypoint);
     if (ret != RET_OSSL_OK) {
         goto done;
     }
--- a/src/obj/object.h
+++ b/src/obj/object.h
@@ -69,8 +69,7 @@ CK_RV p11prov_derive_key(P11PROV_OBJ *key, CK_MECHANISM *mechanism,
                          P11PROV_SESSION **_session, CK_OBJECT_HANDLE *dkey);
 const char *p11prov_obj_get_ec_group_name(P11PROV_OBJ *obj);
 bool p11prov_obj_get_ec_compressed(P11PROV_OBJ *obj);
-int p11prov_obj_export_public_key(P11PROV_OBJ *obj, CK_KEY_TYPE key_type,
-                                  bool search_related, OSSL_CALLBACK *cb_fn,
+int p11prov_obj_export_public_key(P11PROV_OBJ *obj, OSSL_CALLBACK *cb_fn,
                                   void *cb_arg);
 int p11prov_obj_export_params(P11PROV_OBJ *obj, OSSL_CALLBACK *cb_fn,
                               void *cb_arg);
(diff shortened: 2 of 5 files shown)


src/encoder.c
```diff
@@ -62,15 +80,257 @@ static void p11prov_encoder_freectx(void *ctx)
...
+    }
+
+    type = p11prov_obj_get_key_type(key);
+    switch (type) {
+    case CKK_RSA:
+        ret = p11prov_rsa_pubkey_to_x509(pubkey, key);
+        break;
+    case CKK_EC:
+        ret = p11prov_ec_pubkey_to_x509(pubkey, key);
+        break;
```
Jakuje, line 248: Missing handling of edwards and montgomery curves?

src/encoder.c
```diff
@@ -62,15 +80,257 @@ static void p11prov_encoder_freectx(void *ctx)
...
+        P11PROV_raise(ctx->provctx, CKR_GENERAL_ERROR, "Failed to init BIO");
+        ret = RET_OSSL_ERR;
+        goto done;
+    }
+
+    type = p11prov_obj_get_key_type(key);
+    if (type != CKK_RSA) {
+        /* we exclude RSA because we need to recompute a different
+         * SubjectPublicKeyInfo in the RSA-PSS case, PKCS#11 only stores
+         * a RSA-PKCS SubjectPublicKeyInfo*/
```
Jakuje, line 305: This is interesting idea. Should the PKCS#11 store the RSA-PSS SPKI already when the allowed mechanisms force the PSS mechanisms only?

src/encoder.c
```diff
@@ -88,69 +348,243 @@ static int p11prov_rsa_print_public_key(const OSSL_PARAM *params, void *bio)
...
+
+            nid = OBJ_obj2nid(poid);
+            BIO_printf(out, "ASN1 OID: %s\n", OBJ_nid2sn(nid));
+
+            name = EC_curve_nid2nist(nid);
+            if (name) {
+                BIO_printf(out, "NIST CURVE: %s\n", name);
+            }
+        } else if (ptype == V_ASN1_SEQUENCE) {
+            /* TODO */
```
Jakuje: should this fall back to some error until the TODO is resolved? Or print at least some warning?

src/obj/fetch.c
```diff
@@ -3,38 +3,46 @@
 
 #include "obj/internal.h"
 
+#define COMMON_KEY_ATTRIBUTES \
+    { { CKA_ID, NULL, 0 }, true, false }, \
+    { \
+        { CKA_LABEL, NULL, 0 }, true, false \
+    }
```
Jakuje: I hate this formatting, but I am not sure if we can make it looking better and more readable. Could some trailing coma make it formatted nicer?

src/obj/object.h
```diff
@@ -11,6 +11,8 @@
 #define CK_P11PROV_IMPORTED_HANDLE (CK_UNAVAILABLE_INFORMATION - 1)
 /* Special value for "new key" */
 #define CKO_P11PROV_NEW_KEY CKA_P11PROV_BASE + 1
+/* Special value for public key create for private on */
```
Jakuje: not sure if I understand the original comment correctly
```suggestion
/* Special value for public key created for private key */
```

src/obj/store.c
```diff
@@ -696,6 +691,13 @@ static CK_RV return_dup_key(P11PROV_OBJ *dst, P11PROV_OBJ *src)
         dst->numattrs++;
     }
 
+    rv = obj_add_to_pool(dst);
+    if (rv != CKR_OK) {
+        /* Perhaps we should fail here, but failing to add the object on the
+        * poolid will not prevent its usage, so we just print a debug line */
```
Jakuje: ?
```suggestion
         * pool id will not prevent its usage, so we just print a debug line */
```

tests/turi
```diff
@@ -4,15 +4,23 @@
 
 source "${TESTSSRCDIR}/helpers.sh"
 
+PROPQUERY="-propquery provider=pkcs11"
+if [[  "$TOKENTYPE" = "softhsm" ]]; then
+    # softhsm breaks if we force all operation son the token...
```
Jakuje:
```suggestion
    # softhsm breaks if we force all operations on the token...
```

src/encoder.c
simo5: sounds like it ... our encoder coverage is not great...

src/encoder.c
simo5: I thought about this, but it may make the code more complicated, because we'll need to account for tokens that creates a RSA-PKCS CKA_PUBLIC_KEY_INFO struct regardless of CKA_ALLOWED_ALGORITHMS.

Even in Kryoptic, if we were to do that then we'd need special code to re-encode CKA_PUBLIC_KEY_INFO when CKA_ALLOWED_ALGORITHMS changes?

It is kind of messy, so perhaps we should open a Kryoptic token to consider it, but currently we should leave this as is.

src/encoder.c
simo5: NO, I just need to implement the TODO :-D

src/obj/object.h
simo5: sigh: "created from a private one"

src/obj/store.c
simo5: poolid is the name of the filed .. but indentation will be fixed

src/obj/fetch.c
simo5: I tried, it makes other things much worse :(

simo5: check the fixup commit, let me know if that is acceptable

Jakuje: Much better!

src/encoder.c
Jakuje: Right. I did not think this would be actionable here and now, but it might be something to consider for kryoptic eventually. The specification is saying: 

> DER-encoding of the SubjectPublicKeyInfo for the associated public key (MAY be empty; DEFAULT derived from
the underlying private key data; MAY be manually set for specific key types; if set; MUST be consistent with the underlying private key data)

which suggests it should be the RSA-PSS one in cases where we have the limitation or related certificate having the PSS SPKI (but obviously even if some HW vendor would implement it, they would do it wrong). But we can try to make it right in kryoptic :)

simo5: Feel fre to open a tracker there :-D

Jakuje: Filled latchset/kryoptic#370.

src/encoder.c
```diff
@@ -88,69 +352,265 @@ static int p11prov_rsa_print_public_key(const OSSL_PARAM *params, void *bio)
...
+                pkeyinfo = p11prov_obj_get_attr(key, CKA_PUBLIC_KEY_INFO);
+                if (pkeyinfo && pkeyinfo->ulValueLen == 0) {
+                    pkeyinfo = NULL;
+                }
+            }
+            if (!public && !pkeyinfo) {
+                /* Finally try the most expensive option */
+                public = p11prov_obj_find_associated(key, CKO_PUBLIC_KEY);
+                free_public = true;
+            }
```
Jakuje: This order sounds to me odd -- first we try `CKA_PUBLIC_KEY_INFO`  from associated public key object, then the `CKA_PUBLIC_KEY_INFO` from the private key object and then we search for the associated object.
Should not the first two be swapped?

src/encoder.c
```diff
@@ -88,69 +352,265 @@ static int p11prov_rsa_print_public_key(const OSSL_PARAM *params, void *bio)
...
+        if (ptype == V_ASN1_OBJECT) {
+            const char *name;
+
+            nid = OBJ_obj2nid(pval);
+            BIO_printf(out, "ASN1 OID: %s\n", OBJ_nid2sn(nid));
+
+            name = EC_curve_nid2nist(nid);
+            if (name) {
+                BIO_printf(out, "NIST CURVE: %s\n", name);
+            }
```
Jakuje, line 451: I see this is just moved from elsewhere, but I can not find any documentation about this function in openssl. No manual page, just a function coming from nowhere in the header file. Should not we just print the "CURVE NAME' as in the case when we have the sequence below using `OSSL_EC_curve_nid2name()`?
The same function is used also in `p11prov_print_public_key()`.

In any case, filled openssl/openssl#29180

Jakuje (dismissed):

src/encoder.c
simo5: I guess we could swap them, should make no real difference. Note that p11prov_obj_get_associated() does *not* perform a search, so it is a very cheap operation.

src/encoder.c
simo5: It is a different name, OSSL_EC_curve_nid2name() gives the OpenSSL name, while EC_curve_nid2nist(0 gives the name as NIST recognizes it.

To be fair we can probably just remove printing the NIST name, it is not necessary ...

Jakuje: I was mostly stumped with not finding any documentation about that. Not sure how much different/better/worse the output of these two is.

src/encoder.c
simo5: pushed a fixup to check if reversing order works, if it does I will squash and merge, if not I will chuck and merge w/o this change

src/encoder.c
simo5: Just to be clear, this is what we print out in that case:
```
ASN1 OID: prime256v1
NIST CURVE: P-256
```

I'll leav this in for now, it is informational and causes no issues, should that function go away we can drop it w/o consequence.

simo5:
Thanks for the review

PR #668, opened 2026-01-08
Add support for AES-GCM.

### LINKED ISSUE #584: Add support for AES GCM ciphers
**Describe the feature**
AES GCM is currently missing

**Expected behavior**
AES GCM is available

**Additional context**
This should be implemented using the Message Encryption interfaces from pkcs#11 modules, rather than the older C_Encrypt* based functions

ilie-halip-nxp:
#### Description

Add support for AES-GCM (128, 192 and 256 bits key sizes), using the PKCS11 message-based encryption functions. This currently does not work with the partial IV feature used by TLS1.2, I plan to address that in the near future.

OpenSSL doesn't seem to provide a tool to test AEAD encryption/decryption (as is `openssl enc` for most ciphers), which is why I wrote a test program to do it.

Fixes #584 

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/tests/ttls
+++ b/tests/ttls
@@ -187,7 +187,12 @@ PROPQ="-propquery \"?provider=pkcs11\""
 # Try again forcing all operations on the token
 # We need to disable digest operations as OpenSSL depends on context duplication working
 ORIG_OPENSSL_CONF=${OPENSSL_CONF}
-sed -e "s/^#pkcs11-module-block-operations/pkcs11-module-block-operations = digest/" \
+BLOCKED_OPERATIONS="digest"
+if [[ "${SUPPORT_SYMMETRIC}" = "0" ]]; then
+    # For tokens that do not support symmetric key operations, disable these too
+    BLOCKED_OPERATIONS="${BLOCKED_OPERATIONS} cipher skeymgmt"
+fi
+sed -e "s/^#pkcs11-module-block-operations/pkcs11-module-block-operations = ${BLOCKED_OPERATIONS}/" \
     "${OPENSSL_CONF}" > "${OPENSSL_CONF}.forcetoken"
 OPENSSL_CONF=${OPENSSL_CONF}.forcetoken
 
--- a/tests/ttlsfuzzer
+++ b/tests/ttlsfuzzer
@@ -108,7 +108,12 @@ title SECTION "Run TLS fuzzer forcing the provider for all server operations"
 PROPQ="\"-propquery\", \"?provider=pkcs11\", "
 # We need to disable digest operations as OpenSSL depends on context duplication working
 ORIG_OPENSSL_CONF=${OPENSSL_CONF}
-sed -e "s/^#pkcs11-module-block-operations/pkcs11-module-block-operations = digest/" \
+BLOCKED_OPERATIONS="digest"
+if [[ "${SUPPORT_SYMMETRIC}" = "0" ]]; then
+    # For tokens that do not support symmetric key operations, disable these too
+    BLOCKED_OPERATIONS="${BLOCKED_OPERATIONS} cipher skeymgmt"
+fi
+sed -e "s/^#pkcs11-module-block-operations/pkcs11-module-block-operations = ${BLOCKED_OPERATIONS}/" \
     "${OPENSSL_CONF}" > "${OPENSSL_CONF}.forcetoken"
 export OPENSSL_CONF=${OPENSSL_CONF}.forcetoken
 


src/cipher.c
```diff
@@ -437,15 +506,24 @@ static int p11prov_cipher_legacy_init(void *ctx, CK_FLAGS op,
     return RET_OSSL_OK;
 }
 
+static CK_FLAGS p11prov_cipher_get_op(struct p11prov_cipher_ctx *ctx,
+                                      bool encrypt)
+{
+    bool message = (ctx->mech.mechanism == CKM_AES_GCM);
+    if (encrypt) return (message) ? CKF_MESSAGE_ENCRYPT : CKF_ENCRYPT;
```
simo5: please do not use compact conditional styles, change to:
```
if (ctx->mech.mechanism == CKM_AES_GCM) {
  if (encrypt) {
    return CKF_MESSAGE_ENCRYPT;
  } else {
    return CKF_MESSAGE_DECRYPT;
  }
} else { ...
```

src/cipher.c
```diff
@@ -437,15 +506,24 @@ static int p11prov_cipher_legacy_init(void *ctx, CK_FLAGS op,
     return RET_OSSL_OK;
 }
 
+static CK_FLAGS p11prov_cipher_get_op(struct p11prov_cipher_ctx *ctx,
+                                      bool encrypt)
```
simo5: instead of passing in a boolean, pass in the right flag to use as a CK_FLAGS:

```
static CK_FLAGS p11prov_cipher_get_op(struct p11prov_cipher_ctx *ctx, CK_FLAGS def_flag)
```

And later do:
```
{
    if (ctx->mech.mechanism == CKM_AES_GCM) {
        switch (def_flag) {
        case CKF_ENCRYPT:
            return CKF_MESSAGE_ENCRYPT;
            ...
        }
    }

    return def_flag;
}
```

src/cipher.c
```diff
@@ -701,6 +781,38 @@ static int p11prov_cipher_update(void *ctx, unsigned char *out, size_t *outl,
                                        (void *)in, inlen, out, &outlen);
         }
         break;
+    case CKF_MESSAGE_ENCRYPT:
+        if (in && inl && !out) {
```
simo5: When `cctx->tlsver` is not zero this won't be reachable because we check earlier that `out` must be equal to `in`. So it is unclear to me that this is correct here without an explicit exclusion for the GCM case in the earlier data checks at the very top of the function (this may be what is causing tls tests to fail?)

src/cipher.c
```diff
@@ -946,6 +1101,61 @@ static int p11prov_aes_set_ctx_params(void *vctx, const OSSL_PARAM params[])
...
+        if (ctx->mech.mechanism == CKM_AES_GCM) {
+            if (!ctx->mech.pParameter) {
+                ctx->mech.ulParameterLen = sizeof(CK_GCM_MESSAGE_PARAMS);
+                ctx->mech.pParameter = OPENSSL_zalloc(ctx->mech.ulParameterLen);
+                if (!ctx->mech.pParameter) return CKR_HOST_MEMORY;
+            }
+
+            CK_GCM_MESSAGE_PARAMS_PTR gcm =
+                (CK_GCM_MESSAGE_PARAMS_PTR)ctx->mech.pParameter;
+            if (!ivlen || ivlen > UINT32_MAX - 1) {
```
simo5: I am pretty sure an IV len longer than 96 is already broken, so we can reduce the ceiling greatly here.

src/cipher.c
ilie-halip-nxp: This code is reachable - the condition in that check is `if (!in || in != out || outsize < inl || !cctx->pad) { return; }` - so the check passes in this case: `in` is non-NULL (containing the GCM authentication data) and `out` is NULL.

simo5: Ah I read the first check on `in` backwards, nevermind

src/cipher.c
simo5: Please do not mark resolved until you actually address the comment.

ilie-halip-nxp: Sorry, I thought [the change in the latest fixup](https://github.com/latchset/pkcs11-provider/pull/668/commits/6e780ad45f4da3107dbda695fc687e1419466fc9#diff-fc1b2e03aa742953089557927d03e0e536b2bf0ebac5e5088eda7e9fab7aa007R549) addressed this.

src/cipher.c
```diff
@@ -11,6 +11,9 @@
 #include "openssl/rand.h"
 #include <string.h>
 
+#define BITS_TO_BYTES(bits) ((bits) / 8)
```
simo5: Technically this needs to be ((bits + 7) / 8), although in practice we always have bits expressed as multiples of eight..

src/cipher.c
simo5: I think github did a number on me, I saw you had pushed some new commits, which is why I re-reviewd the code, and then found nothing had changed in this area ... must have had the files tab somehow caching the old code and not updating.

Apologies.

ilie-halip-nxp:
I think you're right, yes. A quick gdb session in the rawhide container leads me to think the `C_SignUpdate()` calls fails because of the `session->getAllowMultiPartOp()` check from [here](https://github.com/softhsm/SoftHSMv2/blob/497481484a5c6c5edcb197d56a0f1706ba2d9cc9/src/lib/SoftHSM.cpp#L4746). For all mechanisms in the PR you linked, `bAllowMultiPartOp` is `false`.

For this PR, is it ok with you to ignore the rawhide softhsm errors for now?

Jakuje:
> I think you're right, yes. A quick gdb session in the rawhide container leads me to think the `C_SignUpdate()` calls fails because of the `session->getAllowMultiPartOp()` check from [here](https://github.com/softhsm/SoftHSMv2/blob/497481484a5c6c5edcb197d56a0f1706ba2d9cc9/src/lib/SoftHSM.cpp#L4746). For all mechanisms in the PR you linked, `bAllowMultiPartOp` is `false`.

Can you report the issue to softhsm developers on https://github.com/softhsm/SoftHSMv2/issues so it can get fixed?

> For this PR, is it ok with you to ignore the rawhide softhsm errors for now?

Merging the code in the current state will hide any other following failures so we will likely need to figure out some way to skip this test/softhsm to make the CI green. But I think it can be done outside of this PR.  Let me also have a look to the code and run covscan on that.

src/cipher.c
```diff
@@ -965,6 +1229,136 @@ static int p11prov_aes_set_ctx_params(void *vctx, const OSSL_PARAM params[])
...
+    if (p) {
+        size_t ivlen = 0;
+        int ret = OSSL_PARAM_get_size_t(p, &ivlen);
+        if (ret != RET_OSSL_OK) {
+            P11PROV_raise(ctx->provctx, CKR_MECHANISM_PARAM_INVALID,
+                          "Invalid AEAD IV Length parameter");
+            return RET_OSSL_ERR;
+        }
+
+        if (ctx->mech.mechanism == CKM_AES_GCM) {
```
Jakuje, line 1253: Should we have here the "else" branch with failure as these are only relevant for the GCM?

tests/meson.build
```diff
@@ -171,6 +172,7 @@ tests = {
   'ecxc': {'suites': ['softhsm', 'kryoptic', 'kryoptic.nss']},
   'cms': {'suites': ['softokn', 'kryoptic', 'kryoptic.nss']},
   'pinlock': {'suites': ['kryoptic']},
+  'aead': {'suites': ['softokn', 'kryoptic']},
```
Jakuje: Is there something that does not work with `kryoptic.nss` backend?

Jakuje:
Please, rebase on top of the current main. I merged #669 which should skip these broken mechanisms until this will get fixed on the softhsm side.

tests/meson.build
ilie-halip-nxp: Right, `kryoptic.nss` should be added here - initially I left it out because in my dev env it failed on setup and forgot to add it back.

src/cipher.c
ilie-halip-nxp: Sure thing, I added `else` branches here and for the other parameters.

simo5 (commented):
There are some style issues to be fixed.

Also could you avoid single line `if (..) return X;` constructs? We do not enforce that with clang-format (yet) but we have no code like that we always use:
```
if (..) {
    return X;
}
```

Aside from style issues, everythign else looks good, I will reun the coverity scan and then hopefully we can merge as soon as style-check passes.

Thanks.

simo5 (approved):
LGTM

PR #673, opened 2026-01-20
CHACHA20_POLY1305 support

ilie-halip-nxp:
#### Description

Add support for `CHACHA20_POLY1305` AEAD mechanism. This piggybacks on the current AES-GCM code because in most cases they behave the same.

The IV length is hardcoded to 12, and the tag length to 16. Both these values are specified in PKCS#11.

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/tests/taead
+++ b/tests/taead
@@ -23,14 +23,19 @@ ALGORITHMS=(
     AES-128-GCM
     AES-192-GCM
     AES-256-GCM
-    CHACHA20-POLY1305
+    ChaCha20-Poly1305
 )
 
 title PARA "Test AEAD support"
 
 for alg in "${ALGORITHMS[@]}"; do
     echo "Checking algorithm $alg"
-    $CHECKER "${TESTBLDDIR}/taead" "$alg" "$KEY" "$IV" "$AAD" "$DATA"
+    if $OPENSSL list -cipher-algorithms -propquery "provider=pkcs11" 2>/dev/null | grep -q "$alg"; then
+        title PARA "Run AEAD test with $alg"
+        $CHECKER "${TESTBLDDIR}/taead" "$alg" "$KEY" "$IV" "$AAD" "$DATA"
+    else
+        title PARA "Skipping AEAD test with $alg"
+    fi
 done
 
 echo "PASSED"
--- a/tests/taead.c
+++ b/tests/taead.c
@@ -12,7 +12,7 @@
 #include <openssl/core_names.h>
 #include "util.h"
 
-#if SKEY_SUPPORT == 1
+#if defined(OSSL_FUNC_CIPHER_ENCRYPT_SKEY_INIT)
 
 #define MAX_DATA_LEN 1024
 
@@ -271,7 +271,7 @@ int main(int argc, char *argv[])
                 argv[0]);
         fprintf(stderr,
                 "    Algorithms: AES-128-GCM, AES-192-GCM, AES-256-GCM,\n"
-                "                CHACHA20-POLY1305\n");
+                "                ChaCha20-Poly1305\n");
         exit(EXIT_FAILURE);
     }
 


ilie-halip-nxp:
I think at least some of the failures are due to some ciphers not being built into openssl. I added a check for cipher availability, let's see how the tests look.

src/cipher.c
```diff
@@ -204,6 +204,10 @@ static int p11prov_aes_get_params(OSSL_PARAM params[], int size, int mode,
         ciph_mode = EVP_CIPH_GCM_MODE;
         flags |= MODE_flag_aead | MODE_flag_custom_iv;
         break;
+    case MODE_poly1305:
+        ciph_mode = EVP_CIPH_STREAM_CIPHER;
+        flags |= MODE_flag_aead | MODE_flag_custom_iv;
+        break;
```
simo5, line 228: please create a separate p11prov_chacha_get_params() function for chacha/poly

src/cipher.c
```diff
@@ -864,18 +926,34 @@ static int p11prov_cipher_update(void *ctx, unsigned char *out, size_t *outl,
...
+                        cctx->mech.pParameter;
+                explicitivlen = 0;
+                taglen = EVP_CHACHAPOLY_TLS_TAG_LEN;
+
+                /* RFC7905: XOR the IV with the sequence number, which is
+                 * encoded in the AAD. This is only needed for TLS1.2, because
+                 * OpenSSL already does this for TLS1.3. */
+                for (size_t i = 0; i < 8; i++)
+                    chacha->pNonce[i + 4] ^= cctx->aad[i];
+            }
```
simo5, line 1072: This function is growing quite a bit, it may make sense to create some helper functions to isolate some of this code in smaller chuncks

src/cipher.c
```diff
@@ -1012,7 +1132,8 @@ static int p11prov_aes_get_ctx_params(void *ctx, OSSL_PARAM params[])
 
     if (cctx->mech.mechanism == CKM_AES_ECB) {
         ivsize = 0;
-    } else if (cctx->mech.mechanism == CKM_AES_GCM) {
+    } else if (cctx->mech.mechanism == CKM_AES_GCM
+               || cctx->mech.mechanism == CKM_CHACHA20_POLY1305) {
```
simo5: As above plese separate to a chacha specig get_ctx_params function.

You can create a common_get_ctx_params helper function for handling common parameters if necessary.

src/cipher.c
```diff
@@ -1497,4 +1642,46 @@ DISPATCH_TABLE_CIPHER_FN(aes, 128, gcm, CKM_AES_GCM);
...
+static void *p11prov_chacha20_dupctx(void *ctx)
+{
+    return NULL;
+}
+
+static int p11prov_chacha20_cipher(void *ctx, unsigned char *out, size_t *outl,
+                                   size_t outsize, const unsigned char *in,
+                                   size_t inl)
+{
+    return p11prov_aes_cipher(ctx, out, outl, outsize, in, inl);
```
simo5: if these are common functions rename them ad p11prov_common_<whatever>() and use them directly from both function tables

simo5: use defines to redefine the common functions if you want to keep using DISPATCH_TABLE_CIPHER_FN or just unroll defining the tables and remove that macro

simo5: that is:

#define p11prov_aes_cipher p11prov_common_cipher
#define p11prov_chacha20_cipher p11prov_common_cipher

static int p11prov_common_cipher(...) { .. }

simo5 (changes requested):
Mostly just code organization issues, I will take another pass once the two ciphers are better separated

src/cipher.c
```diff
@@ -1116,27 +1163,110 @@ static int p11prov_cipher_final(void *ctx, unsigned char *out, size_t *outl,
...
-    OSSL_PARAM *p;
-    int ret;
+    size_t ivsize = 16; /* 128 bits for most modes */
 
     if (cctx->mech.mechanism == CKM_AES_ECB) {
         ivsize = 0;
-    } else if (cctx->mech.mechanism == CKM_AES_GCM
-               || cctx->mech.mechanism == CKM_CHACHA20_POLY1305) {
+    } else if (cctx->mech.mechanism == CKM_AES_GCM) {
         ivsize = 12;
```
simo5: Storing `ivsize` in `struct p11prov_cipher_ctx` (initialized in the init function based on the mechanism) would avoid the need to have separate aes and chacha20 functions, and you could just have them defined as aliases of `p11prov_common_get_ctx_params()` like done elsewhere.

ilie-halip-nxp: How about a few defines such as `IVSIZE_mode` (where mode could be ctr, ecb, gcm, ....), and slightly modify the big macro in cipher.so like so:
```diff
     static int p11prov_##cipher##size##mode##_get_params(OSSL_PARAM params[]) \
     { \
-        return p11prov_##cipher##_get_params(params, size, MODE_##mode, \
+        return p11prov_##cipher##_get_params(params, size, IVSIZE_##mode, MODE_##mode, \
                                              mechanism); \
     } \
```

I think this looks better than having another (large) switch/case.

On an unrelated note, I found another issue - the AEAD tests aren't actually running anything. `SKEY_SUPPORT` is not defined for the test sources, so all the code is compiled away. I guess I rushed to fix the test failures and didn't notice it back then. Sorry about that, I'll get that fixed too.

simo5: IVSIZE_##mode sounds like a reasonable alternative.

You can fix the AEAD test stuff in a separate PR if you prefer to separate these concerns.

src/cipher.c
```diff
@@ -343,12 +372,17 @@ static int set_iv(struct p11prov_cipher_ctx *ctx, const unsigned char *iv,
...
 static CK_RV p11prov_cipher_prep_gcm(CK_MECHANISM_PTR mech,
-                                     const unsigned char *iv, size_t ivlen,
-                                     const OSSL_PARAM params[])
+                                     const unsigned char *iv, size_t ivlen)
 {
+    if (!mech) {
+        return CKR_ARGUMENTS_BAD;
+    }
+
+    if (!ivlen && ivlen > EVP_MAX_IV_LENGTH) {
```
simo5: This is hard to parse, is this trying to ensure (ivlen != 0) ? if that is the case please explicitly write it that way, we use !var basically only for pointers.

src/cipher.c
```diff
@@ -382,6 +418,52 @@ static CK_RV p11prov_cipher_prep_gcm(CK_MECHANISM_PTR mech,
...
+        mech->pParameter = OPENSSL_zalloc(mech->ulParameterLen);
+        if (!mech->pParameter) {
+            return CKR_HOST_MEMORY;
+        }
+    }
+
+    CK_SALSA20_CHACHA20_POLY1305_MSG_PARAMS_PTR chacha =
+        (CK_SALSA20_CHACHA20_POLY1305_MSG_PARAMS_PTR)mech->pParameter;
+
+    if (iv && ivlen) {
```
simo5: `&& ivlen != 0` please

src/cipher.c
simo5: Note that Covscan also mark this as dead code because I this translates to:
`if (ivlen == 0 && ivlen > EVP_MAX_IV_LENGHT)`
which is an impossible condition.

PR #676, opened 2026-01-30
Use accessors for ASN1_STRING values

bob-beck:
OpenSSL is going to make ASN1_STRING opaque. The accessors in question have been around in OpenSSL for a very long time.

see https://github.com/openssl/openssl/issues/29117
and specifically https://github.com/openssl/openssl/pull/29862
#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/encoder.c
+++ b/src/encoder.c
@@ -451,8 +451,8 @@ static int p11prov_print_pkeyinfo(CK_ATTRIBUTE *pkeyinfo, BIO *out)
             }
         } else if (ptype == V_ASN1_SEQUENCE) {
             const ASN1_STRING *pstr = pval;
-            const unsigned char *pm = pstr->data;
-            int pmlen = pstr->length;
+            const unsigned char *pm = ASN1_STRING_get0_data(pstr);
+            int pmlen = ASN1_STRING_length(pstr);
             EC_GROUP *group;
 
             group = d2i_ECPKParameters(NULL, &pm, pmlen);
@@ -1054,6 +1054,7 @@ static int p11prov_ec_set_keypoint_data(const OSSL_PARAM *params, void *key)
         keypoint->curve_type = V_ASN1_OBJECT;
     } else {
         EC_GROUP *group = EC_GROUP_new_from_params(params, NULL, NULL);
+        int len = 0;
         if (!group) {
             return RET_OSSL_ERR;
         }
@@ -1063,12 +1064,14 @@ static int p11prov_ec_set_keypoint_data(const OSSL_PARAM *params, void *key)
             EC_GROUP_free(group);
             return RET_OSSL_ERR;
         }
-        pstr->length = i2d_ECPKParameters(group, &pstr->data);
+        unsigned char *buf = NULL;
+        len = i2d_ECPKParameters(group, &buf);
         EC_GROUP_free(group);
-        if (pstr->length <= 0) {
-            ASN1_STRING_free(pstr);
+        if (len <= 0) {
+            OPENSSL_free(buf);
             return RET_OSSL_ERR;
         }
+        ASN1_STRING_set0(pstr, buf, len);
         keypoint->curve.sequence = pstr;
         keypoint->curve_type = V_ASN1_SEQUENCE;
     }
--- a/src/obj/keymgmt.c
+++ b/src/obj/keymgmt.c
@@ -332,12 +332,14 @@ CK_RV decode_ec_point(P11PROV_CTX *provctx, CK_KEY_TYPE key_type,
         }
     }
 
-    ec_point->data = octet->data;
-    ec_point->length = octet->length;
-
-    /* moved octet data, do not free it */
-    octet->data = NULL;
-    octet->length = 0;
+    ec_point->data =
+        OPENSSL_memdup(ASN1_STRING_get0_data(octet), ASN1_STRING_length(octet));
+    if (!ec_point->data) {
+        ret = CKR_HOST_MEMORY;
+        ec_point->length = 0;
+        goto done;
+    }
+    ec_point->length = ASN1_STRING_length(octet);
 
     ret = CKR_OK;
 done:
@@ -410,11 +412,11 @@ CK_RV p11prov_obj_set_ec_encoded_public_key(P11PROV_OBJ *key,
                                             const void *pubkey,
                                             size_t pubkey_len)
 {
-    CK_RV rv;
+    CK_RV rv = CKR_GENERAL_ERROR;
     CK_ATTRIBUTE *pub;
     CK_ATTRIBUTE *ecpoint;
     CK_ATTRIBUTE new_pub;
-    ASN1_OCTET_STRING oct;
+    ASN1_OCTET_STRING *oct = NULL;
     unsigned char *der = NULL;
     int add_attrs = 0;
     int len;
@@ -427,7 +429,8 @@ CK_RV p11prov_obj_set_ec_encoded_public_key(P11PROV_OBJ *key,
         /* not matching, error out */
         P11PROV_raise(key->ctx, CKR_KEY_INDIGESTIBLE,
                       "Cannot change public key of a token object");
-        return CKR_KEY_INDIGESTIBLE;
+        rv = CKR_KEY_INDIGESTIBLE;
+        goto done;
     }
 
     switch (key->data.key.type) {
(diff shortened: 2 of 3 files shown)


src/obj/keymgmt.c
```diff
@@ -332,12 +332,14 @@ CK_RV decode_ec_point(P11PROV_CTX *provctx, CK_KEY_TYPE key_type,
         }
     }
 
-    ec_point->data = octet->data;
-    ec_point->length = octet->length;
-
-    /* moved octet data, do not free it */
-    octet->data = NULL;
-    octet->length = 0;
+    ec_point->data =
```
simo5, line 335: This will add more memory allocation and free... making this whole code slower and more error prone, sad.

src/obj/store.c
```diff
@@ -721,12 +721,17 @@ static CK_RV fix_ec_key_import(P11PROV_OBJ *key, int allocattrs)
...
+        return CKR_GENERAL_ERROR;
+    }
+    if (!ASN1_STRING_set(oct, pub->pValue, pub->ulValueLen)) {
+        ASN1_OCTET_STRING_free(oct);
+        return CKR_GENERAL_ERROR;
+    }
     /* FIXME: should we set just bytes for Edwards and Montgomery keys? */
-    len = i2d_ASN1_OCTET_STRING(&oct, &der);
+    len = i2d_ASN1_OCTET_STRING(oct, &der);
+    ASN1_OCTET_STRING_free(oct);
```
simo5: using a done label at the end of he function and goto done statements on error is usually what we do to handle memory allocation in functions.

src/encoder.c
```diff
@@ -1054,22 +1054,18 @@ static int p11prov_ec_set_keypoint_data(const OSSL_PARAM *params, void *key)
...
+        len = i2d_ECPKParameters(group, &buf);
         EC_GROUP_free(group);
-        if (pstr->length <= 0) {
-            ASN1_STRING_free(pstr);
+        if (len <= 0) {
+            OPENSSL_free(buf);
             return RET_OSSL_ERR;
         }
-        keypoint->curve.sequence = pstr;
+        ASN1_STRING_set0(keypoint->curve.sequence, buf, len);
```
simo5: keypoint is initialized to an empty structure, so keypoint->curve.sequence here is NULL ..

bob-beck: ahh will fix

src/obj/store.c
bob-beck: No problem, was just sort of matching the style where you did this in the other function (and it was done this way). As I prefer that I'll just change 'em both.

src/obj/keymgmt.c
bob-beck: Yes, sadly a consequence of making the type opaque, and returning const pointers to the data. The goal here being so that we an return values from inside a larger const object, and reduce the (vast) numbers of copies OpenSSL makes for this sort of thing, which will increase performance.

simo5: Maybe provide an explicit function that allows to steal the buffer?
That function can internally decide to return a new allocated buffer if stealing is not possible.

src/obj/store.c
bob-beck: done (and converted the other one I touched to doing this as well) - if you'd prefer not to have that one converted to single return, I'm happy to undo that.

src/encoder.c
bob-beck: derp. much simpler now and closer to what you had

src/obj/keymgmt.c
```diff
@@ -495,24 +501,36 @@ CK_RV p11prov_obj_set_ec_encoded_public_key(P11PROV_OBJ *key,
...
-    len = i2d_ASN1_OCTET_STRING(&oct, &der);
+    oct = ASN1_OCTET_STRING_new();
+    if (!oct) {
+        goto done;
+    }
+    if (!ASN1_STRING_set(oct, pubkey, pubkey_len)) {
+        goto done;
+    }
+    len = i2d_ASN1_OCTET_STRING(oct, &der);
+    ASN1_OCTET_STRING_free(oct);
```
simo5: remove this one it will cause a double-free at function exit

src/obj/keymgmt.c
```diff
@@ -495,24 +501,36 @@ CK_RV p11prov_obj_set_ec_encoded_public_key(P11PROV_OBJ *key,
...
     }
 
-    oct.data = (unsigned char *)pubkey;
-    oct.length = (int)pubkey_len;
-    oct.flags = 0;
-
-    len = i2d_ASN1_OCTET_STRING(&oct, &der);
+    oct = ASN1_OCTET_STRING_new();
+    if (!oct) {
+        goto done;
```
simo5, line 510: you need to set rv = CKR_HOST_MEMORY here otherwise we'll return CKR_OK on error...

src/obj/keymgmt.c
```diff
@@ -495,24 +501,36 @@ CK_RV p11prov_obj_set_ec_encoded_public_key(P11PROV_OBJ *key,
...
+    }
+    if (!ASN1_STRING_set(oct, pubkey, pubkey_len)) {
+        goto done;
```
simo5, line 514: rv = CKR_GENERAL_ERROR;

src/obj/store.c
```diff
@@ -703,43 +703,56 @@ CK_RV p11prov_obj_copy_key_data(P11PROV_OBJ *dst, P11PROV_OBJ *src)
...
                       allocattrs);
-        return CKR_GENERAL_ERROR;
+        goto done;
     }
 
     pub = p11prov_obj_get_attr(key, CKA_P11PROV_PUB_KEY);
     if (!pub) {
         P11PROV_raise(key->ctx, CKR_KEY_INDIGESTIBLE, "No public key found");
-        return CKR_KEY_INDIGESTIBLE;
+        ret = CKR_KEY_INDIGESTIBLE;
```
simo5, line 720: given you already changing stuff here, please set ret before invoking P11PROV_raise() and pass ret to that function so we do not risk diverging errors on future modifications.

src/obj/store.c
```diff
@@ -703,43 +703,56 @@ CK_RV p11prov_obj_copy_key_data(P11PROV_OBJ *dst, P11PROV_OBJ *src)
...
+        goto done;
+    }
     /* FIXME: should we set just bytes for Edwards and Montgomery keys? */
-    len = i2d_ASN1_OCTET_STRING(&oct, &der);
+    len = i2d_ASN1_OCTET_STRING(oct, &der);
     if (len < 0) {
         P11PROV_raise(key->ctx, CKR_KEY_INDIGESTIBLE,
                       "Failure to encode EC point to DER");
-        return CKR_KEY_INDIGESTIBLE;
+        ret = CKR_KEY_INDIGESTIBLE;
```
simo5, line 735: as above if you don't mind please set ret before P11PROV_raise() and use it there too.

src/obj/keymgmt.c
bob-beck: Sounds like many concerns (between is the pointer the start of an area that can be freed, is the pointer to a read only object, what are you going to do with it?) for the general case that we have to track, for what based on how we've seen the conversions go here is a pretty specialized use case, which might end up
just always allocating a copy to be safe.

Which means the safe answer is just to have someone who wants to do this (take the raw bytes and convert to their own datatype instead of keeping it in ASN1_STRING) makes a copy.

bob-beck: (with the additional issue that the accessors in question are available pretty much in any version or fork of OpenSSL, they've been around forever - new api such as this would only be available going forward)

src/obj/keymgmt.c
bob-beck: To be honest this one is also CKR_HOST_MEMORY :) 

that's the only reason it would fail, unless you prefer GENERAL_ERROR here.

simo5: CKR_HOST_MEMORY is also ok.

src/obj/keymgmt.c
bob-beck: yeah, missed that shuffling, thanks.

src/obj/keymgmt.c
bob-beck: Also changed the default at the top to GENERAL_ERROR to not fall into that trap.

simo5 (dismissed):
LGTM,
please squash the fixups and add your sign-off to the commit

bob-beck:
> Unfortunately it seems coverity is having some extended maintenance issues, I can waive it off, but the DCO check is required. @bob-beck can you please ass the sign-off-by trailer to your commit ? (`git commit -s --amend`)

ok, I think I have?

simo5:
@bob-beck sorry to add more churn, but we just merged the coverity replacement. Would you mind rebasing on main so we can get the csbuild tests running ?

bob-beck:
On Wed, Feb 11, 2026 at 10:33:10PM +0000, Simo Sorce wrote:
> simo5 left a comment (latchset/pkcs11-provider#676)
> 
> @bob-beck sorry to add more churn, but we just merged the coverity replacement. Would you mind rebasing on main so we can get the csbuild tests running ?

Sure, gimme a sec

simo5:
Thanks for the contribution! (and your patience with CI woes)

PR #683, opened 2026-02-11
Add csbuild to CI

Jakuje:
#### Description

And fix some of the issues reported by the tool so far.

#### Checklist
- [ ] Documentation updated

#### Reviewer's checklist:


simo5:
So the differential build is finding (minor) issues but it is not marked as failed, how is this supposed to work ?
Also the "differential" build seem to be doing a full scan ?

src/encoder.c
```diff
@@ -307,7 +307,7 @@ static int p11prov_common_encoder_spki_der_encode(
     }
     if (pkeyinfo && pkeyinfo->ulValueLen > 0) {
         int len = BIO_write(out, pkeyinfo->pValue, pkeyinfo->ulValueLen);
-        if (len == pkeyinfo->ulValueLen) {
+        if (len >= 0 && (CK_ULONG)len == pkeyinfo->ulValueLen) {
             ret = RET_OSSL_OK;
```
simo5, line 311: pkeyinfo->ulValueLen is already guaranteed to be greater than 0, so I do not think we need to check `len >= 0`, this check is already implicit in the fact that what we compare len to is already assured > 0

src/util.c
```diff
@@ -1226,13 +1226,14 @@ CK_RV p11prov_digest_util(P11PROV_CTX *ctx, const char *digest,
         goto done;
     }
 
-    dlen = EVP_MD_get_size(md);
-    if (dlen == -1) {
+    err = EVP_MD_get_size(md);
+    if (err == -1) {
```
simo5: I think in this case we'll have to check `if (err < 0)` because any negative number will be bad news when later cast to an unsigned int.

src/encoder.c
Jakuje: The idea is that if we get negative `len` from openssl, converting it to `ULONG` will wrap it to large value. I think it will never be `pkeyinfo->ulValueLen`, but I though it might be better to avoid the wrap-around.

src/util.c
Jakuje: True. I was referring to the manual page which explicitly says:

> Returns the digest or block size in bytes or -1 for failure.

but it would be better to be safe with checking all negatives. Will fix.

PR #688, opened 2026-02-18
Release Version 1.2.0

### LINKED ISSUE #687: Release 1.2
Time to make a release, for 1.2

simo5:
#### Description

PR to verify Release 1.2 is in good shape.

Fixes #687

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -7,7 +7,8 @@ This table will specify which versions are supported:
 | Version | Supported          |
 | ------- | ------------------ |
 | 1.0     | :x:                |
-| 1.1     | :white_check_mark: |
+| 1.1.0   | :x:                |
+| 1.2.0   | :white_check_mark: |
 | ------- | ------------------ |
 
 ## Reporting a Vulnerability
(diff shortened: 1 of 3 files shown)


SECURITY.md
```diff
@@ -7,7 +7,8 @@ This table will specify which versions are supported:
 | Version | Supported          |
 | ------- | ------------------ |
 | 1.0     | :x:                |
-| 1.1     | :white_check_mark: |
+| 1.1     | :x:                |
+| 1.2     | :white_check_mark: |
```
Jakuje: The version is now 1.2.0

It should match in all places, also here. I am not completely sure what the following note on the release page means, but I think it means adding the last number for the version:

> NOTE: Before the next release reconcile meson version (currently 1.1) with tagging (currently 1.1.0) so that tarball versions and tags match.

https://github.com/latchset/pkcs11-provider/wiki/Notes-on-release-process

simo5: Yeah that was for meson.build and the spec file ... but I guess you have a point that we should do it in this table as well.

PR #697, opened 2026-02-26
Add coverage CI job

Jakuje:
#### Description

There is no such nice report like we have in gitlab, but this at least allows running `make coverage` to get HTML reports to see where we have untested code.

Local run shows to me ok-ish results:
```
Overall coverage rate:
  lines......: 64.8% (9971 of 15394 lines)
  functions......: 80.3% (771 of 960 functions)
  branches......: 38.3% (4780 of 12476 branches)
```

and drop leftovers of p11-kit-client.

#### Checklist
- [X] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,43 @@
+---
+name: coverage
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+
+jobs:
+  scan:
+    name: Coverage report
+    runs-on: ubuntu-22.04
+    container: quay.io/fedora/fedora:latest
+    steps:
+      - name: Install Dependencies
+        run: |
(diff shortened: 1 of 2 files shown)


simo5:
Unfortunately there does not seem to be any easy way to expose generate files from builds in github :/ But you can archive them as build artifacts that can be downloaded as a tar.gz and displayed in a browser

simo5 (dismissed):
LGTM

Jakuje:
Probably not. I mostly wanted to be able to get this to see where we have dead code and what might need some attention (and have makefile target to be able to run it locally). Given that github does not have a good representation, I think its easier to run it locally than downloading tarballs, but if you prefer, I can add that too.

One though I had would be differential scan to fail if the coverage goes down, but it would need two runs and some regexes. We can also include the tlsfuzzer as I think it has large coverage of odd mechanisms too. That should be quite easy (while slower).

Issue #719, opened 2026-04-08
Uncoditinal inclusion of generic attributes makes fetching slower


Jakuje:
**Describe the bug**
Fetching attributes is slow because when we include one attribute that is not present in the target key, whole operation fails and we fall back to querying one-by-one.

This is mostly visible since we introduced the support for PQC Keys with `CKA_PARAMETER_SET`, which is included in each of the queries and fails for non-pqc keys

Certificate fetching is "crippled" by including the `CKA_KEY_TYPE` too.

**To Reproduce**
1. Run any openssl command through pkcs11-provider
2. Review the debug log (search for  `Error 18 returned by C_GetAttributeValue`)

**Expected behavior**
A lot of errors like this with a lot of retries.

**Operating environment (please complete the following information):**
 - OS: Fedora
 - Version [e.g. 36] 42

**Token and application used (please complete the following information):**
 - Device: Kryoptic
 - PKCS11 Driver version: master
 - Application [e.g. Apache Httpd] - 
 - Version [e.g. 22] - 

**Additional context**
Optionally, we might want to implement something like we got in rust-cryptoki recently, which accelerates fetching of the potentially unsupported attributes to at most 3 round trips: https://github.com/parallaxsecond/rust-cryptoki/pull/341

Jakuje:
It might be only with the latchset/kryoptic#436 though. I actually checked the logs from this PR only so far.

The places where this happens is `p11prov_obj_from_handle()` if I see right.

simo5:
I think the helper may be too aggressive and rushes to split into individual calls trying to pull individual attributes when the error is clearly that those attributes are not available ... I am sure we can improve the situation here.

PR #722, opened 2026-04-09
obj: Do not assume CKR_ATTRIBUTE_TYPE_INVALID means the token does no…

Jakuje:
…t support the attribute at all

This return code means just the attribute is not present on the CURRENT object so assuming from this return code that this attribute is not supported at all is wrong and was just accidentally working for kryoptic, because we did not return the correct value when searching for attributes:

https://github.com/latchset/kryoptic/pull/436

#### Checklist

- [~] Test suite updated with functionality tests
- [~] Test suite updated with negative tests
- [~] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/obj/fetch.c
+++ b/src/obj/fetch.c
@@ -443,11 +443,6 @@ CK_RV p11prov_obj_from_handle(P11PROV_CTX *ctx, P11PROV_SESSION *session,
             if (ret == CKR_OK) {
                 obj->attrs[obj->numattrs] = a[0].attr;
                 obj->numattrs++;
-            } else if (ret == CKR_ATTRIBUTE_TYPE_INVALID) {
-                token_supports_allowed_mechs = CK_FALSE;
-                (void)p11prov_token_sup_attr(ctx, obj->slotid, SET_ATTR,
-                                             CKA_ALLOWED_MECHANISMS,
-                                             &token_supports_allowed_mechs);
             }
         }
         break;


Jakuje:
Ugh. So it looks like the Address Sanitizer is broken now due to the OpenSC update. It now fails with:
```
+ env LD_PRELOAD=/usr/lib64/libasan.so.8.0.0:/__w/pkcs11-provider/pkcs11-provider/builddir/tests/fake_dlclose.so pkcs11-tool --module=/usr/lib64/pkcs11/libsofthsm2.so '--token-label=SoftHSM Token' --login --pin=0123456789ABCDEFFEDCBA9876543210 --keypairgen --key-type=RSA:2048 --id=0000 --label=caCert
==643==You are trying to dlopen a /usr/lib64/pkcs11/libsofthsm2.so shared library with RTLD_DEEPBIND flag which is incompatible with sanitizer runtime (see https://github.com/google/sanitizers/issues/611 for details). If you want to run /usr/lib64/pkcs11/libsofthsm2.so library under sanitizers please remove RTLD_DEEPBIND from dlopen flags.
```

We introduced the deepbind to the pkcs11-tool/pkcs11-spy due to some other issues, but its really not playing well with sanitizers. So I guess I will extend the fake_dlclose to make sure the dlopen does not do the deepbind anywhere in the chain.

simo5:
So while this minimal change "works" I think we should basically reshape how we deal with this, but I guess we can tackle that as part of dealing with #719

simo5:
> Ugh. So it looks like the Address Sanitizer is broken now due to the OpenSC update. It now fails with:
> 
> ```
> + env LD_PRELOAD=/usr/lib64/libasan.so.8.0.0:/__w/pkcs11-provider/pkcs11-provider/builddir/tests/fake_dlclose.so pkcs11-tool --module=/usr/lib64/pkcs11/libsofthsm2.so '--token-label=SoftHSM Token' --login --pin=0123456789ABCDEFFEDCBA9876543210 --keypairgen --key-type=RSA:2048 --id=0000 --label=caCert
> ==643==You are trying to dlopen a /usr/lib64/pkcs11/libsofthsm2.so shared library with RTLD_DEEPBIND flag which is incompatible with sanitizer runtime (see https://github.com/google/sanitizers/issues/611 for details). If you want to run /usr/lib64/pkcs11/libsofthsm2.so library under sanitizers please remove RTLD_DEEPBIND from dlopen flags.
> ```
> 
> We introduced the deepbind to the pkcs11-tool/pkcs11-spy due to some other issues, but its really not playing well with sanitizers. So I guess I will extend the fake_dlclose to make sure the dlopen does not do the deepbind anywhere in the chain.

Yeah this is the only way I could think of too, I have seen this problem elsewhere already ...

Jakuje:
> So while this minimal change "works" I think we should basically reshape how we deal with this, but I guess we can tackle that as part of dealing with #719

Technically yes, but the query for the ALLOWED_MECHANISMS is now separate GetAttributes call which makes it slightly orthogonal and wanted to keep it separate from potentially large change that would result from 719.

Jakuje:
Sigh. For some reason the dlopen from openssl is caught, but the dlopen from opensc is not with the change I wrote. I am not sure when I will be able to get back to this to dig into this.

For the record, we had very long discussion about the DEEPBIND in OpenSC/OpenSC#2875, but it brings just more issues than it solves ...

simo5:
@Jakuje can you rebase this on main so that addres sanitizer test can pass?

PR #726, opened 2026-04-27
Fix default rsa sign pss for short rsa key

addmors:
For RSA signature in PSS mode with a short MODULE, the length must be the maximum possible, and not equal to the hash length. For example rsa 512 sha256 is 32 +32 + 2 = 66 bytes but module is 64 bytes and that is incorrect by default.

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/sig/rsasig.c
+++ b/src/sig/rsasig.c
@@ -285,6 +285,8 @@ static CK_RV p11prov_sig_pss_restrictions(P11PROV_SIG_CTX *sigctx,
     return CKR_OK;
 }
 
+static int p11prov_rsasig_set_pss_saltlen_max(void *ctx, bool max_to_digest);
+
 /* fixates pss_params based on defaults if values are not set */
 static CK_RV pss_defaults(P11PROV_SIG_CTX *sigctx)
 {
@@ -299,15 +301,11 @@ static CK_RV pss_defaults(P11PROV_SIG_CTX *sigctx)
         sigctx->pss_params.mgf = data->mgf;
     }
     if (sigctx->pss_params.sLen == 0) {
-        /* default to digest size if not set */
-        size_t size;
-        CK_RV ret;
-
-        ret = p11prov_digest_get_digest_size(data->digest, &size);
-        if (ret != CKR_OK) {
-            return ret;
+        /* default to digest size or max awailable if not set */
+        int ret = p11prov_rsasig_set_pss_saltlen_max(sigctx, true);
+        if (ret != RET_OSSL_OK) {
+            return CKR_MECHANISM_INVALID;
         }
-        sigctx->pss_params.sLen = size;
     }
 
     return CKR_OK;
@@ -341,6 +339,7 @@ static int p11prov_rsasig_set_pss_saltlen_max(void *ctx, bool max_to_digest)
     size_t digest_size;
     CK_ULONG key_size;
     CK_ULONG key_bit_size;
+    CK_ULONG min_required_size;
     CK_RV rv;
 
     if (sigctx->digest == 0) {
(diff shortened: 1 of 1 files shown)


Jakuje:
why did you move the functions around? The diff makes it very hard to review.

addmors:
pss_defaults was higher than p11prov_rsasig_set_pss_saltlen_max

So I had to move it

addmors:
I corrected it stylistically, and it detected false positives from the sanitizer and statistical analyzer, since the code they were called by did not change, and did not depend on it either.

Jakuje:
Does this have any practical implications for reasonable key sizes? Your example with RSA 512b key is really not very practical. There is a know limitation for small RSA modulus with larger hashes, but I am not completely sure this is the right solution for the problem. This forces the salt length to be significantly larger than needed for larger RSA keys, isn't it?

src/sig/rsasig.c
addmors: according to this line of code, if the digest size is available, then the digest size is applied, otherwise the maximum allowed one is applied

addmors: max_to_digest is true for this function in pss_default

addmors: All these changes are exclusively for smaller sizes; for larger sizes, the execution logic has not changed.

Jakuje: Thanks for the clarification! In that case, I think the function name is a bit confusing, but thats not your fault.

src/sig/rsasig.c
```diff
@@ -377,6 +327,52 @@ static int p11prov_rsasig_set_pss_saltlen_max(void *ctx, bool max_to_digest)
...
+        int ret = p11prov_rsasig_set_pss_saltlen_max(sigctx, true);
+        if (ret != RET_OSSL_OK) {
+            return CKR_MECHANISM_INVALID;
+        }
+    }
+
+    return CKR_OK;
+}
+
+static int p11prov_rsasig_set_pss_saltlen_from_digest(void *ctx)
```
Jakuje, line 316: This function is left unused (if I see right), why not remove it then?

simo5 (changes requested):
move the functions back where they were, and just pre-declare the static function you need to use earlier.

I will do further review after this is done.

addmors:
https://github.com/openssl-projects/pkcs11-provider/pull/726#pullrequestreview-4181840227
fix

src/sig/rsasig.c
addmors: in rsasig_set_saltlen

in case 

if (saltlen == RSA_PSS_SALTLEN_DIGEST) {
    return p11prov_rsasig_set_pss_saltlen_from_digest(sigctx);
}

simo5 (changes requested):
Code is much more readable now.

Looking at it though, if you really expect to use this with small key sizes you need to add a check in p11prov_rsasig_set_pss_saltlen_max() for `key_size > digest_size -2` and bail out if not, otherwise sLen will wrap to a huge unsigned long.

addmors:
https://github.com/openssl-projects/pkcs11-provider/pull/726#pullrequestreview-4183050222
fix

src/sig/rsasig.c
```diff
@@ -366,11 +365,16 @@ static int p11prov_rsasig_set_pss_saltlen_max(void *ctx, bool max_to_digest)
         return RET_OSSL_ERR;
     }
 
-    /* from openssl */
-    sigctx->pss_params.sLen = key_size - digest_size - 2;
-    if ((key_bit_size & 0x07) == 1) {
-        sigctx->pss_params.sLen -= 1;
+    min_required_size = digest_size + 2 + (((key_bit_size & 0x07) == 1) ? 1 : 0);
```
simo5: please do:
```
min_required_size = digest_size + 2;
if ((key_bit_size & 0x07) == 1) {
    min_required_size += 1;
}
```

src/sig/rsasig.c
```diff
@@ -366,11 +365,16 @@ static int p11prov_rsasig_set_pss_saltlen_max(void *ctx, bool max_to_digest)
...
+
+    if (key_size < min_required_size) {
+        P11PROV_raise(sigctx->provctx, rv, "Unavailable digest for current key size");
```
simo5: This message is incorrect, it should say something like "key size too small for the requested digest"

src/sig/rsasig.c
```diff
@@ -354,7 +353,7 @@ static int p11prov_rsasig_set_pss_saltlen_max(void *ctx, bool max_to_digest)
         P11PROV_raise(sigctx->provctx, rv, "Unavailable digest");
         return RET_OSSL_ERR;
     }
-
+    
     key_size = p11prov_obj_get_key_size(sigctx->key);
     if (key_size == CK_UNAVAILABLE_INFORMATION) {
         P11PROV_raise(sigctx->provctx, rv, "Unavailable key");
```
simo5, line 360: rv is unset here, can you set it to CKR_KEY_INDIGESTIBLE ?

src/sig/rsasig.c
simo5, line 365: rv is unset here, can you set it to CKR_KEY_INDIGESTIBLE ?

src/sig/rsasig.c
simo5: also rv is unset here, can you set it to CKR_KEY_INDIGESTIBLE ?

simo5 (changes requested):
Just some tidyups othewise looks good to me.

We'll have to rebase on #729 once it is merged to check that address sanitizer is happy

addmors:
all fix) wait https://github.com/openssl-projects/pkcs11-provider/pull/729

src/sig/rsasig.c
Jakuje: sorry. My bad. Thanks for checking.

simo5:
FYI #729 was merged so address-sanitizer tests should pass now if you rebase on main

simo5:
@addmors hey why did you close this one ?

simo5 (approved):
LGTM

addmors:
@simo5 sorry rebase in github is merge, i fix history of commit for correct mergin

simo5:
ignore the scan build failure, it is fixed in #730

addmors:
sigctx->mechanism.mechanism = data->pkcs_pss;
 
 need to fix line like:
 `if (sigctx->digest_op && data) {`
 or skip it?)

simo5:
> sigctx->mechanism.mechanism = data->pkcs_pss;
> 
> need to fix line like: `if (sigctx->digest_op && data) {` or skip it?)

see #730

addmors:
sanitaize pass. passes looks good)

addmors:
> > sigctx->mechanism.mechanism = data->pkcs_pss;
> > need to fix line like: `if (sigctx->digest_op && data) {` or skip it?)
> 
> see #730

Or like this, thanks)

addmors:
looks like an occasional false positive.

simo5:
Thank you for the contribution! Appreciated the fast turn-around on the reviews!

PR #727, opened 2026-04-27
Asan workaround

Jakuje:
#### Description

Workaround for change in OpenSC forcing the use of RTLD_DEEPBIND in dlopen().

The respective build for Fedora rawhide:

https://koji.fedoraproject.org/koji/taskinfo?taskID=144902342

And respective fix in OpenSC:

https://github.com/OpenSC/OpenSC/pull/3670/

#### Checklist

- [X] No Code modified for feature

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/address-sanitizer.yml
+++ b/.github/workflows/address-sanitizer.yml
@@ -18,7 +18,7 @@ jobs:
         # , debian]
         include:
           - name: fedora
-            container: quay.io/fedora/fedora:latest
+            container: quay.io/fedora/fedora:rawhide
         #  - name: debian
         #    container: debian:sid
     container: ${{ matrix.container }}
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -85,7 +85,7 @@ if get_option('b_sanitize') == 'address'
   add_env = {
     'ASAN_OPTIONS': 'fast_unwind_on_malloc=0:detect_leaks=1',
     'LSAN_OPTIONS': 'suppressions=@0@/lsan.supp'.format(meson.current_source_dir()),
-    'FAKE_DLCLOSE': fake_dlclose.full_path(),
+    'OPENSC_DEEPBIND': '0',
     'CHECKER': checker,
   }
 


simo5:
Looks good, I will approve now and feel free to merge once that feature lands in rawhide

simo5 (approved):

simo5:
should we still pursue this given we found a simpler workaround for now?

addmors:
Using only the openssl utility under sanitizer can eliminate the need to troubleshoot problems and find workarounds in other utilities. However, it's up to you.

simo5:
> Using only the openssl utility under sanitizer can eliminate the need to troubleshoot problems and find workarounds in other utilities. However, it's up to you.

That is what I implemented in #729

addmors:
So you're only skipping pkcs11-tool. For example, echo still runs with the sanitizer (I don't think there will be any problems with it), but suddenly there are other utilities that also need to be run without it. And the environment is configured to run everything under the sanitizer.

addmors:
I agree that my method will lead to significant changes in CI jobs. Аnd its use now is most likely overhead

simo5:
> So you're only skipping pkcs11-tool. For example, echo still runs with the sanitizer (I don't think there will be any problems with it), but suddenly there are other utilities that also need to be run without it. And the environment is configured to run everything under the sanitizer.

As far as I can tell we interpose $CHECKER (which defines all of the env vars to pull in libasan) only in the ossl() helper which only invokes the openssl binary.

simo5:
> > So you're only skipping pkcs11-tool. For example, echo still runs with the sanitizer (I don't think there will be any problems with it), but suddenly there are other utilities that also need to be run without it. And the environment is configured to run everything under the sanitizer.
> 
> As far as I can tell we interpose $CHECKER (which defines all of the env vars to pull in libasan) only in the ossl() helper which only invokes the openssl binary.

I was focused only on the setup phase, but yes we also call checker in front of every compiled tests in the test suite, but that is intentional, which tests run echo under the checker ?

addmors:
I made this conclusion based on the fact that unset LD_PRELOAD fixed the sanitizer error.

Most likely this means that sanitizer is inside the environment in the LD_PRELOAD variable, right?)

simo5:
> I made this conclusion based on the fact that unset LD_PRELOAD fixed the sanitizer error.
> 
> Most likely this means that sanitizer is inside the environment in the LD_PRELOAD variable, right?)

yes but that happens only for the pinlock test which recursively calls into ptool, normally it doesn't happen

addmors:
I'm not familiar enough with the code to say anything, but the specific changes specified in the code and the subsequent behavior indicate that it may be setting sanitizer into environment variables.
My suggestion was to run only openssl  under sanitizer, but since that's already being done, it's probably best to find the location where the LD_PRELOAD is set with asan, and if it possible, disable the global LD_PRELOAD. in globaly this variable causes problems like these.

addmors:
Judging by the code, this introduces a messon.

`meson setup builddir -Db_sanitize=address -Dpreload_libasan=/usr/lib/x86_64-linux-gnu/libasan.so.8.0.0`

An alternative would be to stop specifying the path there and make LD_PRELOAD more controlled within scripts.

simo5:
> Judging by the code, this introduces a messon.
> 
> `meson setup builddir -Db_sanitize=address -Dpreload_libasan=/usr/lib/x86_64-linux-gnu/libasan.so.8.0.0`
> 
> An alternative would be to stop specifying the path there and make LD_PRELOAD more controlled within scripts.

You should look at the changes done in #729, not this PR, besides the meson setup is necessary as is. We are currently ok with the changes made in #729 which already minimize the use of asan in all but one test, should there be more changes necessary we'll deal with them at that time.

simo5:
Closing this PR for now, given we have a workaround that does not require a specific version of opensc which makes things easier. We can add the special env var later should the problem resurface.

PR #729, opened 2026-04-27
Attempt to avoid using libasan with pkcs11-tool which now uses DEEPBIND

simo5:
#### Description

Asan can't currently work with RTLD_DEEPBIND flags at dlopen() time, and it let's you know by terminating the process.

We use opensc tools only during setup, and they recently changed to use RTLD_DEEPBIND when opening pkcs11 modules, so we should be able to deal with address sanitizer by simply excluding it during setup.

Token setup via pkcs11-tools uses no pkcs11-provider code so we do not care for running the sanitizer on those actions.

#### Checklist

- [x] This is not a code change
- [x] Test suite changed

#### Reviewer's checklist:

### DIFF
--- a/tests/helpers.sh
+++ b/tests/helpers.sh
@@ -110,10 +110,22 @@ fi
 ptool() {
     # NSS uses the second slot for certificates, so we need to provide the token
     # label in the args to allow pkcs11-tool to find the right slot
-    CMDOPTS=(--module="${P11LIB}" --token-label="${TOKENLABEL}")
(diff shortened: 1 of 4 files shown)


Jakuje (approved):

addmors:
This might seem like a hack, but as a workaround, you can run only the necessary utilities and ignore the rest.

For example:
LD_PRELOAD="sanitizer" openssl ...

PR #730, opened 2026-04-28
Avoid code scan false positives

simo5:
#### Description

This commit should prevent the false positives raise for example on PR #726 
The code is fine but for some reason the scan assumes data can be empty even though the same guards cover both usages ... Change the code around so code scan doesn't trip. this does not change the code behavior.

Also add some consts to make the compiler raise warnings if we modify static structures that should never be touched.

#### Checklist

- [x] Cosmetic change

#### Reviewer's checklist:

### DIFF
--- a/src/sig/ecdsa.c
+++ b/src/sig/ecdsa.c
@@ -82,7 +82,7 @@ struct ecdsa_data {
     { CK_UNAVAILABLE_INFORMATION, 0, NULL, 0 },
 };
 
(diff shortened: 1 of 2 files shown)


Jakuje:
interesting that just moving the code around like this makes the compiler happy ..

Jakuje (approved):

simo5:
> interesting that just moving the code around like this makes the compiler happy ..

The compiler as always happy, its the scanner that is too stupid to look past if () clauses and correlate .. sigh

PR #736, opened 2026-05-26
Make CKA_PUBLIC_KEY_INFO, CKA_EC_POINT, and CKA_ALWAYS_AUTHENTICATE n…

jalex609:
…on-required for better PKCS#11v2.2 compatability

#### Description

Was having errors interoperating with PKCS#11v2.2 HSM without making `allocate` for these attributes false. Wanted to see if this is something that you are interested in changing or if it needs to be left as is.

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/obj/fetch.c
+++ b/src/obj/fetch.c
@@ -10,7 +10,7 @@
 
 #define COMMON_KEYPAIR_ATTRIBUTES \
     COMMON_KEY_ATTRIBUTES, \
-    { { CKA_PUBLIC_KEY_INFO, NULL, 0 }, true, false }
+    { { CKA_PUBLIC_KEY_INFO, NULL, 0 }, false, false }
 /* clang-format on */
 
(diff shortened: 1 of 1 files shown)


simo5:
So, where to start here, first of all you just changed the flag that determines whether we need to allocate a buffer, without allocation, there is no point in requesting any information. The only case where you wouldn't want to allocate on fetching is if you have a pre-allocated buffer that can receive the information (like ULONG values).
 
We also do not care much for the completely obsolete PKCS#11 v2.2, the oldest thing we consider supporting organically is v2.4, but really we will optimize only for v3.0+ tokens.

That said if you need some backwards compatibility then you would have to add detection code that determines that a token is of a certain vintage, and then conditionalize the attributes requested based on that detection.

This current PR only breaks things as is but an unconditional removal would never be appropriate as it would remove features for modern tokens.

PR #738, opened 2026-05-31
obj: Don't let the session-object key cache leak into URI lookups

### LINKED ISSUE #737: pkcs11-provider key cache corrupts concurrent key lookups
**Describe the bug**

`named` loads and uses HSM keys concurrently across zones and worker threads
(zone signing, `keymgr`, dynamic-update re-signing). Once a key is cached, a
concurrent lookup of that key hits defect (1) and the operation fails with
`DST_R_INVALIDPRIVATEKEY` ("private key is invalid"). [Observed in the
`enginepkcs11` system test under parallel CI load](https://gitlab.isc.org/isc-projects/bind9/-/pipelines/330094):

- zone fails to sign — `SOA`/`DNSKEY RRset not signed`;
- two views sharing one policy each create a key — `expected 1 keys, got 2`;
- secure dynamic update fails — `could not get zone keys ... RRSIG/NSEC/NSEC3
  update failed`, returned to the client as `SERVFAIL`;
- `keymgr` cannot read an existing key — `dns_dnssec_findmatchingkeys: error
  reading key file ...: private key is invalid`.

The failures are intermittent (they depend on a cache copy existing during a
concurrent lookup), so the test passes in isolation but flakes under load.

**To Reproduce**
1. Run the enginepkcs11 system test from the `main` branch repeatedly and under the load

**Expected behavior**
When cached, only one key is returned.

**Operating environment:**
 - OS: Debian
 - Version 13 (trixie)

**Token and application used:**
 - Device: SoftHSM2
 - PKCS11 Driver version: 2.6.1
 - Application BIND
 - Version 9.21 / 9.20

**Additional context**

pkcs11-provider's key cache (`pkcs11-module-cache-keys`, **enabled by
default**) makes a session-object copy of each key it uses. The copy keeps
the original key's `CKA_ID`/`CKA_LABEL`, so once a key has been cached it is
returned **twice** by object lookups — once as the real token key and once as
the cache copy. Callers that expect a single match fail intermittently under
concurrent use. With SoftHSM2 this makes BIND 9's `enginepkcs11` system test
flaky.

## The issue

`cache_key()` (src/obj/object.c) copies a token key into a session object to
speed up operations:

```c
C_CopyObject(key, { CKA_TOKEN = false }, &cached);  /* inherits CKA_ID, CKA_LABEL */
```

Two defects follow:

1. **Lookup pollution (primary).** The copy is a separate object that still
   carries the original `CKA_ID`/`CKA_LABEL`. `p11prov_obj_find()` searches by
   `CKA_CLASS`/`CKA_ID`/`CKA_LABEL` with **no `CKA_TOKEN` filter**, so after a
   key is cached, loading it by `pkcs11:` URI returns *both* objects. A caller
   that requires a unique result rejects the key as invalid.

2. **Search disruption (secondary, SoftHSM2).** Creating the copy
   (`C_CopyObject`) concurrently with another thread's `C_FindObjects` can make
   the search return **zero** objects. Setting `pkcs11-module-cache-keys=false`
   avoids both (no copy is ever made).

## Workaround and fix

- **Workaround (BIND 9 CI):** run the PKCS#11 tests with
  `pkcs11-module-cache-keys = false`.
- **Fix (pkcs11-provider):** blank `CKA_ID` and `CKA_LABEL` on the cache copy
  so it is never returned by id/label lookups (the provider refers to the copy
  only by handle). This resolves defect (1). Defect (2) is a deeper SoftHSM2
  concurrency limitation; for SoftHSM2, disabling the key cache remains the
  robust option.

## References

- pkcs11-provider issue #559 (maintainer: "probably better to completely
  disable key caching for SoftHSMv2").

Mno-hime:
cache_key() copies a token key into a session object to speed up operations, passing only { CKA_TOKEN = false } to C_CopyObject, so the copy keeps the original key's CKA_ID and CKA_LABEL.  Being a separate object, it is then returned -- alongside the real key -- by object searches that match on id or label, such as p11prov_obj_find() when a key is loaded by pkcs11: URI.  Callers expecting a single match (e.g. BIND dnssec key loading) then fail intermittently once a key has been cached and is looked up concurrently ("private key is invalid").

Blank CKA_ID and CKA_LABEL on the copy so it is not returned by such lookups; the provider only ever refers to it by handle.

I was able to [verify this fix](https://gitlab.isc.org/isc-projects/bind9/-/pipelines/329930) against BIND 9 in our CI (with an additional BIND 9 fix that is by itself not enough).

#### Reviewer's checklist:

Closes #737


simo5 (commented):
I wonder if i wouldn't be more effective to actually prefix the label with a "cache_" + original label and the CKA_ID, similarly with some kind of prefix (perhaps just the ASCII representation of "cache_" again ...

simo5:
Thinking about this some more I think that copying by default may be too much. Perhaps we should change this confgi option to default top false, and let people turn it on for vry slow tokens that actually need it.

@Jakuje WDYT?

Jakuje:
> Thinking about this some more I think that copying by default may be too much. Perhaps we should change this confgi option to default top false, and let people turn it on for vry slow tokens that actually need it.
> 
> @Jakuje WDYT?

Clearly, this does not make sense on software tokens and CIs, but arguments about network HSMs and smart cards that are slow is something which we started with. I would rather be fast by default than having to answer discussion topics why this was faster before.

If we just disable it by default, these issues will just get hidden by non-default configuration option so I still see a value in having this fixed.

I do not like setting empty ID/labels. I like the idea of prefixing IDs and labels with something.

In any case, I think we should have a regression test for this problem. From the description, it should be quite easy to reproduce within the existing testsuite.

simo5:
@Mno-hime would you be willing to add a simple test that creates a key with a well defined ID label on the token, then searches it twice ? The current code should return two result on the second search and your fix should make it return again only one.

Mno-hime:
@simo5 See 316568717662e887ba6e927f73c5d404e223de90.

src/obj/object.c
```diff
@@ -59,7 +59,17 @@ static void cache_key(P11PROV_OBJ *obj)
 {
     P11PROV_SESSION *session = NULL;
     CK_BBOOL val_false = CK_FALSE;
-    CK_ATTRIBUTE template[] = { { CKA_TOKEN, &val_false, sizeof(val_false) } };
+    /* Cache as a session object, and make it anonymous by blanking CKA_ID and
```
Jakuje: Is there some pre-existing documentation or best practice that blanking ID and LABEL makes the token "anonymous"? I could not find anything in the PKCS#11 specs

Mno-hime: Right, that was too far. See 200d521faee376b571c44491a0c402169aae275d.

simo5:
Can you change the template so that CKA_LABEL is set to something like "Internal Session Only Copy" or similar? I think we should make it possible to internally search these objects w/o confusion. Ideally we should append the original label too, but that is not super-important.

Please also squash in the fixups so I can merge this PR w/o having to squash all commits myself.

Mno-hime:
Fixed, squashed, and pushed.

Jakuje (approved):

simo5:
Thank you for the contribution!

PR #741, opened 2026-06-16
Set timeout for server in ttls test

bukka:
This is useful when server gets stuck as mason kill does not properly invoke the trap and the server output is not visible.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -256,6 +256,7 @@ jobs:
             openssl@3 \
             opensc \
             p11-kit \
+            coreutils \
             gcovr \
             lcov
           if [ "${{ matrix.token }}" = "softokn" ]; then
--- a/tests/ttls
+++ b/tests/ttls
@@ -9,6 +9,15 @@ if [[ "$NSS_FIPS" = "1" ]] && [[ "$TOKENTYPE" = "softokn" ]]; then
     exit 77;
 fi
 
+if command -v timeout >/dev/null 2>&1; then
+    TIMEOUT=timeout
+elif command -v gtimeout >/dev/null 2>&1; then
+    TIMEOUT=gtimeout
+else
+    title "timeout or gtimeout not available"
+    exit 77;
+fi
+
 TEST_ML_DSA=0
 if [ "${SUPPORT_ML_DSA}" -eq 1 ]; then
     if [[ "$TOKENTYPE" = "kryoptic" ]]; then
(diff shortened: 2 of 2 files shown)


simo5 (changes requested):
you need to use the timeout value from the meson testsuite

bukka:
hmm but this is per test timeout (run_test is triggered multiple times) so it just needs to be higher that the `set timeout 10;`. I picked 15 just in case.

bukka:
or does that https://github.com/openssl-projects/pkcs11-provider/actions/runs/27637477267/job/81728434178?pr=741 need higher value to match meson timeout?

simo5:
> hmm but this is per test timeout (run_test is triggered multiple times) so it just needs to be higher that the `set timeout 10;`. I picked 15 just in case.

uhm it may be enough indeed, but then I am not sure why we see test failures

bukka:
OpenSSL master is failing pkcs11-provider TLS test since merging https://github.com/openssl/openssl/pull/31018 . See https://github.com/openssl/openssl/commit/5191045371b4ae1383f0ae1a0f078117e9d9b1c4 for more info.

simo5:
only kryoptic (and I think NSS softokn) can do proper MD duplication, softhsm can't.
Master should not fail with those two, so there is something else going on that needs investigation

simo5:
> only kryoptic (and I think NSS softokn) can do proper MD duplication, softhsm can't. Master should not fail with those two, so there is something else going on that needs investigation

This is what I see in the tls test logs:
kryoptic:
```
## Test setting cert/keys on TLS Context w/o pub key
Failed to set Private Key: error:05800074:x509 certificate routines::key values mismatch (in function ossl_x509_check_private_key in crypto/x509/x509_cmp.c:416): 
```

softokn:
```
## Test setting cert/keys on TLS Context w/o pub key
Failed to set Private Key: error:40800000:pkcs11::pkcs11 (in function p11prov_obj_find_associated in ../src/obj/fetch.c:635): Error in C_FindObjects (count=0)
  caused by: error:40800005:pkcs11::General Error (in function match_public_keys in ../src/obj/keymgmt.c:713): Could not find associated public key object
  caused by: error:40800005:pkcs11::General Error (in function match_key_with_cert in ../src/obj/keymgmt.c:657): Failed to get public attrs from cert
  caused by: error:05800074:x509 certificate routines::key values mismatch (in function ossl_x509_check_private_key in crypto/x509/x509_cmp.c:416): 
```

softhsm:
```
## Test setting cert/keys on TLS Context w/o pub key
Failed to set Private Key: error:05800074:x509 certificate routines::key values mismatch (in function ossl_x509_check_private_key in crypto/x509/x509_cmp.c:416): 
```

simo5:
this could be due to the recent change to always free away keys stored on the session ...

bukka:
Yeah I saw the same failure in openssl when trying master: https://github.com/openssl/openssl/actions/runs/27622082365/job/81673471824?pr=31545 (line 9284). So not related to this PR...

simo5 (approved):

simo5:
Indeed that failure is not related to this PR, so merging this and then will open a bug to investigate what broke with openssl master

Issue #742, opened 2026-06-16
TLS test fail with latest openssl master branch


simo5:
> only kryoptic (and I think NSS softokn) can do proper MD duplication, softhsm can't. Master should not fail with those two, so there is something else going on that needs investigation

This is what I see in the tls test logs:
kryoptic:
```
## Test setting cert/keys on TLS Context w/o pub key
Failed to set Private Key: error:05800074:x509 certificate routines::key values mismatch (in function ossl_x509_check_private_key in crypto/x509/x509_cmp.c:416): 
```

softokn:
```
## Test setting cert/keys on TLS Context w/o pub key
Failed to set Private Key: error:40800000:pkcs11::pkcs11 (in function p11prov_obj_find_associated in ../src/obj/fetch.c:635): Error in C_FindObjects (count=0)
  caused by: error:40800005:pkcs11::General Error (in function match_public_keys in ../src/obj/keymgmt.c:713): Could not find associated public key object
  caused by: error:40800005:pkcs11::General Error (in function match_key_with_cert in ../src/obj/keymgmt.c:657): Failed to get public attrs from cert
  caused by: error:05800074:x509 certificate routines::key values mismatch (in function ossl_x509_check_private_key in crypto/x509/x509_cmp.c:416): 
```

softhsm:
```
## Test setting cert/keys on TLS Context w/o pub key
Failed to set Private Key: error:05800074:x509 certificate routines::key values mismatch (in function ossl_x509_check_private_key in crypto/x509/x509_cmp.c:416): 
```

_Originally posted by @simo5 in https://github.com/openssl-projects/pkcs11-provider/issues/741#issuecomment-4723394925_

bukka:
I just did a bit more checking when this started and it's this PR: https://github.com/openssl/openssl/pull/31018#issuecomment-4728698132

nhorman:
@bukka see commit 5191045371b4ae1383f0ae1a0f078117e9d9b1c4 from #31018 

we knew about the tls test problem here when this was merged, and its not something easily fixable in openssl 

Theres a comment in p11prov_digest_dupctx() in pkcs_provider/src/digests.c about this:
```
/* This is not really funny. OpenSSL by default assumes contexts with
     * operations in flight can be easily duplicated, with all the
     * cryptographic status and then both context can keep going
     * independently. We'll try to save/restore state here. for OpenSSL versions
     * greater than 3.2 this will be a hard failure, while for older OpenSSL
[...]

PR #744, opened 2026-06-18
Bind X509 decoding to library context

simo5:
#### Description

Pre-allocate the X509 object using X509_new_ex with the current library context and a default provider property query before calling d2i_X509. This ensures certificate decoding is properly associated with the specific OpenSSL library context rather than relying on the global default, preventing provider resolution issues.

Related to #742

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -166,7 +166,7 @@ tests = {
   'readkeys': {'suites': all_suites},
   'tls': {'suites': all_suites, 'is_parallel': false, 'timeout': 60},
   'tlsfuzzer': {'suites': all_suites, 'timeout': 90},
-  'uri': {'suites': all_suites, 'timeout': 90},
+  'uri': {'suites': ['softokn', 'kryoptic', 'kryoptic.nss'], 'timeout': 90},
   'ecxc': {'suites': ['softhsm', 'kryoptic', 'kryoptic.nss']},
   'cms': {'suites': ['softokn', 'kryoptic', 'kryoptic.nss']},
   'pinlock': {'suites': ['kryoptic']},
--- a/src/obj/export.c
+++ b/src/obj/export.c
@@ -33,9 +33,15 @@ CK_RV get_attrs_from_cert(P11PROV_OBJ *crt, CK_ATTRIBUTE *attrs, int num)
         return CKR_GENERAL_ERROR;
     }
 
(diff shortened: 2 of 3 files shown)


simo5:
uhmm softhsm2 as usual having some random issue unrelated with the PR ...

bukka:
It looks good! The question is whether there are more such places so you might want to do a review of the codebase... :)

SoftHSMv2 is not tested with OpenSSL master so you might consider disabling it in this test. There is a huge backlog of things to do so getting it working with OpenSSL master is not exactly a priority. We don't currently even test it agains 4.0 (3.5 is the main target atm)...

tests/meson.build
```diff
@@ -166,7 +166,7 @@ tests = {
   'readkeys': {'suites': all_suites},
   'tls': {'suites': all_suites, 'is_parallel': false, 'timeout': 60},
   'tlsfuzzer': {'suites': all_suites, 'timeout': 90},
-  'uri': {'suites': all_suites, 'timeout': 90},
+  'uri': {'suites': ['softokn', 'kryoptic', 'kryoptic.nss'], 'timeout': 90},
```
bukka, line 169: it's broken only with openssl master -  I think better solution would be to completely remove softhsm from all_suites if openssl master is used...

simo5: I am tempted to completely remove SoftHSM2 entirely ... for the past year or so the only  issues found were ... in SoftHSM2 itself ... but until then I will leave it on on most tests.
URI testing is really not about tokens, but more about pkcs11-provider internals, so it is not a big miss if it is not executed for softhsm on any version.

bukka: It should get better. It's just catching up on missing features (PQC and so on) and then the fixes to stabelize things should happen. There's just a big backlog so it might take a bit of time.

simo5: Unfortunately SoftHSM2 has structural problems that will make it always risky to use with providers.

Until SoftHSM2 starts using its own libctx instead of the default context, it will always risk deadlocks and looping into itself when pkcs11-provider is configured, and there isn't much that can be done beyond changing the whole code base to not use the default context.

I suggest people move to use Kryoptic and would welcome SoftHSM2 people to collaborate there instead of duplicating efforts needlessly.

simo5:
Anyway thanks for the review @bukka merging!

PR #746, opened 2026-06-19
fetch: Get parameter set for ML-DSA

flynd:
#### Description

fetch_key() checks the parameter set for ML-DSA keys but didn't actually request it, resulting in param_set always being 0 and the the fetch failing with CKR_KEY_INDIGESTIBLE.

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/obj/fetch.c
+++ b/src/obj/fetch.c
@@ -57,11 +57,13 @@ const struct fetch_attrs EC_private_attrs[] = {
 
 const struct fetch_attrs ML_DSA_public_attrs[] = {
     COMMON_KEYPAIR_ATTRIBUTES,
+    { { CKA_PARAMETER_SET, NULL, 0 }, true, true },
     { { CKA_VALUE, NULL, 0 }, true, true },
 };
 
 const struct fetch_attrs ML_DSA_private_attrs[] = {
     COMMON_KEYPAIR_ATTRIBUTES,
+    { { CKA_PARAMETER_SET, NULL, 0 }, true, true },
     { { CKA_ALWAYS_AUTHENTICATE, NULL, 0 }, true, false },
 };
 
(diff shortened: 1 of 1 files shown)


flynd:
So I guess this wasn't as easy as I thought.
I want to create a ML-DSA-87 signature using an HSM.  It worked fine for RSA, ECDSA, and ED25519 (with the change from PR745) but for ML-DSA I got CKR_KEY_INDIGESTIBLE.  Debugging showed that key.param_set was zero.  I couldn't find any code being run that was setting it, so the obvious fix was to add CKA_PARAMETER_SET to the attributes fetched from the HSM and use that.  After making this change, it works fine.  The key is fetched and a signature can be created.
So what did I break?  Is it just a matter of updating tests or is this code path used for some other use case where key.param_set is already populated?

simo5:
> So what did I break? Is it just a matter of updating tests or is this code path used for some other use case where key.param_set is already populated?

Not your fault, I think tlsfuzzer break is related to fedora rawhide getting rebased to openssl 4.0

I will review your changes

src/obj/fetch.c
```diff
@@ -195,6 +197,13 @@ static CK_RV fetch_key(P11PROV_CTX *ctx, P11PROV_SESSION *session,
         /* decode CKA_EC_PARAMS and store some extra attrs for convenience */
         return pre_process_ec_key_data(key);
     case CKK_ML_DSA:
+        a = p11prov_obj_get_attr(key, CKA_PARAMETER_SET);
```
simo5: given this needs to be done for ML-KEM as well, move it up right after p11prov_move_alloc_attrs and do not fail if it returns null, but only if it has an invalid length. The current code will taker care of failing if it wasn't present.

simo5 (changes requested):
Just one minor change that will allow us to address the fact ML-KEM has the same problem.

simo5 (approved):
LGTM

flynd:
Is there anything else I should do here?  From what I can tell, the failing tests are the same tlsfuzzer as before.

simo5:
> Is there anything else I should do here? From what I can tell, the failing tests are the same tlsfuzzer as before.

I am working to suppress the failures in #747, once that is merged we can rebase this PR on top and merge as soon as CI is green.

simo5:
I meged fixes, if you rebase on latest main CI should go green!

simo5:
thank you!

PR #748, opened 2026-06-23
Allow the user type to be set from configuration

flynd:
#### Description

Thales Luna HSMs define a vendor specific user type called Crypto User with (numerical 0x80000001) that can create signatures but not manage keys.  Make it possible to use such roles by making the user type configurable.

#### Checklist

- [x] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/docs/provider-pkcs11.7.md
+++ b/docs/provider-pkcs11.7.md
@@ -182,6 +182,10 @@ Example:
 Workarounds that may be needed to deal with some tokens and cannot be
 autodetcted yet are not appropriate defaults.
 
+### luna-non-standard-cku-user
+Use the non-standard Crypto User (user type 0x80000001) available in
+Thales Luna HSMs instead of the normal Crypto Officer (CKU_USER).
+
 ### no-deinit
 It prevents de-initing when OpenSSL winds down the provider.
 NOTE this option may leak memory and may cause some modules to
--- a/src/provider.c
+++ b/src/provider.c
@@ -41,6 +41,7 @@ struct p11prov_ctx {
     bool no_deinit;
     bool no_allowed_mechanisms;
     bool no_session_callbacks;
+    CK_USER_TYPE user_type;
     uint64_t blocked_calls;
     bool blocked_ops[OSSL_OP__HIGHEST + 1];
 
@@ -536,6 +537,11 @@ CK_RV p11prov_ctx_status(P11PROV_CTX *ctx)
 CK_UTF8CHAR_PTR p11prov_ctx_pin(P11PROV_CTX *ctx)
 {
     return (CK_UTF8CHAR_PTR)ctx->pin;
+}
+
+CK_USER_TYPE p11prov_ctx_user_type(P11PROV_CTX *ctx)
+{
+    return ctx->user_type;
 }
 
 static void p11prov_ctx_free(P11PROV_CTX *ctx)
@@ -1825,6 +1831,7 @@ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
         return RET_OSSL_ERR;
     }
     ctx->handle = handle;
+    ctx->user_type = CKU_USER;
 
     P11PROV_debug("Starting provider %s %d.%d", PACKAGE_NAME, PACKAGE_MAJOR,
                   PACKAGE_MINOR);
@@ -1982,6 +1989,10 @@ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
             } else if (strncmp(str, "no-session-callbacks", toklen) == 0) {
                 show_quirks = true;
                 ctx->no_session_callbacks = true;
+            } else if (strncmp(str, "luna-non-standard-cku-user", toklen)
+                       == 0) {
+                show_quirks = true;
+                ctx->user_type = CKU_LUNA_CU;
             }
             len -= toklen;
             if (sep) {
@@ -2002,6 +2013,9 @@ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
         }
         if (ctx->no_session_callbacks) {
             P11PROV_debug(" No session callbacks");
+        }
+        if (ctx->user_type != CKU_USER) {
+            P11PROV_debug(" Vendor user type: [%08x]", ctx->user_type);
         }
         if (ctx->blocked_calls) {
             P11PROV_debug(" Blocked calls: [%08lx]", ctx->blocked_calls);
--- a/src/provider.h
+++ b/src/provider.h
@@ -281,6 +281,9 @@ CK_RV p11prov_ctx_set_quirk(P11PROV_CTX *ctx, CK_SLOT_ID id, const char *name,
 CK_RV p11prov_token_sup_attr(P11PROV_CTX *ctx, CK_SLOT_ID id, int action,
                              CK_ATTRIBUTE_TYPE attr, CK_BBOOL *data);
 
+#define CKU_LUNA_CU 0x80000001UL
+CK_USER_TYPE p11prov_ctx_user_type(P11PROV_CTX *ctx);
+
 #define ALLOW_EXPORT_PUBLIC 0
 #define DISALLOW_EXPORT_PUBLIC 1
 int p11prov_ctx_allow_export(P11PROV_CTX *ctx);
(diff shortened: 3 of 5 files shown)


simo5:
Sorry but this approach is not acceptable.

First of all this is not testable, so I won't add a full option for it.
Second of all it seems to be a requirement for Luna HSMs so configuration is not necessary, what you want to do is auto-detection when a module is loaded based on the token info/library name, and enable that behavior only in that case.

flynd:
The Luna HSM has the standard Crypto Officier and it can be used as normal.  It is not a requirement to use the Crypto User role, so I don't think it should be automatically selected for signing operations.

I am a bit surprised that it isn't standard to have these roles separated.  Signing things is a very different task than managing the keys so a person or service that creates signatures shouldn't automatically be allowed to add or modify the keys.

Since I don't know what other token implementations that exist in the world, I assumed that if one has non-standard users, there would likely be others that also has, so making it configurable would allow for more than just the Luna HSM without needing to add specific code for each non-standard case.
Also, since the pin can be specified via the configuration, it seem logical to also specify the user in the same place.

To be clear, I'm not saying that we need have it this way, I'm just explaining my reasoning that led me to make this change this way.

simo5:
No there aren't other tokens that abuse the standard like this as far as I know.

The user types are defined by the standard and adding a vendor type is not a good way to deal with this problem.

For supporting multiple users the standard offers the C_LoginUser interface, where a specific user name can be provided. Then the token can assume all CKU_USER sessions belong to that user instead.

CKU_USER and CKU_CONTEXT_SPECIFIC are absolutely not interchangeable either as CKU_CONTEXT_SPECIFIC identifies an authentication for a specific case, and the semantics between the two types are absolutely different.

If Luna supports C_LoginUser() I can definitely see a way to add that otherwise I think you should open a feature request to them to use the proper interfaces, and in the meanwhile we can have a build time option (--luna-behavior or whatever) and completely segregate this thing out of the normal code paths.

flynd:
Luna does not support C_LoginUser() and it is very unlikely that they would implement it.
This current HSM generation complies to PKCS11 version 2.4 only and they are only adding minimal changes to support newer things like ML-DSA.  The future G8 models are more likely to be version 3 compliant though but that doesn't help us right now.

Since you don't want a configuration option that overrides the user type, would it be more acceptable to allow it in the URI since that also allows it to be paired with the PIN?  E.g. something like this "pkcs11:object=key-18;type=private?user-type=0x80000001&pin-value=12345"
This would allow the user to be selected per use case instead of being fixed in the configuration files or selected automatically in the provider code.

If this is also not okay, then I don't have any further ideas to suggest and will abandon this PR and just use a locally patched pkcs11-provider on our signing server.  Although we can live with that, it is of course preferable to get support in the official versions of the provider so we won't need to have a local build.

simo5:
The URI is a RFC standard and there is no user_type option there either.
I could perhaps accept setting a quirk called something like luna-non-standard-cku-user.

Then we can have a context variable called cku_user that is normally set to CKU_USER, except when the quirk is set, the quirk will change it to the vendored LUNA variable.

Code will compare/pass in ctx->cku_user rather than CKU_USER

That could work. But I want this as a #defin-able away thing with a configuration option, so that we can get rid of it if unwanted, because the way they abused the spec is gross.

flynd:
I have updated the change to use the quirks parameter instead of being a separate configuration parameter.

I haven't added a build option to disable this yet though.  I'm not fluent in meson and there are no other flags to disable functionality in this manner to copy-paste from here so it isn't clear to me how you want this done.  Also, this would require adding a preprocessing step to conditionally include or hide the quirk option from the man page.

docs/provider-pkcs11.7.md
```diff
@@ -182,6 +182,10 @@ Example:
 Workarounds that may be needed to deal with some tokens and cannot be
 autodetcted yet are not appropriate defaults.
 
+### luna-non-standard-cku-user
+Use the non-standard Crypto User (user type 0x80000001) available in
+Thales Luna HSMs instead of the normal Crypto Officier (user type 1).
```
simo5: (user type 1) -> CKU_USER

src/provider.c
```diff
@@ -41,6 +41,7 @@ struct p11prov_ctx {
     bool no_deinit;
     bool no_allowed_mechanisms;
     bool no_session_callbacks;
+    uint32_t cku_user;
```
simo5: this should be of type CK_USER_TYPE / CK_ULONG

src/provider.c
```diff
@@ -1982,6 +1989,10 @@ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
             } else if (strncmp(str, "no-session-callbacks", toklen) == 0) {
                 show_quirks = true;
                 ctx->no_session_callbacks = true;
+            } else if (strncmp(str, "luna-non-standard-cku-user", toklen)
+                       == 0) {
+                show_quirks = true;
+                ctx->cku_user = 0x80000001;
```
simo5: please use something like `#define CKU_THALES_USER_LUNA 0x80000001` in a header.

src/provider.c
```diff
@@ -2003,6 +2014,9 @@ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle, const OSSL_DISPATCH *in,
         if (ctx->no_session_callbacks) {
             P11PROV_debug(" No session callbacks");
         }
+        if (ctx->cku_user != CKU_USER) {
+            P11PROV_debug(" User type: [%08x]", ctx->cku_user);
```
simo5: User type -> Vendor user type

src/provider.h
```diff
@@ -267,6 +267,7 @@ typedef struct p11prov_obj_pool P11PROV_OBJ_POOL;
 /* Provider ctx */
 P11PROV_INTERFACE *p11prov_ctx_get_interface(P11PROV_CTX *ctx);
 CK_UTF8CHAR_PTR p11prov_ctx_pin(P11PROV_CTX *ctx);
+CK_USER_TYPE p11prov_ctx_cku_user(P11PROV_CTX *ctx);
```
simo5: p11prov_ctx_cku_user -> p11prov_ctx_user_type

src/session.c
```diff
@@ -592,7 +592,7 @@ static CK_RV token_login(P11PROV_SESSION *session, P11PROV_URI *uri,
     ret = p11prov_Login(session->provctx, session->session, user_type, pin,
                         pinlen);
     if (ret == CKR_OK) {
-        if (user_type == CKU_USER) {
+        if (user_type != CKU_CONTEXT_SPECIFIC) {
```
simo5: This should be:
```
if user_type != ctx->user_type) {
```

src/provider.c
simo5: and the name should be user_type not cku_user

simo5 (changes requested):
I think this is not invasive enough to require a configure switch after all, but there are nits to be addressed.

src/pkcs11.h
```diff
@@ -1208,6 +1208,9 @@ ULONGDEF(CK_XMSS_PARAMETER_SET_TYPE);
 #define CKU_USER 1UL
 #define CKU_CONTEXT_SPECIFIC 2UL
 
+/* CKU (vendor specific) */
+#define CKU_LUNA_CU 0x80000001UL
+
```
simo5: Sorry this doe not belong here, this file is imported wholesale from the pkcs11-headers project and should be considered read-only

src/provider.c
```diff
@@ -538,6 +539,11 @@ CK_UTF8CHAR_PTR p11prov_ctx_pin(P11PROV_CTX *ctx)
     return (CK_UTF8CHAR_PTR)ctx->pin;
 }
 
+CK_USER_TYPE p11prov_ctx_user_type(P11PROV_CTX *ctx)
+{
+    return (CK_USER_TYPE)ctx->user_type;
```
simo5: the casting is not needed anymore here

simo5 (changes requested):
Pleas move the new define to provider.h and this is good to go.

src/pkcs11.h
flynd: Ah, it wasn't obvious where it would fit in in the other headers.  I hope the location in provider.h I chose is okay.

simo5 (approved):
LGTM

src/pkcs11.h
simo5: If we get one or two more I will move everything into a vendors.h, but currently it is fine.

simo5:
Thank you!

PR #751, opened 2026-07-16
Destroy stored and owned ephemeral keys on free

### LINKED ISSUE #739: Imported public keys leak token session objects
**Describe the bug**

When a public key is stored, a new session object is created (C_CreateObject), but that object is never destroyed when the P11PROV_OBJ is freed. Under workloads that repeatedly import distinct public keys that get later stored (in my case TLS server side client certificate verification), this leaks one session object per unique key on the long-lived login session, which is never closed. 

The freeing logic in p11prov_obj_free when only cache handle is released makes so private keys as the handle should not be destroyed but it doesn't make that much sense for imported objects as far as I see (correct me if I'm missing anything important).

**To Reproduce**

To reproduce, the code needs to store public key object repeatedly. In my case, it can be reproduced with client verification.

**Expected behavior**

The imported object handle should be destroyed - calling C_DestroyObject

 - OS: Linux but shouldn't matter

 - Device: SoftHSMv2 but shouldn't matter

bukka:
#### Description

When an imported public key is used in a token operation, the provider stores it on the token as a session object. The object was never destroyed, so every unique imported public key leaked one object on the long lived login session.

Mark objects stored by the provider and destroy their handle when the object is freed. Objects created from a pool copy borrow the handle and keep a reference on the owning object, so the stored object is only destroyed when the last object using it goes away. Refreshing an imported object now stores the key again instead of searching the token for an object that cannot be found there.

Fixes #739

#### Checklist

#### Reviewer's checklist:


src/obj/object.c
```diff
@@ -147,6 +180,25 @@ static void p11prov_obj_refresh(P11PROV_OBJ *obj)
 
     P11PROV_debug("Refresh object %p", obj);
 
+    if (obj->handle_owner) {
```
simo5: This would require us to know that the owner was refreshed before this object?

src/obj/store.c
```diff
@@ -659,7 +659,12 @@ CK_RV p11prov_obj_copy_key_data(P11PROV_OBJ *dst, P11PROV_OBJ *src)
     }
 
     dst->slotid = src->slotid;
-    dst->handle = src->handle;
+    dst->handle = p11prov_obj_get_handle(src);
+    if (src->handle_owner || src->imported) {
+        /* keep the owner alive as the handle is destroyed on its free */
+        dst->handle_owner = p11prov_obj_ref_no_cache(
+            src->handle_owner ? src->handle_owner : src);
```
simo5: bring this out as a proper if/else, please

src/obj/internal.h
```diff
@@ -23,6 +23,7 @@ struct p11prov_crt {
 struct p11prov_obj {
     P11PROV_CTX *ctx;
     bool raf; /* re-init after fork */
+    bool imported;
```
simo5: I think the proper name for this should be either "ephemeral" or "session_obj", "imported" is ambiguous because I can import and store in the token database.

src/obj/store.c
```diff
@@ -659,7 +659,12 @@ CK_RV p11prov_obj_copy_key_data(P11PROV_OBJ *dst, P11PROV_OBJ *src)
     }
 
     dst->slotid = src->slotid;
-    dst->handle = src->handle;
+    dst->handle = p11prov_obj_get_handle(src);
+    if (src->handle_owner || src->imported) {
```
simo5: Looking at this code TBH I feel like I want to do away with p11prov_obj_copy_key_data() but the incoming key comes from openssl so it is not really possible, sigh ...

That said, "handle_owner" is really not very clear, because what matters here is that you want to keep a reference to the original object so that it is not deleted. So I think this field is probably better called "ref_obj" or maybe "ref_orig_obj" to make it clear this is a reference being hold.

src/obj/internal.h
simo5: Actually now that I think of it, what is the difference between this "imported" bool set to true and cka_token below set to false?

simo5 (commented):
Some questions that should be answered before changing the names.

Overall I understand the intention and makes sense to me, but I wonder if "imported" is needed at all.

src/obj/object.c
bukka: That's not needed because owner will refresh in `p11prov_obj_get_handle(obj->handle_owner)` if its `raf` is set..

src/obj/store.c
bukka: Yeah your name makes more sense. Will change it.

src/obj/internal.h
bukka: Hmm, couldn't this result in attempt to destroy objects that are not imported potentially causing some mix up. In other word aren't there cases where cka_token is false but imported is also false. It looked to me there are few suspicious places where this could happen but will need to verify it properly if you are not sure?

bukka:
Thanks for the review. I was just looking into it again and not sure if the logic is actually right for cached objects. Will need to verify it. Also there might be a race in pool_find that this could make worse so will need to check that properly too.

src/obj/object.c
simo5: but we have no guarantee that function is called before `p11prov_obj_get_handle(obj)` ... do we ?

src/obj/internal.h
simo5: well any object that has cka_token = false, is a session object and therefore ephemeral in nature. What I am trying to understand here is if there is a case where the distinction between:
- generated as a session object by pkcs11-provider
vs
- imported as as session object by ask of openssl

is an important question for this functionality.

If you do not have any reference anymore to an object do we still need to retain it in the session at all if we generated it?

Generally these would be ephemeral symmetric keys that are not "recalled" once the EVP_PKEY object is destroyed, so I can't think of a case where the distinction matters for the purpose of retaining versus destroying the object.

src/obj/object.c
bukka: hmm I'm not sure I follow. The owner gets refreshed the call below so if owner needs refreshing and the "borrower" is called first, then it's like this:
```
get_handle(borrower) -> check if there is owner_handle (this line) -> get_handle(owner) (2 lines below this line) -> raf check -> refresh(owner) -> returned refreshed handle
```

src/obj/internal.h
bukka: I will check this out when I have more time for this and verify it.

src/obj/object.c
simo5: ah sorry I missed that, and I assume that circular references can't be created here so this will just cascade up refresh everything and then cascade back down.

Thanks.

simo5:
@bukka hey Jakub, do you think you'll come back to this soon? I would like to close this problem soonish, so if you have no time to work on it I can take over if you are ok with it.

src/obj/internal.h
bukka: So did a bit more checking and Claude analyse and those are the findings:

imported = true is set in exactly one place — p11prov_obj_store_public_key() (src/obj/store.c:1274) — after the provider itself calls CreateObject to store an OpenSSL-imported public key. So the flag means "this wrapper exclusively owns a session handle the provider created for it". cka_token == false just means "session object", which is a strictly larger set that includes handles the wrapper does not own. Concrete cases where cka_token == false but imported == false, and destroying on free would be wrong:

1. The refresh temp object would destroy the handle it just handed over (src/obj/object.c:259-287). p11prov_obj_refresh() wraps the found handle in tmp via p11prov_obj_from_handle() — which fetches CKA_TOKEN from the token (src/obj/fetch.c:415) — moves tmp->handle into obj, then calls p11prov_obj_free(tmp). If the refreshed object is a session object, !cka_token-based destroy would destroy the very handle obj just adopted. With imported, tmp is safe since from_handle never sets it.
2. Handle borrowers via p11prov_obj_copy_key_data (src/obj/store.c:670). dst copies cka_token from src and borrows the handle (that's what the handle_owner ref this PR adds is for). If src is a session object, dst would get cka_token = false and on free destroy the handle while the owner still uses it — use-after-destroy, then double destroy. You could guard with !obj->handle_owner, but it shows !cka_token alone doesn't encode ownership.
3. Store search results can wrap session objects owned by someone else (src/obj/fetch.c:563). PKCS#11 session objects are visible to all sessions of the same application, so C_FindObjects can return e.g. the volatile cache copies created with CKA_TOKEN=false in cache_key() (src/obj/object.c:102). Those are owned via obj->cached and destroyed by destroy_key_cache(); a found-object wrapper destroying one on free would yank it out from under the live owner — exactly the "mix up" you suspected. Same applies to p11prov_obj_find_associated (src/obj/fetch.c:652).
4. Provider-generated/derived session keys — ECDH derive (src/exchange.c:323), HKDF (src/kdf.c:367, 692), AES gen/import (src/kmgmt/skey.c:254, 612), generic secrets (src/obj/create.c:51). These are cka_token=false, imported=false today and are not destroyed on free (they die with the session). This is the one category where Simo's point holds — destroying them on free would arguably be a leak fix — but it's a behavior change orthogonal to this PR, and cases 1–3 mean it can't be done by simply testing cka_token.


If you think it's all wrong, then feel free to convert it and use just cka_token = false.

src/obj/store.c
bukka: This is done

bukka:
@simo5 Hey I added some changes that I had in mind. They are a bit rusty and done with Claude's help but wanted to spit them out as it's something that I noted before so it doesn't get lost (especially the potential locking issue). I don't have that much time so if you have time to finalise it, that would be awesome! Feel free to just continue here and change whatever you think. Thanks!

src/obj/internal.h
simo5: This analysis looks right, and therefore indeed we can't collapse imported with CKA_TOKEN = false, thanks for spending the time to fully think this through.

Given I want to address point 4, I am thinkg if we should simply change the name from imported to ephemeral anyway, I suspect the solution for 4 will use the same attribute but the name "imported" would be wrong then.

simo5:
@bukka I renamed owns_handle to owns_key, and added two commits to complete support for destroying all create ephemeral keys (ie including symmetric keys created for TLS KDF), then I force pushed to your branch after a rebase to main.

src/obj/object.c
```diff
@@ -55,6 +55,39 @@ static void destroy_key_cache(P11PROV_OBJ *obj, P11PROV_SESSION *session)
...
+    if (obj->handle == CK_INVALID_HANDLE
+        || obj->handle == CK_P11PROV_IMPORTED_HANDLE) {
+        return;
+    }
+
+    ret = p11prov_try_session_ref(obj, CK_UNAVAILABLE_INFORMATION, false, false,
+                                  &session);
+    if (ret != CKR_OK) {
+        P11PROV_debug("Failed to get session to destroy owned key. "
+                      "Error %lx",
```
Jakuje, line 76: this does not fit on a line?
```suggestion
        P11PROV_debug("Failed to get session to destroy owned key. Error %lx",
```

src/obj/store.c
```diff
@@ -1265,14 +1310,41 @@ CK_RV p11prov_obj_store_public_key(P11PROV_OBJ *key)
...
+        /* destroy the session object when key is freed */
+        key->owns_key = true;
     }
 
     return rv;
 }
 
+CK_RV p11prov_obj_re_store_key(P11PROV_OBJ *key)
+{
+    if (key->class != CKO_SECRET_KEY) {
```
bukka: Shouldn't this be?

```suggestion
    if (key->class == CKO_SECRET_KEY) {
```

bukka: Is this for the future use or can it actually be triggered for secret key?

simo5: good catch!
It is for re-creating Key Derivation keys created by the KDF functions in case of fork in the middle of the process, and for ephemerall imported symmetric keys (AES) when the skey stuff is available.

simo5: I wonder if I can modify a test to exercise this part...

simo5: I can and I did, added a commit with a test

simo5:
Added a test that exercise the raf stuff, by importing a public key and keeping all operations forcibly on the token, performing an operation that forces storing the key on the token, then forking which kills the session in the child and requires an object refresh, and repeating verification.

Examination of the logs shows that store_key is called indeed twice (and in one test run I explicitly bork the new p11prov_obj_re_store_key() and verified the test failed.

src/obj/create.c
```diff
@@ -3,79 +3,50 @@
...
+        goto done;
     }
 
-    ret = p11prov_CreateObject(provctx, sess, template, tcount, &key_handle);
-    if (ret != CKR_OK) {
-        return NULL;
-    }
+    /* save data in case we need to refresh on fork */
+    if (session_key) {
+        CK_ATTRIBUTE value = { CKA_VALUE, (void *)secret, secretlen };
```
Jakuje, line 25: This sounds odd. While I understand what we are trying to achieve (have keys surviving forks), I would consider this an unexpected behavior -- user with HSM will give the pkcs11-provider a secret key with the trust it will be stored in HSM, but we secretly store it in memory. I understand that this is smoothing out the misconception of the keys lifetime between 2 different world (openssl and pkcs11), but I believe this should be at least documented somewhere except for this comment in the code (before you will start receiving private security reports disclosing this as a vulnerability).

src/obj/object.c
simo5: check style did not complain so I didn't either :)

I have a followup cleanup patchset (no MR yet) that will change this anyway, so I'll leave as is if that's ok.

Jakuje (approved):
aside the inline comments it looks good.

src/obj/create.c
simo5: These are stored in a session object, not in the HSM (token object), and are ephemeral keys (they will be deleted anyway), and we already had the key to start with, so there is no "additional" exposure, and they do handle a real case (see the test).

simo5:
Thanks, I think I will merge then!

PR #755, opened 2026-07-30
mldsa: Add digest_sign and digest_verify dispatch

### LINKED ISSUE #716: ml-dsa signing
**Describe the feature**
ml-dsa sign operation fail with softhsm

**Expected behavior**
Sign operation to succeed 

**Additional context**
Tested in PKIX-SSH MLDSA-DEMO branch.

simo5:
#### Description

These are just alises of sign and verify, openssl code does the same, as digesting is intrinsic in mldsa.

Fixes #716 

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:


Jakuje:
Do we have some test that could hit these code paths (regardless of being aliases)? Is the @petrovr implementation using some speicific API we do not have test coverage for?

Jakuje (approved):
anyway approved. Code is straightforward. But I am curious what API was used in the failing case and if this could be an issue for other mechanisms.

simo5:
> anyway approved. Code is straightforward. But I am curious what API was used in the failing case and if this could be an issue for other mechanisms.

I think what happens here is that if you are calling EVP_DigestSign, the code will attempt to use DigestSign from the provider, and fallback to EVP_DigestSignInit/Update/Final if that function is not provided.

While if the application explicitly uses EVP_DigestSignInit/Update/Final then those functions are used.

Should we add a small test application that uses one and the other approach to check that both work? (pretty sure both APIs must be used though tests but we can make it very explicit).

petrovr:
Hello,

На 31.07.26 г. в 15:31, Simo Sorce написа:
> simo5 left a comment (openssl-projects/pkcs11-provider#755)
> 
>> anyway approved. Code is straightforward. But I am curious what API was used in the failing case and if this could be an issue for other mechanisms.
> 
> I think what happens here is that if you are calling EVP_DigestSign, the code will attempt to use DigestSign from the provider, and fallback to EVP_DigestSignInit/Update/Final if that function is not provided.
> 
> While if the application explicitly uses EVP_DigestSignInit/Update/Final then those functions are used.
> 
> Should we add a small test application that uses one and the other approach to check that both work? (pretty sure both APIs must be used though tests but we can make it very explicit).
> 

For ML-DSA we could use two models:

- full: OpenSSL 3.4 SIGNATURE and "message" set of functions. This model allows to set "context" if needed. My case is with the default - empty "context-string".
Remark: I cannot get this model to work with current provider and module code. Also I do not have time to review logs.

- "empty context" : This is the case when is used EVP_DigestSignInit with NULL digest. This model match functional implementation of Ed25519.
Remark: Applicable for Ed448 as well.


The request for the second case.

Regards
Roumen Petrov

Jakuje:
But my understanding is that we have test coverage for the second case our testsuite (with kryoptic). We do `EVP_DigestSignInit_ex()`  and `EVP_DigestSign()` (oneshot) in here for all ml-dsa variants:

https://github.com/openssl-projects/pkcs11-provider/blob/main/tests/tpkey.c#L36

What I am worried is that the SoftHSM does not implement some of the PKCS#11 API which makes the code path execution different, but I think its ok as it is given that we know it works somehow.

simo5:
we do not test ML-DSA with SoftHSM2 so we won't know until e we require a version that supports ML-DSA and update the setup to allow it.
Underneath if SoftHSM2 does not support C_SignUpdate with ML-DSA then it will indeed fail when Init/Update/Final paths are used, but at least by adding explicit one-shot Sign it has a chance of working in that case.

petrovr:
Hello,

На 3.08.26 г. в 16:09, Simo Sorce написа:
> simo5 left a comment (openssl-projects/pkcs11-provider#755)
> 
> we do not test ML-DSA with SoftHSM2 so we won't know until e we require a version that supports ML-DSA and update the setup to allow it.
> Underneath if SoftHSM2 does not support C_SignUpdate with ML-DSA then it will indeed fail when Init/Update/Final paths are used, but at least by adding explicit one-shot Sign it has a chance of working in that case.
> 


If I remember PKCS#11 provider test plans it is based on assumption related to OS vendor package release. As exception availability of some tests is hard coded in test suite.


My test cases does not run by default provider related tests. The reasons - it is not possible to find conditions a test to pass or fail. The test could be run explicitly and test code scans "module mechanisms'. In the same time test code allows override from environment.
By example:
.....
# set to true if pkcs#11 module support ML-DSA keys
mldsa_mechanism=false
.....
if fgrep 'ML-DSA ...' ... <module-mechanisms-list> ... ; then
   mldsa_mechanism=${MLDSA_MECHANISM-:}
fi
.....

Sample above allows in development (or CI) environment to deactivate test even if module support functionality.


In context of pkcs#11 provider

$ fgrep -ri mldsa tests/
.....
tests/meson.build:  'mldsa': {'suites': ['kryoptic']},
.....

$ fgrep -ri ml-dsa tests/
.....

$ fgrep -ri ml_dsa tests/
.....
tests/softokn-init.sh:export SUPPORT_ML_DSA=0
.....

Tests are not activated for softhsm and other modules. The ML-DSA test with softhsm is explicitly disabled.


I would like to see some improvements in test code that allows tests to be activated depending of "current" status of packages.


Regards,
Roumen

simo5:
> I would like to see some improvements in test code that allows tests to be activated depending of "current" status of packages.

We are always open to contributions in that sense.

bukka:
> Underneath if SoftHSM2 does not support C_SignUpdate with ML-DSA then it will indeed fail when Init/Update/Final

FWIW SoftHSMv2 supports multipart ML-DSA since https://github.com/softhsm/SoftHSMv2/commit/c274be21c08a0db21023aaa3028bb47985ac2417 (just master branch like all PQC support there)

Jakuje:
> FWIW SoftHSMv2 supports multipart ML-DSA since [softhsm/SoftHSMv2@c274be2](https://github.com/softhsm/SoftHSMv2/commit/c274be21c08a0db21023aaa3028bb47985ac2417) (just master branch like all PQC support there)

Do you have some timeline when the new softhsm might land? There is quite a lot of bugfixes (and features accumulated), but I see also still quite some work in progress...

PR #756, opened 2026-08-03
Add SLH DSA support

### LINKED ISSUE #701: SLH-DSA and LMS support
**Describe the feature**
[PKCS#11v3.2](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/pkcs11-spec-v3.2.html#_Toc195693759) has support for both SLH-DSA (a.k.a [FIPS 205](https://csrc.nist.gov/pubs/fips/205/final)) as well as LMS/HSS [RFC8554](https://www.rfc-editor.org/info/rfc8554).

OpenSSL also supports both of these via the provider interface (see [oqs-provider](https://github.com/open-quantum-safe/oqs-provider/tree/main))

Being able to use SLH-DSA or LMS provided by a PKCS11 library would be great.

**Expected behavior**
SLH-DSA and LMS based private keys residing in a PKCS11 accessible interface are accessible in OpenSSL via the pkcs11-provider in exactly the same way that e.g. ML-DSA based keys are.

**Additional context**
none

simo5:
#### Description

Adds support for SLH-DSA key objects, key generations and signatures using SLH-DSA

Fixes #701 

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:


src/kmgmt/slhdsa.c
```diff
@@ -0,0 +1,372 @@
+/* Copyright (C) 2025 Simo Sorce <simo@redhat.com>
```
Jakuje:
```suggestion
/* Copyright (C) 2026 Simo Sorce <simo@redhat.com>
```

src/kmgmt/slhdsa.c
```diff
@@ -0,0 +1,372 @@
...
+static void *p11prov_slhdsa_gen(void *genctx, OSSL_CALLBACK *cb_fn,
+                                void *cb_arg)
+{
+    struct key_generator *ctx = (struct key_generator *)genctx;
+    void *key;
+    CK_RV ret;
+
+#define SLHDSA_PUBKEY_TMPL_SIZE 3
+    CK_ATTRIBUTE pubkey_template[SLHDSA_PUBKEY_TMPL_SIZE + COMMON_TMPL_SIZE] = {
+        { CKA_TOKEN, DISCARD_CONST(&val_true), sizeof(CK_BBOOL) },
```
Jakuje, line 110: this is a topic for other PR, but would it make sense to move the `CKA_TOKEN` to the part of the common template given that its true for all key types key generation?

src/kmgmt/slhdsa.c
```diff
@@ -0,0 +1,372 @@
...
+        case CKP_SLH_DSA_SHAKE_192F:
+            sigsize = SLH_DSA_SHA2_192F_SIG_SIZE;
+            break;
+        case CKP_SLH_DSA_SHA2_256S:
+        case CKP_SLH_DSA_SHAKE_256S:
+            sigsize = SLH_DSA_SHA2_256S_SIG_SIZE;
+            break;
+        case CKP_SLH_DSA_SHA2_256F:
+        case CKP_SLH_DSA_SHAKE_256F:
+            sigsize = 49856;
```
Jakuje: there is no
```suggestion
            sigsize = SLH_DSA_SHA2_256F_SIG_SIZE;
```

src/obj/fetch.c
```diff
@@ -70,6 +70,9 @@ const struct fetch_attrs ML_DSA_private_attrs[] = {
 #define ML_KEM_public_attrs ML_DSA_public_attrs
 #define ML_KEM_private_attrs ML_DSA_private_attrs
 
+#define SLH_DSA_public_attrs ML_DSA_public_attrs
+#define SLH_DSA_private_attrs ML_DSA_private_attrs
```
Jakuje: Could we rename these rather than  making a define. I guess it makes it more confusing than needed

src/obj/object.c
```diff
@@ -486,6 +486,42 @@ CK_ATTRIBUTE *p11prov_obj_get_attr(P11PROV_OBJ *obj, CK_ATTRIBUTE_TYPE type)
...
+        return NULL;
+    }
+
+    for (int i = 0; i < obj->numattrs; i++) {
+        if (obj->attrs[i].type == type) {
+            return &obj->attrs[i];
+        }
+    }
+
+    /* search public key if the private key does not have it */
```
Jakuje, line 502: AFAIK this comment is a bit confusing. We search public key because private key has the VALUE, but its the private key value (which is likely not exportable anyway.

src/sig/mldsa.c
```diff
@@ -31,8 +31,14 @@ DISPATCH_MLDSA_FN(settable_ctx_params);
 static CK_RV p11prov_mldsa_set_mechanism(P11PROV_SIG_CTX *sigctx)
 {
     sigctx->mechanism.mechanism = CKM_ML_DSA;
-    sigctx->mechanism.pParameter = NULL;
-    sigctx->mechanism.ulParameterLen = 0;
+    if (sigctx->additional_context.pContext
+        || sigctx->additional_context.hedgeVariant != CKH_HEDGE_PREFERRED) {
```
Jakuje, line 37: its not clear to me how the hedge variant comes into the play here. I assume its explained in the commit message. But code comment might make it more clear to future readers.

src/sig/slhdsa.c
```diff
@@ -0,0 +1,742 @@
...
+static void *p11prov_slhdsa_sha2_256f_newctx(void *provctx,
+                                             const char *properties)
+{
+    return p11prov_slhdsa_newctx(provctx, properties, CKP_SLH_DSA_SHA2_256F);
+}
+static void *p11prov_slhdsa_shake_256f_newctx(void *provctx,
+                                              const char *properties)
+{
+    return p11prov_slhdsa_newctx(provctx, properties, CKP_SLH_DSA_SHAKE_256F);
+}
```
Jakuje: could we make this boilerplate with macros similarly as in the kmgmt?

src/sig/slhdsa.c
```diff
@@ -0,0 +1,742 @@
...
+
+static const unsigned char der_slh_dsa_sha2_128s_alg_id[] = {
+    DER_SEQUENCE,     DER_NIST_SIGALGS_LEN + 3,
+    DER_OBJECT,       DER_NIST_SIGALGS_LEN + 1,
+    DER_NIST_SIGALGS, 0x14
+};
+static const unsigned char der_slh_dsa_shake_128s_alg_id[] = {
+    DER_SEQUENCE,     DER_NIST_SIGALGS_LEN + 3,
+    DER_OBJECT,       DER_NIST_SIGALGS_LEN + 1,
+    DER_NIST_SIGALGS, 0x15
```
Jakuje, line 428: Per RFC 9814, this is wrong. The SHAKE 128s Algorithm ID is `26` (`0x1A`). The list is sorted by hash first, not by security bits. All the following OIDs are wrong too (except for the last one).
https://datatracker.ietf.org/doc/rfc9814/

src/provider.c
```diff
@@ -871,7 +871,8 @@ static CK_RV alg_set_op(OSSL_ALGORITHM **op, int idx, OSSL_ALGORITHM *alg)
         CKM_ECDSA_SHA3_256, CKM_ECDSA_SHA3_384, CKM_ECDSA_SHA3_512
 
 #define PQC_MECHS \
-    CKM_ML_DSA, CKM_ML_DSA_KEY_PAIR_GEN, CKM_ML_KEM, CKM_ML_KEM_KEY_PAIR_GEN
+    CKM_ML_DSA, CKM_ML_DSA_KEY_PAIR_GEN, CKM_ML_KEM, CKM_ML_KEM_KEY_PAIR_GEN, \
+        CKM_SLH_DSA, CKM_SLH_DSA_KEY_PAIR_GEN
```
Jakuje, line 875: we have also the `CKM_HASH_SLH_DSA` and  `CKM_HASH_SLH_DSA_<hash>` mechanisms. Should they be reflected somehow in this list or somewhere in this file?

src/provider.c
```diff
@@ -1448,6 +1477,66 @@ static CK_RV static_operations_init(P11PROV_CTX *ctx)
...
+    ADD_ALGO_EXT(SLH_DSA_SHAKE_256S, encoder, DEFAULT_PROPERTY(",output=text"),
+                 p11prov_slhdsa_encoder_text_functions);
+    ADD_ALGO_EXT(SLH_DSA_SHAKE_256S, encoder,
+                 DEFAULT_PROPERTY(",output=der,structure=SubjectPublicKeyInfo"),
+                 p11prov_slhdsa_encoder_spki_der_functions);
+    ADD_ALGO_EXT(SLH_DSA_SHAKE_256F, encoder, DEFAULT_PROPERTY(",output=text"),
+                 p11prov_slhdsa_encoder_text_functions);
+    ADD_ALGO_EXT(SLH_DSA_SHAKE_256F, encoder,
+                 DEFAULT_PROPERTY(",output=der,structure=SubjectPublicKeyInfo"),
+                 p11prov_slhdsa_encoder_spki_der_functions);
```
Jakuje, line 1539: I see this is already macro, but I wonder if wrapping this in another one would make this boilerplate more readable. Ideally really by parametrize it with something like having a shared macro with the list of parameters that could be iterated over somehow.

tests/setup.sh
```diff
@@ -642,6 +642,42 @@ if [ "$SUPPORT_ML_KEM" -eq 1 ]; then
...
 fi
 
+if [ "$SUPPORT_SLH_DSA" -eq 1 ]; then
+    title PARA "generate SLH-DSA Key pair"
+    get_next_keyid
+    TSTCRTN="slhDsa"
+
+    # not supported by the pkcs11-tool yet. Do it for now with OpenSSL CLI
+    # ptool --keypairgen --key-type="SLH-DSA-SHAKE-128S" --id="$KEYID" \
+    #       --label="${TSTCRTN}" 2>&1
```
Jakuje, line 652: Actually this should already work in current opensc. But I am not sure if it made it already to all our container images.

tests/tslhdsa
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash -e
+# Copyright (C) 2025 Simo Sorce <simo@redhat.com>
```
Jakuje:
```suggestion
# Copyright (C) 2026 Simo Sorce <simo@redhat.com>
```

tests/tslhdsa
```diff
@@ -0,0 +1,109 @@
...
+        -out ${TMPPDIR}/slh-dsa-shake-128s.sig'
+
+title LINE "Verify signature with SLH-DSA-SHAKE-128s"
+ossl '
+pkeyutl -verify
+        -inkey "pkcs11:type=public;object=Test-SLH-DSA-SHAKE-128s" -pubin
+        -in ${RAND64FILE} -rawin
+        -sigfile ${TMPPDIR}/slh-dsa-shake-128s.sig'
+
+title LINE "Sign random data with SLH-DSA-SHAKW-128s using deterministic signature"
```
Jakuje:
```suggestion
title LINE "Sign random data with SLH-DSA-SHAKE-128s using deterministic signature"
```

src/provider.c
simo5: Not yet, we also have CKM_HASH_ML_DSA and do not implement that either.
I am not sue OpenSSL supports prehashing (it didn't when ML/SLH-DSA were introduced), but even if it does it would be only in master (I think) so there is time to properly consider those. (they should be rarely used, most IETF standards do not want to deal with pre-hashing the way NIST specified it).

src/provider.c
simo5: maybe open a bug to reign in? I think we should probably move the registration into the individual mechanism files with a hook to be called from the main registration function.

tests/setup.sh
simo5: well not having to depend opensc catching up is not a bad thing, but we can deal with this another time (perhaps by just removing these comments from both mldsa and slhdsa)

src/kmgmt/slhdsa.c
simo5: maybe, I think we could have an auxiliary function used for all key generation indeed, but, for another time

src/kmgmt/slhdsa.c
simo5: ugh ... shakes fist at gemini :-)

src/obj/fetch.c
simo5: ok

simo5: The define will need to remain because it is used in the FILL_ATTRS macro, but I can use a generic structure for them all

src/obj/object.c
simo5: the private key will never have CKA_VALUE because it will never be returned by the token, so when you try to look it up it will fail for MLDSA/ML-KEK/SLH_DSA, but for example RSA has the MODULES and public exponent also on the private key, so they will be returned just fine. And EC keys have artificial attributes we generate on fetch.

Finally generally after key generation the public key is already associated to the private key so not real search happens. The comment basically warns what the "worst" case can be, and that is why it says "if the private keys does not have it".

src/sig/mldsa.c
simo5: CKH_HEDGE_PREFERRED is the default by spec, so we care to explicitly set the additional context only if it is not set to default (or it introduces a different context than default (empty)

src/sig/slhdsa.c
simo5: ok

src/sig/slhdsa.c
simo5: thanks gemini!
but most importantly thank @Jakuje this could have gone undetected easily

Jakuje: follow-up: This might be a good reason to have extensive test of all parameters, even if it would not run by default and it would be somehow marked as "slow" just to verify the correctness of all the parameteres in the vast amount of tables we have here.

tests/setup.sh
Jakuje: Created #758.

src/provider.c
Jakuje: Created #759.

Jakuje (approved):

simo5:
Thanks for the reviews!

PR #483, opened 2024-11-26
Add support for CKA_EC_POINT as not DER encoded

simo5:
#### Description

Gracefully decode CKA_EC_POINT (for Edwards curves) whether it is DER encoded or not.

Use Kryoptic compatibility mode only during the setup phase to deal with certtool incompatibility with non-DER encoded CKA_EC_POINTs

Fixes #478

#### Checklist

- [x] Code modified for feature
- ~[ ] Test suite updated with functionality tests~
- ~[ ] Test suite updated with negative tests~
- ~[ ] Documentation updated~

#### Reviewer's checklist:

- [ ] Any issues marked for closing are addressed
- [ ] There is a test suite reasonably covering new functionality or modifications
- [ ] This feature/change has adequate documentation added
- [ ] Code conform to coding style that today cannot yet be enforced via the check style test
- [ ] Commits have short titles and sensible commit messages
- [ ] Coverity Scan has run if needed (code PR) and no new defects were found

Issue #478, opened 2024-11-25
Support for CKA_EC_POINT multiple encodings


simo5:
**Describe the bug**
The latest PKCS#11 Spec define CKA_EC_POINT for CKA_EC_EDWARDS and CKA_EC_MONTGOMERY as:
"Public key bytes in little endian order as defined in RFC 8032"

That is just the raw bytes, like for the respective CKA_VALUE in the private keys.

However historically the encoding of CKA_EC_POINT has been problematic.
The 3.0 Spec had defined: "DER-encoding of the b-bit public key value in little endian order as defined in RFC 8032" which means DER encoding using BIT STRING type.

And yet the SoftHSM2 implementation instead surfaces CKA_EC_POINT as DER encoding with OCTET STRING.

For maximum compatibility with tokens pkcs11-provider should support CKA_EC_POINT encoded in all three representation for reading:
Pure byte string
DER encoded OCTET STRING
DER encoded BIT STRING

As for creating public keys we may need to retry or add a tweak option if some token support only a specific encoding in input.

PR #767, opened 2026-08-11
Fix ambiguous decoding of raw Edwards and Montgomery EC points

### LINKED ISSUE #766: Raw ECX CKA_EC_POINT values can be misparsed as DER OCTET STRINGs, causing intermittent TLS failures
## Problem

PKCS#11 3.1 and later specify `CKA_EC_POINT` as raw bytes for Edwards and Montgomery keys. However, `decode_ec_point()` currently calls `d2i_ASN1_OCTET_STRING()` first for every EC key and falls back to the raw attribute only when ASN.1 decoding fails.

This is a follow-up to #478 and its implementation in #483. Supporting both the specification's raw encoding and legacy DER OCTET STRING-wrapped encodings is useful, but selecting between them by trying DER first is ambiguous for arbitrary public-key bytes.

The current PKCS#11 3.2 standard specifies the raw encodings in the Edwards and Montgomery public-key object sections:
https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/os/pkcs11-spec-v3.2-os.html

A raw ECX public key is arbitrary binary data and can accidentally begin with a valid DER OCTET STRING header. In that case ASN.1 decoding succeeds and the provider treats the decoded contents as the public key, potentially discarding part of the raw attribute.

For example, the following generated 32-byte raw X25519 public value is interpreted as a 30-byte DER OCTET STRING:

```text
04 1e f8 1d 4b f2 30 d8 a8 31 9a 07 50 62 bb eb
66 6b e7 3c 7c ac 92 05 43 d8 58 01 95 a9 a6 35
```

The current code neither checks that the decoded length is the expected curve size nor, for shorter DER prefixes, that the decoder consumed the complete attribute. A malformed/truncated public key can therefore be exported as a TLS key share and fail later during key derivation.

## Observed symptom

OpenSSL's `external-tests-pkcs11-provider` job intermittently failed in the TLS test with Kryoptic after negotiating X25519:

```text
error:0680009B:asn1 encoding routines:ASN1_get_object:too long
error:06800066:asn1 encoding routines:asn1_check_tlen:bad object header
error:0688010A:asn1 encoding routines:asn1_d2i_ex_primitive:nested asn1 error
error:0A00006C:SSL routines:ssl_derive:bad key share
```

Failing CI job:
https://github.com/openssl/openssl/actions/runs/31406926081/job/93515424162

The same OpenSSL PR head, 3a668f716fa8c3a08d8e89a212511a9682b18c17, later passed the same job:
https://github.com/openssl/openssl/actions/runs/31497156507/job/93797655305

The pass and failure at an identical PR head are consistent with a data-dependent failure involving a randomly generated ECX public value.

The failing CI log doesn't retain the raw `CKA_EC_POINT`, so it can't prove which public value triggered that particular failure. The `04 1e ...` value above is a deterministic reproducer for the underlying decoding defect.

## Expected behavior

For `CKK_EC_MONTGOMERY` and `CKK_EC_EDWARDS` keys:

- An attribute whose size exactly matches the selected curve's public-key size must be treated as raw data without attempting ASN.1 decoding.
- A DER OCTET STRING-wrapped value should remain accepted for compatibility with older or non-conforming modules, but only if the decoder consumes the complete attribute and the decoded value has the expected curve size.
- Values with trailing data or an incorrect decoded length should be rejected.

Traditional `CKK_EC` point decoding should remain unchanged.

## Impact

This is primarily an intermittent availability and interoperability problem.
For X25519, a short-form DER collision occurs whenever a raw value starts with `04` followed by a plausible length byte. It can surface as a sporadic TLS 1.3 `bad key share` failure and is difficult to reproduce from normal randomized test runs.

## Affected version

The issue is present in pkcs11-provider 1.2.0 at least through commit 088393bbc7e630999955c0771a28be60d79393ad.

idrassi:
Fixes #766.

#483, addressing #478, added support for raw `CKA_EC_POINT` values alongside the existing DER OCTET STRING decoding. The decoder currently tries the DER form first and uses raw bytes only when that fails. That order is ambiguous because an arbitrary raw Edwards or Montgomery public key can start with a valid DER OCTET STRING header.

For example, a 32-byte raw X25519 value beginning with `04 1e` is accepted as a DER OCTET STRING containing only 30 bytes. The truncated value can later be exported as a TLS key share and cause an intermittent `bad key share` failure.

Make EC point decoding curve-length-aware:

- Treat an Edwards or Montgomery `CKA_EC_POINT` whose length matches the selected curve as raw, without probing it as ASN.1.
- Continue accepting legacy DER OCTET STRING-wrapped ECX points, but require the decoder to consume the complete attribute and require the decoded point to have the expected curve size.
- Leave existing traditional `CKK_EC` point decoding behavior unchanged.
- Keep the ASN.1 error-stack handling around the decoder so rejected probes don't contaminate later OpenSSL operations.

The decoding logic is isolated in a small internal helper so it can be tested without relying on randomized token key generation. The regression test uses deterministically generated 32-byte `04 1e ...` values and verifies raw X25519 and Ed25519 handling, legacy DER OCTET STRING compatibility, rejection of trailing and wrong-length wrapped values, and traditional EC behavior.

#### Testing

- Built the provider and unit test against OpenSSL 3.0.13.
- Built the provider and unit test against OpenSSL 4.1.0-dev.
- Ran the token-independent `ec-point-decode` unit-test instance under the `softokn`, `softhsm`, `kryoptic`, and `kryoptic.nss` Meson suite labels against both OpenSSL versions.
- Ran all four suite-tagged unit-test instances under Valgrind against both OpenSSL versions.
- Ran `make check-style` and `clang-format --dry-run --Werror` on the new sources.

#### Checklist

- [x] Code modified for bug fix

#### Reviewer's checklist:

### DIFF
--- a/src/obj/ec_point.c
+++ b/src/obj/ec_point.c
@@ -1,0 +1,94 @@
+/* Copyright (C) 2026 Mounir IDRASSI
+   SPDX-License-Identifier: Apache-2.0 */
+
+#include <limits.h>
+#include <stdbool.h>
+#include <openssl/asn1.h>
+#include <openssl/crypto.h>
+
+#include "ec_point.h"
+
+CK_RV p11prov_decode_ec_point_value(CK_KEY_TYPE key_type,
+                                    CK_ULONG expected_size,
+                                    const CK_ATTRIBUTE *attr,
+                                    unsigned char **decoded,
+                                    size_t *decoded_len)
+{
+    ASN1_OCTET_STRING *octet = NULL;
+    const unsigned char *value;
+    const unsigned char *value_end;
+    const unsigned char *point = NULL;
+    size_t point_len = 0;
+    int asn1_len;
+    bool is_ecx;
+    CK_RV ret = CKR_KEY_INDIGESTIBLE;
+
+    if (attr == NULL || decoded == NULL || decoded_len == NULL) {
+        return CKR_ARGUMENTS_BAD;
+    }
+
+    *decoded = NULL;
+    *decoded_len = 0;
+
+    is_ecx = key_type == CKK_EC_EDWARDS || key_type == CKK_EC_MONTGOMERY;
+    if (key_type != CKK_EC && !is_ecx) {
+        return CKR_KEY_TYPE_INCONSISTENT;
+    }
+    if (is_ecx
+        && (expected_size == 0
+            || expected_size == CK_UNAVAILABLE_INFORMATION)) {
+        return CKR_KEY_INDIGESTIBLE;
+    }
+
+    if (attr->pValue == NULL || attr->ulValueLen == 0
+        || attr->ulValueLen > (CK_ULONG)LONG_MAX) {
+        return CKR_KEY_INDIGESTIBLE;
+    }
+
+    /*
+     * PKCS#11 3.1 and later specify raw CKA_EC_POINT values for Edwards and
+     * Montgomery keys. Check the expected raw size before trying DER because
+     * arbitrary public-key bytes can accidentally form a valid DER prefix.
+     */
+    if (is_ecx && attr->ulValueLen == expected_size) {
+        point = attr->pValue;
+        point_len = attr->ulValueLen;
+        goto done;
+    }
+
+    value = attr->pValue;
+    value_end = value + attr->ulValueLen;
+    octet = d2i_ASN1_OCTET_STRING(NULL, &value, (long)attr->ulValueLen);
+    if (octet == NULL || (is_ecx && value != value_end)) {
+        goto done;
+    }
+
+    asn1_len = ASN1_STRING_length(octet);
+    if (asn1_len <= 0) {
+        goto done;
+    }
+    point_len = (size_t)asn1_len;
+
+    /*
+     * Retain compatibility with modules that DER OCTET STRING-wrap ECX
+     * points, but reject a decoded value that does not match the selected
+     * curve.
+     */
+    if (is_ecx && point_len != expected_size) {
+        goto done;
+    }
+    point = ASN1_STRING_get0_data(octet);
+
+done:
+    if (point != NULL) {
+        *decoded = OPENSSL_memdup(point, point_len);
+        if (*decoded == NULL) {
+            ret = CKR_HOST_MEMORY;
+        } else {
+            *decoded_len = point_len;
+            ret = CKR_OK;
+        }
+    }
+    ASN1_OCTET_STRING_free(octet);
+    return ret;
+}
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -123,8 +123,21 @@ foreach t, sources : test_programs
   test_executables += [t]
 endforeach
 
+ec_point_test = executable(
+  'tecpoint',
+  ['tecpoint.c', '../src/obj/ec_point.c'],
+  build_by_default: false,
+  include_directories: [configinc, include_directories('../src')],
+  dependencies: [libcrypto],
+)
+
 setup_script=find_program('setup.sh')
 all_suites=['softokn', 'softhsm', 'kryoptic', 'kryoptic.nss']
+test(
+  'ec-point-decode',
+  ec_point_test,
+  suite: ['unittests'] + all_suites,
+)
 foreach suite : all_suites + ['kryoptic.multislot']
   test(
     'setup',
--- a/src/meson.build
+++ b/src/meson.build
@@ -19,6 +19,7 @@ pkcs11_provider_sources = [
   'interface.c',
   'registration.c',
   'obj/create.c',
+  'obj/ec_point.c',
   'obj/export.c',
   'obj/keymgmt.c',
   'obj/fetch.c',
--- a/src/obj/ec_point.h
+++ b/src/obj/ec_point.h
@@ -1,0 +1,17 @@
+/* Copyright (C) 2026 Mounir IDRASSI
+   SPDX-License-Identifier: Apache-2.0 */
+
+#ifndef _OBJ_EC_POINT_H_
+#define _OBJ_EC_POINT_H_
+
+#include <stddef.h>
+
+#include "pkcs11.h"
+
+CK_RV p11prov_decode_ec_point_value(CK_KEY_TYPE key_type,
+                                    CK_ULONG expected_size,
+                                    const CK_ATTRIBUTE *attr,
+                                    unsigned char **decoded,
+                                    size_t *decoded_len);
+
+#endif /* _OBJ_EC_POINT_H_ */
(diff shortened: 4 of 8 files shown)


src/obj/ec_point.c
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    value = attr->pValue;
+    value_end = value + attr->ulValueLen;
+    octet = d2i_ASN1_OCTET_STRING(NULL, &value, (long)attr->ulValueLen);
+    if (octet == NULL || (is_ecx && value != value_end)) {
+        goto done;
+    }
+
+#if OPENSSL_VERSION_NUMBER >= 0x40100000L
```
simo5: just always use ASN1_STRING_length for now it is unclear if ASN1_STRING_length_ex will stay there is a PR renaming it to ASN1_STRING_get_length

src/obj/ec_point.c
```diff
@@ -0,0 +1,102 @@
...
+     * PKCS#11 3.1 and later specify raw CKA_EC_POINT values for Edwards and
+     * Montgomery keys. Check the expected raw size before trying DER because
+     * arbitrary public-key bytes can accidentally form a valid DER prefix.
+     */
+    if (is_ecx && expected_size != 0
+        && expected_size != CK_UNAVAILABLE_INFORMATION
+        && attr->ulValueLen == expected_size) {
+        point = attr->pValue;
+        point_len = attr->ulValueLen;
+        goto copy;
```
simo5: we generally do not allow any other goto than goto done.

Initialize point to null and goto done

in "done" duplicate the point if it is not null or return an error

simo5 (changes requested):

simo5:
good PR, just minor style issues

idrassi:
Thanks for the review. I have implemented the requested changes.

src/obj/ec_point.c
```diff
@@ -0,0 +1,95 @@
...
+        point_len = (size_t)asn1_len;
+    }
+
+    /*
+     * Retain compatibility with modules that DER OCTET STRING-wrap ECX
+     * points, but reject a decoded value that does not match the selected
+     * curve.
+     */
+    if (is_ecx
+        && (expected_size == 0 || expected_size == CK_UNAVAILABLE_INFORMATION
```
simo5: This is two checks in one.
if we have ECX keys it seems we will always fail if expected_size is 0 or CK_UNAVAILABLE_INFORMATION
then the second test is if point_len is equal to expected_size

The first check should be probably moved to the top, there is no point doing any works if we are always going to reject any result.

As for the second check, we had to recently change some code to accept shorter octet strings for some of the keys because the first octet was zero and openssl would return a normalized BN in that case (ie shorter data), I wonder if this check may cause issues if someone encoded a shorter DER struct from an openssl reduced point ?

simo5: Feels to me that regardless of is_ecx we should fail `if (point_len > expected_len) {` ?

simo5 (changes requested):

src/obj/ec_point.c
idrassi: Good point about the unavailable size check. I moved it up so ECX keys fail before any ASN.1 parsing, and simplified the two later conditions.

I kept the exact length check for decoded ECX values. This path carries public keys as octet strings, and OpenSSL expects the fixed size for each ECX key type. Passing a shorter value would only move the failure further down the path.

I didn't apply `point_len > expected_size` to `CKK_EC`, since `expected_size` there is the order size in bytes rather than the encoded point size. For a 256 bit curve it is 32, while a valid encoded point can be 33 or 65 bytes.

The BN normalization case sounds different from this path. If you had a particular earlier change in mind, could you point me to it? I may be missing a compatibility case worth handling separately.

simo5: no need for more changes, the way you did it looks good!

simo5 (dismissed):
LGTM

src/obj/ec_point.c
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    {
+        int asn1_len = ASN1_STRING_length(octet);
+
+        if (asn1_len <= 0) {
+            goto done;
+        }
+        point_len = (size_t)asn1_len;
+    }
```
Jakuje: why to create a nested block like this? Can we move the declaration to the beginning of the block?

tests/meson.build
```diff
@@ -123,8 +123,23 @@ foreach t, sources : test_programs
...
+
 setup_script=find_program('setup.sh')
 all_suites=['softokn', 'softhsm', 'kryoptic', 'kryoptic.nss']
+foreach suite : all_suites
+  test(
+    'ec-point-decode',
+    ec_point_test,
+    suite: suite,
+  )
+endforeach
```
Jakuje: if I see right, this is more an unit test, that really does not run against either of the pkcs11 modules and if all of them are enabled, it runs 3 times the same test:
```
  1/125 softokn - pkcs11-provider:ec-point-decode        OK              0.00s
  2/125 softhsm - pkcs11-provider:ec-point-decode        OK              0.00s
  3/125 kryoptic - pkcs11-provider:ec-point-decode       OK              0.00s
```
https://github.com/openssl-projects/pkcs11-provider/actions/runs/31610653422/job/94162092731?pr=767

I think its our first "unit test" so I wonder if we should make a suite "unittests" and add such tests into that.

It would, though, need some change in how we run the tests in CI as we select only one suite in some of the jobs.

src/obj/ec_point.c
idrassi: Good catch. That block was left over after the amend from the earlier version check that selected different OpenSSL APIs. I will move `asn1_len` with the other declarations and keep the signed check before casting to `size_t`.

tests/meson.build
idrassi: This is indeed a unit test and the loop registers the same executable several times. I did some research and Meson allows a test to carry several suite labels, so I will register it once with `unittests` plus the existing module suites. That keeps the current CI filters working and makes a full test run execute it only once.

idrassi: Done.

src/obj/ec_point.c
idrassi: Done.

idrassi:
I checked the CI failures. They all stop while installing Fedora Rawhide packages, before checkout or compilation. Rawhide has just moved from Fedora 45 to 46, and some mirrors are still serving packages signed with the Fedora 45 key.
The failed jobs should be rerun once the mirrors catch up.

Jakuje (approved):
Correct. The Fedora rawhide is broken now. We might want to consider going back to fedora:latest as it should have new-enough openssl already, but that would be for different PR.

simo5:
I am rerunning them, as some mirrors are still ok, so with enough luck we'll get all green in a few retries

simo5 (approved):
LGTM

simo5:
Thank you!

PR #770, opened 2026-08-18
Detect mechanism support and pkcs11-tool support for key generation

### LINKED ISSUE #758: Generate PQC keys using pkcs11-tool
The tests/setup.sh has a comment about pkcs11-tool not supporting PQC keys. This is no longer the case since opensc 0.27.1, but we need to double-check we have opensc updated in all the  environments that run PQC tests (with kryoptic as far as I know for now).

> Actually this should already work in current opensc. But I am not sure if it made it already to all our container images. 

 _Originally posted by @Jakuje in [#756](https://github.com/openssl-projects/pkcs11-provider/pull/756/changes#r3710714129)_

simo5:
#### Description

Use pkcs11-tool -M to do some more heavy lifting and auto-detection of what token supports and what pkcs11-tool supports

For keypair generation we'll have to also wait until https://github.com/OpenSC/OpenSC/pull/3795 makes into Fedora, but the fallback to openssl will work in the meanwhile.

Fixes #758

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/tests/setup.sh
+++ b/tests/setup.sh
@@ -21,11 +21,11 @@ fi
 TOKENTYPE=$1
 
 # defaults -- overridden below or in the per-token setup
-SUPPORT_ED25519=1
-SUPPORT_ED448=1
+SUPPORT_ED25519=0
+SUPPORT_ED448=0
 SUPPORT_EDDSA_PARAMS=1
-SUPPORT_X25519=1
-SUPPORT_X448=1
+SUPPORT_X25519=0
+SUPPORT_X448=0
 SUPPORT_RSA_PKCS1_ENCRYPTION=1
 SUPPORT_RSA_KEYGEN_PUBLIC_EXPONENT=1
 SUPPORT_TLSFUZZER=1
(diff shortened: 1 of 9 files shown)


tests/softokn-init.sh
```diff
@@ -29,17 +29,15 @@ export TOKENCONFIGVARS="export NSS_LIB_PARAMS=configDir=${TOKDIR}"
...
-export SUPPORT_ML_DSA=0
-export SUPPORT_ML_KEM=0
-export SUPPORT_SLH_DSA=0
+# Edward curves are not well supported in NSS, Ed25519 requires a special OID
+# for generation in ECParams, while Ed448 is not supported at all
+export SUPPRESS_ED25519=1
+export SUPPRESS_ED448=1
+
+# Montgomery curves are not supported in NSS yet, but there is no way to
+# audtodetect it
```
Jakuje:
```suggestion
# Montgomery curves are not supported in NSS yet, but there is no way to
# autodetect it
```

tests/setup.sh
```diff
@@ -330,13 +452,25 @@ if [ "${SUPPORT_ED25519}" -eq 1 ]; then
...
+        OPENSSL_CONF=${URIPEM_OPENSSL_CONF}
+        ossl '
+        genpkey -propquery "provider=pkcs11"
+                -algorithm ED448
+                -pkeyopt "pkcs11_uri:pkcs11:object=${ED2CRTN};id=${URIKEYID}"'
+        OPENSSL_CONF=${BASE_OPENSSL_CONF}
+    else
+        echo "Bad value for SUPPORT_ED448_KEY_GEN: $SUPPORT_ED448_KEY_GEN"
+        exit 1
+    fi
```
Jakuje: I wonder if we should not make some helper function from these as they repeat a lot, with just couple of variables.

simo5: perhaps, let me take a look

simo5: added a commit to change that

Jakuje: you changed the detection (which also had a lot of repetition), but not the key generation I pointed out here (which has also quite a lot of repetition).

tests/setup.sh
```diff
@@ -21,33 +21,52 @@ fi
...
+SUPPRESS_ML_DSA=${SUPPRESS_ML_DSA:-0}
+SUPPRESS_ML_KEM=${SUPPRESS_ML_KEM:-0}
+SUPPRESS_SLH_DSA=${SUPPRESS_SLH_DSA:-0}
 
 # Ed448 requires OpenSC 0.26.0
 OPENSC_VERSION=$(opensc-tool -i | grep OpenSC | sed -e "s/OpenSC 0\.\([0-9]*\).*/\1/")
 if [[ "$OPENSC_VERSION" -le "25" ]]; then
-    SUPPORT_ED448=0
+    SUPPRESS_ED448=1
 fi
```
Jakuje, line 58: I think this requirement was only for the pkcs11-tool -- we should be able to generate the ed448 also with openssl if the old opensc is present. But not sure how it should be expressed in this environment.

tests/setup.sh
```diff
@@ -305,13 +362,13 @@ echo ""
...
-if [ "${SUPPORT_ED25519}" -eq 1 ]; then
+if [ "$SUPPORT_ED25519_KEY_GEN" -ne 0 ]; then
     # generate ED25519
     get_next_keyid
     EDCRTN="edCert"
 
-    ptool --keypairgen --key-type="EC:edwards25519" --id="$KEYID" \
-    	  --label="${EDCRTN}" 2>&1
+    gen_keypair "$SUPPORT_ED25519_KEY_GEN" "EC:edwards25519" "ED25519" \
+                "${EDCRTN}" "${KEYID}" "${URIKEYID}"
```
Jakuje, line 371: I wonder if we could drop the KEYID and URIKEYID params as they are same in all the cases. And maybe also the TSTCRTN as its already set in most of the cases to make it more "macro like". Given thats just local function, I think it would be acceptable.

tests/setup.sh
Jakuje, line 338: I wonder if we should replace also these to keep it consistent.
```suggestion
gen_keypair "1" "EC:secp256r1" "" \
            "${ECPEERCRTN}" "${KEYID}" "${URIKEYID}"
```

Jakuje (approved):
Actually approving as these are mostly cosmetic changes

tests/setup.sh
simo5: Yeah I thought about that, but then again it is not a huge deal to keep them there, and will allow us to do some "special" stuff if needed.

tests/setup.sh
simo5: well gen_keypair is really built to deal with the SUPPORT_* stuff, and would be a little useless to do it for RSA and EC which are always assumed for granted.

Perhaps a future patch once we start disabling those algorithms on some token ...

tests/setup.sh
simo5: true, but I do not think it is worth the added complexity.

Eventually we'll just remove this line and just fail CI if opensc is too old.

PR #779, opened 2026-08-24
Misc fixes 

simo5:
#### Description

A minor issue found by coverity.
I do not really think the same object may concurrently be used by multiple threads, but it cost nothing to make Coverity happy so I added a commit for it.

An issue I found while working on the last PR, the SLH-DSA code was missing the part sin obj/store.c needed to actually store SLH-DSA public keys on import, when needed (rarely).

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/obj/store.c
+++ b/src/obj/store.c
@@ -643,6 +643,105 @@ static CK_RV prep_mlkem_find(P11PROV_CTX *ctx, const OSSL_PARAM params[],
     return CKR_OK;
 }
 
+static CK_RV prep_slhdsa_find(P11PROV_CTX *ctx, const OSSL_PARAM params[],
+                              struct pool_find_ctx *findctx)
+{
+    const OSSL_PARAM *p;
+    CK_RV rv;
+
+    switch (findctx->param_set) {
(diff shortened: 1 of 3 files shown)


src/obj/store.c
```diff
@@ -1142,6 +1249,39 @@ static CK_RV p11prov_store_mlkem_public_key(P11PROV_OBJ *key)
     return store_key(key, NULL, template, tmpl_cnt);
 }
 
+static CK_RV p11prov_store_slhdsa_public_key(P11PROV_OBJ *key)
```
Jakuje, line 1252: This function is exactly same as the `p11prov_store_mldsa_public_key()`. While technically different key type, structure is the same and I would prefer some de-duplication.

The `p11prov_store_slhdsa_private_key()` is different only in the macro `OSSL_PKEY_PARAM_SLH_DSA_SEED` name (while the value is same).

simo5: I have a followup PR in the works that will deal with this, not worth changing it in this PR.

Jakuje (approved):

PR #780, opened 2026-08-25
Refactor how storing public/private keys work

simo5:
#### Description

Refactor how storing public/private keys work

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:


src/kmgmt/ec.c
```diff
@@ -94,10 +94,390 @@ DISPATCH_KEYMGMT_FN(ec, gettable_params);
...
+            return -1;
+        }
+
+        template[3].type = CKA_CLASS;
+        template[3].pValue = DISCARD_CONST(&priv_class);
+        template[3].ulValueLen = sizeof(CK_OBJECT_CLASS);
+
+        template[4].type = CKA_ID;
+        template[4].pValue = NULL;
+        template[4].ulValueLen = 0;
```
Jakuje: no `CKA_ID` on public key?

src/kmgmt/ec.c
```diff
@@ -94,10 +94,390 @@ DISPATCH_KEYMGMT_FN(ec, gettable_params);
...
+                          OSSL_PKEY_PARAM_PUB_KEY);
+            rv = CKR_KEY_INDIGESTIBLE;
+            goto done;
+        }
+
+        if (((char *)p->data)[0] == '\x02' || ((char *)p->data)[0] == '\x03') {
+            int plen;
+
+            P11PROV_debug(
+                "OpenSSL 3.0.7 BUG - received compressed EC public key");
```
Jakuje, line 365: should we run this branch only for the affected openssl version(s)? We do not have any code coverage for this block either.

src/kmgmt/ec.c
```diff
@@ -94,10 +94,390 @@ DISPATCH_KEYMGMT_FN(ec, gettable_params);
...
+
+    if (!template) {
+        switch (class) {
+        case CKO_PUBLIC_KEY:
+            return EC_PUBKEY_TEMPLATE_SIZE;
+        case CKO_PRIVATE_KEY:
+            return EC_PRIVKEY_TEMPLATE_SIZE;
+        default:
+            return -1;
+        }
```
Jakuje: This code is never executed. Is it needed? This applies to the same block in all the functions for other key types.

src/kmgmt/ec.c
```diff
@@ -94,10 +94,390 @@ DISPATCH_KEYMGMT_FN(ec, gettable_params);
...
+            template[5].type = CKA_VERIFY;
+            template[5].pValue = DISCARD_CONST(&val_true);
+            template[5].ulValueLen = sizeof(CK_BBOOL);
+            return 6;
+
+        case CKK_EC_MONTGOMERY:
+            template[5].type = CKA_DERIVE;
+            template[5].pValue = DISCARD_CONST(&val_true);
+            template[5].ulValueLen = sizeof(CK_BBOOL);
+            return 6;
```
Jakuje: This code is not executed in tests. Not sure why

src/kmgmt/ec.c
```diff
@@ -94,10 +94,390 @@ DISPATCH_KEYMGMT_FN(ec, gettable_params);
...
+
+        if (curve_name) {
+            rv = p11prov_kmgmt_privkey_to_id(
+                ctx, attrs, &numattrs, (uint8_t *)curve_name,
+                strlen(curve_name), (uint8_t *)ecparams, ecplen, p->data,
+                p->data_size);
+        } else {
+            rv = p11prov_kmgmt_privkey_to_id(ctx, attrs, &numattrs,
+                                             (uint8_t *)ecparams, ecplen,
+                                             p->data, p->data_size, NULL, 0);
```
Jakuje, line 416: is there a way to implement a code coverage for this branch?

src/kmgmt/slhdsa.c
```diff
@@ -21,17 +32,282 @@ DISPATCH_KEYMGMT_FN(slhdsa, gen_settable_params);
...
+_Static_assert(SLHDSA_PUBKEY_TEMPLATE_SIZE <= P11PROV_PUBKEY_MAX_TEMPLATE_SIZE,
+               "SLH-DSA public key template size exceeds maximum");
+_Static_assert(SLHDSA_PRIVKEY_TEMPLATE_SIZE
+                   <= P11PROV_PRIVKEY_MAX_TEMPLATE_SIZE,
+               "SLH-DSA private key template size exceeds maximum");
+
+static int p11prov_slhdsa_get_template(P11PROV_OBJ *obj, CK_OBJECT_CLASS class,
+                                       const OSSL_PARAM *params,
+                                       CK_ATTRIBUTE *template)
+{
```
Jakuje, line 55: This function has no code coverage -- sounds like a bug somewhere on the way to this function.

src/kmgmt/ec.c
simo5: Apparently we never stored a CKA_ID on imported public keys ... after all what would it be ?

simo5: Ah this is missing on private keys now ... sad, let me fix it, we have the CKA_ID stored in the object already at this stage, but we didn't while I was working on the PR and the transfer got lost

simo5: Currently the code is technically correct as it backfills the CKA_ID in the store.c file after retrieving the template, but this can be improved:
```
    for (int i = 0; i < tmpl_cnt; i++) {
        if (template[i].type == CKA_ID) {
            template[i].pValue = findctx.attrs[0].pValue;
            template[i].ulValueLen = findctx.attrs[0].ulValueLen;                                                                                        
            break;                                                                                                                                       
        }                                                                                                                                                
    }
```

simo5: I pushed a patch to address this

src/kmgmt/ec.c
simo5: I think we still have people using 3.0.7 ... I wonder if we can detect the openssl version we are currently loaded into, mind opening a follow-up issue for this ?

src/kmgmt/ec.c
simo5: Uhmm initially I wanted to be able to request the template size to do allocation, but then I changed strategy and forgot to remove this, I'll push a patch to do that

simo5: removed

src/kmgmt/ec.c
simo5: I added tests, hopefully now is covered

src/kmgmt/slhdsa.c
simo5: added tests

src/kmgmt/ec.c
simo5: this should happen every time a private EC key is stored, key generation should cover it ?

simo5: aah the case we do not hae curve_name ... hmmm
importing a private EC key not using curve_name ?

simo5: basically requires importing a private key using explicit params for definition ..

src/kmgmt/ec.c
```diff
@@ -94,10 +94,394 @@ DISPATCH_KEYMGMT_FN(ec, gettable_params);
...
+            template[cnt].ulValueLen = sizeof(CK_BBOOL);
+            cnt++;
+            return cnt;
+
+        case CKK_EC_MONTGOMERY:
+            template[cnt].type = CKA_DERIVE;
+            template[cnt].pValue = DISCARD_CONST(&val_true);
+            template[cnt].ulValueLen = sizeof(CK_BBOOL);
+            cnt++;
+            return cnt;
```
Jakuje, line 214: This branch is still not covered in tests. Not sure if we do something special with the montgomery curves somewhere down the road.

src/kmgmt/slhdsa.c
```diff
@@ -21,17 +32,281 @@ DISPATCH_KEYMGMT_FN(slhdsa, gen_settable_params);
...
+        template[cnt].ulValueLen = p->data_size;
+        cnt++;
+
+        p = OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_SLH_DSA_SEED);
+        if (p) {
+            template[cnt].type = CKA_SEED;
+            template[cnt].pValue = p->data;
+            template[cnt].ulValueLen = p->data_size;
+            cnt++;
+        }
```
Jakuje, line 151: This part is not executed. I am not sure if we try seed-only objects (I do not think so).

Jakuje (approved):

src/kmgmt/ec.c
simo5: This should be hit in the timport test when we run the ECDH exchange as we import once the private key and once the public key....

And now that I look at the code, that is not necessarily true, because if CKA_P11PROV_PUB_KEY is set on the object (and we do set it) then the key is never stored, because the ECDH interface is an odd one in PKCS#11 where the peer key value is passed as a byte string instead of passing a key handle ... and therefore the code is never forced to actually try to store a public key during ECDH exchange ...

src/kmgmt/slhdsa.c
simo5: Uhmmm perhaps we do not store seed on export ??

simo5: (well or openssl does not ...)

simo5: OpenSSL does not

simo5:
Resolved three threads about code coverage, where currently I can't add coverage due to pkcs#11 or OpenSSL limitation, but is worth keeping the code anyway.

simo5:
Thanks for the review!

PR #782, opened 2026-08-27
Restrict EC point conversion to OpenSSL <= 3.0.7

### LINKED ISSUE #781: Consider dropping openssl 3.0,7 workaround (or conditionally compile it)
src/keymgmt/ec.c function p11prov_ec_get_find_attrs() has a code block for OpenSSL 3.0.7 workaround where we receive compressed EC key parameters.

While there might be users out there, we should not run through this code on versions where we know it wont be hit, which I believe should be detectable either at runtime or build time (for example do not do this with OpenSSL 4.0 at all).

should we run this branch only for the affected openssl version(s)? We do not have any code coverage for this block either.

_Originally posted by @Jakuje in https://github.com/openssl-projects/pkcs11-provider/pull/780#discussion_r3860991611_

simo5:
#### Description

Compressed EC public key points only need manual decompression workarounds in OpenSSL versions 3.0.7 and earlier. Wrap the handling logic in a version check macro and move the local buffer allocation inside the conditional block.

Fixes #781 

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:

### DIFF
--- a/src/kmgmt/ec.c
+++ b/src/kmgmt/ec.c
@@ -314,7 +314,6 @@ static CK_RV p11prov_ec_get_find_attrs(
     EC_POINT *point = NULL;
     BN_CTX *bn_ctx = NULL;
     const OSSL_PARAM *p;
-    uint8_t pub_data[MAX_EC_PUB_KEY_SIZE];
     const char *curve_name = NULL;
     int curve_nid;
     unsigned char *ecparams = NULL;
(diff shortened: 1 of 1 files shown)


Jakuje (approved):

beldmit:
It will not work this way. 
OpenSSL 3.2 with RHEL fips provider (3.0.7) will require special efforts.

simo5:
> It will not work this way. OpenSSL 3.2 with RHEL fips provider (3.0.7) will require special efforts.

So are you suggesting to just close the bug and leave this workaround in always enabled for the time being?

simo5:
@jakuje, so we may to retain this for the rare case where a fips provider (based on 3.0.7, like the one in RHEL is) is used in conjunction with pkcs11-provider and we are exporting a public key from the fips provider and importing it into pkcs11-provider...  it is a bit unlikely but not impossible given we added support for trusting  the pkcs11-provider to claim it is FIPS compliant ...

Jakuje:
I guess we will have to wait a bit then. Fine for me.

simo5:
postponing to some future date

PR #784, opened 2026-09-02
Support environment variables in pin-source

### LINKED ISSUE #776: No way to pass PIN via the environment?
Hi!

In the old worls with engines, various tools could be configured to pass the PIN in a specified environment variable. That doesn't work with pkcs11-provider, which is a surprise. Could you explain why not please?

Thanks

simo5:
#### Description

Allow supplying PINs via environment variables using the `env:` schema (for example, `pin-source=env:VARIABLE_NAME`).

This updates PKCS#11 URI parsing and pin resolution to read from the specified environment variable while validating against the maximum allowed PIN length, updates documentation, and adds test coverage.

Fixes #776 

#### Checklist

- [x] Documentation updated

#### Reviewer's checklist:


tests/tbasic
```diff
@@ -121,6 +130,23 @@ if [ $FAIL -eq 2 ]; then
...
+title PARA "Test fetching public key with PIN source env exceeding max length"
+LARGE_PIN_ENV=$(printf '%1024s\n' '' | tr ' ' 'A')
+export LARGE_PIN_ENV
+FAIL=0
+output=$($CHECKER "${OPENSSL}" pkey -in "${PRIURI};pin-source=env:LARGE_PIN_ENV" 2>&1) && FAIL=1
+if [ $FAIL -ne 0 ]; then
+    echo "Expected failure when PIN exceeds maximum length, but command succeeded"
+    exit 1
+fi
+echo "$output" | grep -iE "error|failed|cannot|unable|could not" > /dev/null 2>&1 || FAIL=2
```
Jakuje, line 142: could we grep for more specific error message or are the errors that much different across different openssl versions that we need that vague grep?

src/util.c
```diff
@@ -305,7 +288,61 @@ static int get_pin_file(P11PROV_CTX *ctx, const char *str, size_t len,
...
+
+    if (strncmp((const char *)pinsource, "file:", 5) == 0) {
+        ret = get_pin_file(ctx, pinsource + 5, output);
+    } else if (strncmp((const char *)pinsource, "env:", 4) == 0) {
+        ret = get_pin_env(ctx, pinsource + 4, output);
+    } else if (*pinsource == '|') {
+        ret = EINVAL;
+    } else {
+        /* missing 'file:' is accepted */
+        ret = get_pin_file(ctx, pinsource, output);
```
Jakuje: this code path does not have code coverage. Adding simple test would help.

src/util.c
```diff
@@ -305,7 +288,61 @@ static int get_pin_file(P11PROV_CTX *ctx, const char *str, size_t len,
...
+{
+    CK_ULONG max_pin_len = p11prov_ctx_max_pin_length(ctx);
+    const char *val;
+    char *pin;
+    size_t len;
+
+    val = getenv(envvar);
+    if (val == NULL) {
+        P11PROV_debug("Failed to get pin from %s (env var not found)", envvar);
+        return ENOENT;
```
Jakuje, line 304: covering this with negative test case should be also easy

src/util.c
```diff
@@ -926,7 +963,10 @@ CK_RV p11prov_uri_match_token(P11PROV_URI *uri, CK_SLOT_ID slot_id,
 int p11prov_get_pin(P11PROV_CTX *ctx, const char *in, char **out)
 {
     if (strncmp(in, "file:", 5) == 0) {
-        return get_pin_file(ctx, in, strlen(in), (void **)out);
+        return get_pin_file(ctx, in + 5, (void **)out);
+    }
+    if (strncmp(in, "env:", 4) == 0) {
+        return get_pin_env(ctx, in + 4, (void **)out);
```
Jakuje: no code coverage here either. Please, consider adding separate test for this too

tests/tbasic
simo5: they may change across openssl versions, so I did not want to spend too much time on it

src/util.c
simo5: we do have a test for env passed pin, how is this not covered ?

Jakuje: You have test for `env:` in PKCS#11 URI, but not for pin coming from configuration file (`pkcs11-module-token-pin` option).

simo5: pkcs11-module-token-pin is the default for all other tests?

src/util.c
simo5: added test for this

src/util.c
Jakuje: Yes, but without `env`.

simo5: oh I see, I do not thik I actualy want to support specifying an env var in config, env var should only be supported in URIs IMO

simo5: dropped in a fixup, if you disagree let me know

src/util.c
simo5: I also decided to drop this fallback, I do not think we should be this lax with uris

src/util.c
Jakuje: Agreed.

src/util.c
Jakuje: Sounds good!

tests/tbasic
```diff
@@ -198,7 +244,7 @@ OPENSSL_CONF=${ORIG_OPENSSL_CONF}
 OPENSSL_CONF=${OPENSSL_CONF}.nopin
 
 title PARA "Test interactive Login on key without ALWAYS AUTHENTICATE"
-output=$(expect -c "spawn -noecho $CHECKER ${TESTBLDDIR}/tsession \"$BASEURI\";
+output=$(expect -c "spawn -noecho $CHECKER ${TESTBLDDIR}/tsession \"${BASEURI:?BASEURI is unset}\";
```
Jakuje, line 247: whats the reason for this change?

Jakuje (approved):

tests/tbasic
simo5: $BASEURI naked was causing shellcheck to grumble so I decide to improve the thing to return a proper error if BASEURI is not actually set (I was trying to run tbasic by hand as some point :)

Jakuje: in other cases we are using mostly something like this so I would prefer to keep doing so rather than changing it ad-hoc like this:
```
# shellcheck disable=SC2153 # PUBURIs is assigned
```

simo5: yes, but this disables the check entirely, what I did passes the check instead, isn't it better?

steve-pexip:
Awesome! :tada:

PR #785, opened 2026-09-02
Implement SSHKDF support

### LINKED ISSUE #426: Implement SSH KDF
**Describe the feature**
The SSH KDF is almost like the SP 800 108 Feedback mode, but does not contain the 0x00 separator byte, see

https://datatracker.ietf.org/doc/html/rfc4253#section-7.2

My reading is that this KDF can not be constructed from the SP800-108 KDF itself:

https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-108r1-upd1.pdf

OTOH, the PKCS#11 flexibility of the KDF API should allow to implement the SSH KDF with something like the example in the specs:

https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061457

Therefore I do not think we need separate implementation in the pkcs11 module, but just the wiring in the 

**Expected behavior**
The SSH applications should be able to delegate the KDF to the PKCS#11 module through the provider API.

simo5:
#### Description

Add support for the SSHKDF key derivation function in the PKCS#11 provider. This includes adding wrapper support for PKCS#11 C_DigestKey, implementing the SSHKDF provider operations, and adding a test suite for SSHKDF key derivation.

Fixes #426 

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:


tests/tsshkdf
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash -e
+# Copyright (C) 2024 Simo Sorce <simo@redhat.com>
```
Jakuje:
```suggestion
# Copyright (C) 2026 Simo Sorce <simo@redhat.com>
```

src/kdf.c
```diff
@@ -1002,16 +1029,421 @@ const OSSL_DISPATCH p11prov_tls13_kdf_functions[] = {
...
+            }
+            kctx->type = *(const char *)p->data;
+        } else if (p->data_type == OSSL_PARAM_INTEGER
+                   || p->data_type == OSSL_PARAM_UNSIGNED_INTEGER) {
+            int val;
+            ret = OSSL_PARAM_get_int(p, &val);
+            if (ret != RET_OSSL_OK) {
+                return ret;
+            }
+            kctx->type = (char)val;
```
Jakuje: The `OSSL_KDF_PARAM_SSHKDF_TYPE` is defined as UTF8 string in documentation:

https://docs.openssl.org/3.0/man7/EVP_KDF-SSHKDF/#identity

Under which conditions we should accept octet string or integer?

I see `EVP_KDF_SSHKDF_TYPE_INITIAL_IV_CLI_TO_SRV` is confusingly defined as integer/char value though:
```
openssl/kdf.h:#define EVP_KDF_SSHKDF_TYPE_INITIAL_IV_CLI_TO_SRV 65
```

src/kdf.c
```diff
@@ -1002,16 +1029,421 @@ const OSSL_DISPATCH p11prov_tls13_kdf_functions[] = {
...
+        return RET_OSSL_OK;
+    }
+
+    p = OSSL_PARAM_locate(params, OSSL_KDF_PARAM_SIZE);
+    if (p) {
+        return OSSL_PARAM_set_size_t(p, SIZE_MAX);
+    }
+
+    return RET_OSSL_OK;
+}
```
Jakuje, line 1579: This function is never executed by the openssl tools in the tests now. I think this is the case also for the other KDFs we have here. We should likely implement separate C-test. This should be accessible with `EVP_KDF_CTX_get_kdf_size()` (openssl 4.0 only?) if I see right.

src/kdf.c
```diff
@@ -1002,16 +1029,421 @@ const OSSL_DISPATCH p11prov_tls13_kdf_functions[] = {
...
+    if (strcmp(paramname, OSSL_KDF_PARAM_KEY)) {
+        /* ignore anything but a "key" param */
+        return RET_OSSL_OK;
+    }
+
+    p11prov_obj_free(kctx->key);
+    kctx->key = p11prov_obj_ref(key);
+
+    return RET_OSSL_OK;
+}
```
Jakuje, line 1495: This code path is untested. Should we add separate test generating the skey output?

src/kdf.c
simo5: uhmm it is a utf8 string, let me drop the nonsense alternative checks

src/kdf.c
simo5: is it worth it?
I can add a C program I guess ...

src/kdf.c
simo5: let me add a C test to cover all this instead of a simple script

simo5: converted the test into a C test, with support for both SKEY and normal interface as well as a bunch of tests that check things are properly accepted/denied.

simo5:
I completely replaced the simple test with a more compliacted one, however now the test is entirely skipped on openssl < 4.0 as it depends on OSSL_PROVIDER_load_ex() to be available ...

I think this is fine for some of the newer tests.

simo5:
OSSL_PROVIDER_load_ex() became available in 3.2, I had a bug in check for it

simo5:
Additional commit to handle the idosyncratic way NSS softokn does things, where some mechanism are available only on one slot and some only on the other ...

src/kdf.c
```diff
@@ -1002,16 +1031,558 @@ const OSSL_DISPATCH p11prov_tls13_kdf_functions[] = {
...
+            ret = p11prov_GetAttributeValue(
+                kctx->provctx, p11prov_session_handle(orig_session),
+                p11prov_obj_get_handle(kctx->key), &val_attr, 1);
+            if (ret == CKR_OK && val_attr.ulValueLen > 0
+                && val_attr.ulValueLen != CK_UNAVAILABLE_INFORMATION) {
+                val_attr.pValue = OPENSSL_malloc(val_attr.ulValueLen);
+                if (val_attr.pValue) {
+                    ret = p11prov_GetAttributeValue(
+                        kctx->provctx, p11prov_session_handle(orig_session),
+                        p11prov_obj_get_handle(kctx->key), &val_attr, 1);
```
Jakuje: Why not to use the `p11prov_fetch_attributes` helper here?

tests/meson.build
```diff
@@ -167,6 +168,7 @@ tests = {
   'fork': {'suites': all_suites},
   'oaepsha2': {'suites': ['softokn', 'kryoptic', 'kryoptic.nss']},
   'hkdf': {'suites': ['softokn', 'kryoptic', 'kryoptic.nss']},
+  'sshkdf': {'suites': ['softokn', 'kryoptic', 'kryoptic.nss']},
```
Jakuje: whats not working on softhsm in this case?

tests/tsshkdf.c
```diff
@@ -0,0 +1,840 @@
...
+static void test_aes_encrypt_decrypt(OSSL_LIB_CTX *def_libctx,
+                                     OSSL_LIB_CTX *p11_libctx, EVP_SKEY *skey,
+                                     const unsigned char *expected_key,
+                                     size_t key_len, const char *test_name)
+{
+    const char *cipher_name = NULL;
+    if (key_len == 16) {
+        cipher_name = "AES-128-CBC";
+    } else if (key_len == 24) {
+        cipher_name = "AES-192-CBC";
```
Jakuje, line 263: not a big deal, but these two ciphers are never tested according to the code coverage report (only AES-256)

Jakuje (dismissed):
the inlilne comments are mostly nits. I think this is good to merge with the fixes or without.

src/kdf.c
simo5: good point, let me change that

tests/meson.build
simo5: I think I was doing some testing and removed and forgot to read, will check it out

tests/tsshkdf.c
simo5: true, but doesn't hurt to leave it there eitehr

tests/meson.build
simo5: So this is the issue I get with softhsm:
```
Testing SSHKDF classic interface ...
  SSHKDF classic: SHA256, type A, 512 bytes ...
EVP_KDF_derive failed for 'SHA256, type A, 512 bytes'
808B711D897F0000:error:41000101:pkcs11:p11prov_create_secret_key:The desired action cannot be performed because an appropriate user is not logged in:../src/obj/create.c:45:Failed to create secret key
```

simo5: I added a fallback function to try to log in if creation fails with CKR_USER_NOT_LOGGED_IN

Jakuje (approved):

PR #788, opened 2026-09-04
Treat bignum RSA key attributes as big-endian when comparing.

### LINKED ISSUE #787: Compare bignum RSA key attributes as big-endian
**Describe the bug**
When checking that a public key, typically from a certificate matches a private key, typically from smart card, their MODULUS and PUBLIC EXPONENT are compared as bignums. This comparison is done in `cmd_bn_attr` which converts PKCS#11 attributes from binary to `BIGNUM` using `BN_native2bn`. This usually works because both numbers are converted to the same, backwards, bignums so they compare equal. But the PKCS#11 standard states that numbers like CKA_MODULUS and CKA_PUBLIC_EXPONENT are of type "Big integer", unsigned most-significant byte first. This becomes a problem when a pkcs11 module returns numbers with leading zeros. Leading zeros are used in a DER coded ASN.1 INTEGER to ensure that values are positive if the first bit is 1. So E933 is encoded as 00E933 and some pkcs11 modules return those three bytes when queried for, for example PUBLIC_EXPONENT. `BN_naitive2bn` would convert that to the number 33E900. If the key from a certificate is decoded as E933 it is converted to the BIGNUM 33E9 which does not equal 33E900.

Suggest to replace the use of `BN_naitive2bn` with `BN_bin2bn`. That would be more correct with regard to the PKCS#11 standard and it would allow for pkcs11 implementations that return numbers with leading zeros (which is arguably dubious).

**To Reproduce**

**Expected behavior**
RSA public key attributes should compare equal regardless of if there are leading zeroes or not.

 - OS: all

Discovered in a proprietary pkcs11 module that cannot be disclosed.

**Additional context**

jobbflykt:
#### Description

Closes #787

Treat RSA key attributes as big-endian when checking for key equality. Allows for key attributs, like CKA_PUBLIC_EXPONENT, that have leading zeros to compare equal to the same values without leading zeroes.

This is more correct with regards to the PKCS#11 standard.

#### Checklist
- [x ] Code modified for feature

#### Reviewer's checklist:

### DIFF
--- a/src/obj/keymgmt.c
+++ b/src/obj/keymgmt.c
@@ -514,8 +514,10 @@ static int cmp_bn_attr(P11PROV_OBJ *key1, P11PROV_OBJ *key2,
         return rc;
     }
 
-    bx1 = BN_native2bn(x1->pValue, x1->ulValueLen, NULL);
-    bx2 = BN_native2bn(x2->pValue, x2->ulValueLen, NULL);
+    /* PKCS#11 values, like modulus and public exponent, are of type "Big
+       integer", unsigned integer most-significant byte first. */
+    bx1 = BN_bin2bn(x1->pValue, x1->ulValueLen, NULL);
+    bx2 = BN_bin2bn(x2->pValue, x2->ulValueLen, NULL);
 
     if (BN_cmp(bx1, bx2) == 0) {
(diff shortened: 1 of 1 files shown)


simo5:
Thanks!

PR #789, opened 2026-09-06
Do not log in with the default PIN for public key ops on other slots

### LINKED ISSUE #557: Incorrect slot for new associated object
**Describe the bug**

First of all I'm not really sure if this is actually reproducible in the current master but there is a chance it might happen too. I'm adding this mainly to not forget that this issue is there as it should be resolved before default PIN selection in #536 is merged.

My client has been reporting issue with blocking PIN when they use multiple slots with tokens having different PINs. I should note that they run a version with few patches including the default PIN selection implemented in #536

After digging through the logs (extended with some extra custom info) and code I noticed following.

After the object (public key in this case) is imported, it is followed by MATCH which tries to find associated object which means calling find operation. The issue is in the session (specifically slot) selection before this operation is started. The problem is that when such object is imported, the slotid is set to `CK_UNAVAILABLE_INFORMATION` so when session is selected, it uses following logic:

https://github.com/latchset/pkcs11-provider/blob/c8abb3496b312533824db24f0d2337788d074264/src/session.c#L947-L978 

The slot selected here is then set to the object that is then added to the pool. The selected session in this flow does not require login so it doesn't fail here.

The actual PIN failure happens later when verify is used and the key is duplicated and subsequently token login is required. It will use the slot selected above but because there is not pin in object (no refresh URI), it will use the configured PIN for default slot. However if the default slot does not match the slot for the session (which can easily happen when the default slot is configured and maybe there is a chance it happens anyway), then it result in incorrect PIN failure.

Currently I came up with following change ( https://github.com/bukka/pkcs11-provider/commit/cf899117148811c6e554119842913f6b6ab16c68 ), which is not yet confirmed that it fixes the issue but the idea is to select the default slot before getting the session when looking for an associated object and use that slot for the object. But I'm wondering what's the actual point of the current session slot selection above. Why doesn't it just always use the default slot?

**To Reproduce**

This is not easy to reproduce as it requires some slot preconditions. So far I haven't managed to recreate but it might be possible to recreate it artificially which I will attempt to do later.

**Expected behavior**

New associated object should be assigned to the default slot

 - Linux otherwise does not matter

 - Device: SoftHSMv2 but any device supporting multiple slot can show this issue
 - Application: Nginx

bukka:
#### Description

Fixes #557

After a fork there is no logged in session, so public key operations on a key stored in a non-default slot could attempt a login with the configured PIN of the default slot. That burns a PIN retry, marks the PIN as bad and the next attempt leaves an error on the stack and falls into an interactive PIN prompt.

Two paths did this: `cache_key()` took a login session without a URI, and `p11prov_obj_refresh()` eagerly refreshed the associated private key. The key cache now only uses an already logged in session and suppresses its errors, and associated objects are refreshed on demand only.

The new slotpin test in the kryoptic.multislot suite runs verify, encrypt and sign with the second slot key in a forked child and fails on any error left on the stack, PIN prompt or `CKR_PIN_INCORRECT` in the debug log.

#### Checklist

- [ ] Documentation updated

#### Reviewer's checklist:


src/session.c
```diff
@@ -867,7 +867,8 @@ static CK_RV fetch_session(P11PROV_SESSION_POOL *pool, CK_FLAGS flags,
 #define LOCK_SLEEP 5000
 static CK_RV slot_login(P11PROV_SLOT *slot, P11PROV_URI *uri,
                         OSSL_PASSPHRASE_CALLBACK *pw_cb, void *pw_cbarg,
-                        bool reqlogin, P11PROV_SESSION **_session)
+                        bool reqlogin, bool no_login,
```
simo5, line 870: Having 2 booleans here makes things confusing, esp given reqlogin and no_login both set to true sounds like would be invalid.

What about replacing this pair with an new enum?

enum p11prov_login_request {
    LOGIN_REQUIRED,
    LOGIN_NOT_REQUIRED,
    LOGIN_ONLY_CACHED,
};

a small doc section woudl explain what the three cases do (the latter is: "return success if a login session exist but never try to actively login")

simo5 (changes requested):
I like this, but I would like to see an enum used, see comments

simo5:
About the unsuccessful PIN attempt, in what case do we not have the second slot PIN if w already authenticated once? Is this a case of a user that disables pin caching and does not use URIs with PINs ?

src/session.c
bukka: This is done.

src/session.c
```diff
@@ -1026,6 +997,8 @@ CK_RV p11prov_get_session(P11PROV_CTX *provctx, CK_SLOT_ID *slotid,
     CK_SLOT_ID id = *slotid;
     P11PROV_SESSION *session = NULL;
     CK_FLAGS flags = DEFLT_SESSION_FLAGS;
+    enum p11prov_login_request login =
+        reqlogin ? LOGIN_REQUIRED : LOGIN_NOT_REQUIRED;
```
bukka, line 1001: I was considering changing p11prov_get_session as well but there are quite a few callers so it would get much bigger so I left it like this for now but if you think that it should be changed, I will be happy to do that.

bukka:
> About the unsuccessful PIN attempt, in what case do we not have the second slot PIN if w already authenticated once? Is this a case of a user that disables pin caching and does not use URIs with PINs ?

This is after public key got refreshed and later used in verify which called key caching. I already described that case in https://github.com/openssl-projects/pkcs11-provider/issues/557#issuecomment-2806338604

The PIN caching should not really matter here because it doesn't cache PIN from URI (which is what I use) - as far as I can see it caches only prompt. And when global pin is configured, it still gets a priority.

src/session.c
simo5: yeah I think we should use the enum only and get rid of the boolean across the code base.

src/session.h
```diff
@@ -20,7 +20,14 @@ CK_RV p11prov_get_session(P11PROV_CTX *provctx, CK_SLOT_ID *slotid,
 CK_RV p11prov_try_session_ref(P11PROV_OBJ *obj, CK_MECHANISM_TYPE mechtype,
                               bool reqlogin, bool rw,
                               P11PROV_SESSION **_session);
+/* How a login session is obtained */
+enum p11prov_login_request {
+    LOGIN_REQUIRED, /* wait for a busy login session and log in */
+    LOGIN_NOT_REQUIRED, /* log in if possible but do not wait */
```
simo5, line 26: LOGIN_NOT_REQUIRED means we do not even try to login, and do not fail if the session is not logged in.

simo5 (changes requested):

PR #793, opened 2026-09-12
Fix ChaCha20 cipher key import using the wrong CKK_AES key type

AlexLaroche:
## Summary

`p11prov_cipher_legacy_init()` in `src/cipher.c` always imported raw cipher keys as `CKK_AES` -- a leftover from when AES was the only symmetric cipher this provider drove through a PKCS#11 token. With `CKM_CHACHA20_POLY1305` support now available in some tokens (e.g. [kryoptic](https://github.com/latchset/kryoptic/pull/505)), this causes `C_MessageEncryptInit`/`C_MessageDecryptInit` to fail with `CKR_KEY_TYPE_INCONSISTENT` on any token that correctly enforces `CKA_KEY_TYPE == CKK_CHACHA20` for that mechanism.

This isn't just a test-suite nuisance: it breaks real TLS connections that force ChaCha20-Poly1305 record encryption through the pkcs11 provider (`-propquery ?provider=pkcs11`), surfacing as a confusing "bad record mac" alert on the very first encrypted record, rather than a clear key-type error at cipher init.

## Fix

- `src/cipher.c` -- `p11prov_cipher_legacy_init` now picks the key type from the mechanism (`CKK_CHACHA20` for `CKM_CHACHA20_POLY1305`, `CKK_AES` otherwise) instead of hardcoding AES.
- `src/kmgmt/skey.c` / `src/registration.h` -- registered a new `CHACHA20` SKEY key-management algorithm (mirroring the existing `AES` one), since `EVP_SKEY_import_raw_key()`-based key setup goes through a separate skeymgmt algorithm lookup that only recognized `"AES"` and `"GENERIC-SECRET"`.
- `tests/taead.c` -- hardcoded `"AES"` as the SKEY type name for every algorithm under test; it now derives the type from the algorithm name so ChaCha20-Poly1305 imports as `CKK_CHACHA20` there too.

## Testing

Verified against a local [kryoptic](https://github.com/latchset/kryoptic) build with ChaCha20-Poly1305 support:
- The `aead` and `tls` test suites, previously failing with `CKR_KEY_TYPE_INCONSISTENT` and "bad record mac" respectively, now pass (both the "Skey" and "Legacy" sub-cases of the AEAD test).
- Full existing kryoptic-backed test suite: 35/37 passing (2 unrelated skips), no regression to AES.


simo5 (dismissed):
Except for the style issue, everything looks good to me.

AlexLaroche:
@simo5 Fixed in f9c269d -- the three spots clang-format flagged (`src/cipher.c`, `src/kmgmt/skey.c`, `tests/taead.c`) are now reformatted to match. CI should be green now.

PR #795, opened 2026-09-22
Release Version 1.3.0

simo5:
#### Description

It's been 7 months since the last release, time for a new one.


Jakuje (approved):
Changelog will be generated just on the github release? I wonder if we should either remove the NEWS and ChangeLog files (that are empty) or include the changelog in them to have it in text format (rather than proprietary github format).
Otherwise no concerns!

simo5:
> Changelog will be generated just on the github release? I wonder if we should either remove the NEWS and ChangeLog files (that are empty) or include the changelog in them to have it in text format (rather than proprietary github format). Otherwise no concerns!

Yeah we never really cared for committing the changelog, and it is a remnant from when we used autotools (IIRC), as they required the file to be present. With meson I think we can rmove it in a future commit.

